Issue #14, opened 2017-04-15
Components to tables


korpa:
Hi Rafael,

Thank you very much for these great components. Sadly a few features are missing that I could use it in my projects. One of the most important features I would need is the following:

I would need something similar as described in #8 but for tables. Filters are a good way to change the data it self, but if I want to add html/css as well, this does not work. Therefore injecting components would be a great feature.

Here are a few examples how I would like to represent data in tables:

![screenshot_20170415_115943](https://cloud.githubusercontent.com/assets/2982785/25062835/3675bd2a-21d5-11e7-9f86-8fd746d8de32.png)

![screenshot_20170415_121343](https://cloud.githubusercontent.com/assets/2982785/25062873/c3975e5c-21d5-11e7-8734-58687aaff94a.png)

rafaelpimpa:
For HTML/CSS you can do like this following the [docs example](http://localhost:8080/#/documentation/table).

Add the ``html`` property to Table (be aware this will be changed to ``render-html`` in next release:
```html
<b-table html>
    <!-- Others columns -->
    <b-table-column 
        field="date" 
        label="Date" 
        sortable :format="formatDate">
    </b-table-column>
</b-table>
```

Return the html you want in the filter:
```javascript
formatDate(value, row) {
    return `<span class="tag is-primary">
                ${new Date(value).toLocaleDateString()}
            </span>`
}
```

korpa:
I tried this before but sadly this does not work:
![screenshot_20170416_113120](https://cloud.githubusercontent.com/assets/2982785/25070194/59b8b364-2298-11e7-85fe-af1bf1de79ca.png)

I had a similar problem with one of my components. When I read how to solve this in Vue, the makers of Vue clearly said that you should not add HTML code into variables but use components.
Therefore I still think that you should add a feature to inject components.

If you don't mind I would dig into this the next days and submit a PR as proposal.

rafaelpimpa:
You missed the ``html`` attribute on table.

Sure if you have the time to make a PR I would welcome. If you wait a few days I'll make the code a bit more readable adding comments for every methods.

korpa:
PR was merged => closed

PR #16, opened 2017-04-16
Use components to add HTML to table columns

korpa:
See #14 
v-html should not be used: https://vuejs.org/v2/guide/syntax.html#Raw-HTML Furthermore components are more flexible in my opinion.

Therefore I changed the table component that you can inject components instead of using the format method to style cells.

@rafaelpimpa I have to clean up a few things e.g. documentation. But before I do this I want to show this to you if you would be happy with this way.


korpa:
Perfect. I will fix all these points and provide an updated PR.

rafaelpimpa:
Nice :)

Don't forget to update the docs as well, and if you want a sugestion: I would use the Icon component for the docs example, you can generate another json in this site: https://www.mockaroo.com

rafaelpimpa:
Also I just added the [contribution guide](https://github.com/rafaelpimpa/buefy/blob/master/.github/CONTRIBUTING.md)

korpa:
Hi @rafaelpimpa,

I fixed the topics you mentioned + provided a documentation. I added a few sentences regarding the differences plus pros and cons of the renderHtml and the component injection method. I'm not sure, if you like it this way. If not, feel free to update it.

Please let me know if you are willing to accept this PR already or if there are still things you would like to be updated.

Regarding the contribution guide: You ask contributors to use 4 space indention, but your editorconfig file says 2 space indention.

PR #17, opened 2017-04-17
Fixes on table

zerbfra:
Fix to preserve order when changing data (eventually we can come up with a boolean).
Fix to make checked items reactive (with prop)


src/components/table/Table.vue
```diff
@@ -125,14 +125,18 @@
                 default: 20
             },
             paginationSimple: Boolean,
-            html: Boolean
+            html: Boolean,
+            checkedRows: {
+                type: Array,
+                default: () => []
+            }
```
rafaelpimpa: I think the default should be just an empty array instead of a function.

rafaelpimpa:
I just accepted another PR on this Table file, I think you should clone the repo again and make your changes, since the last PR changed a lot of things.

Also, no need to build, I'll do this before publishing :)

rafaelpimpa:
I'm adding the ``updatingData`` right now to the Table myself, didn' realise it before, thanks :)

src/components/table/Table.vue
rodymol123: Using just the [] or {} as default value on props results in warnings being thrown by Vue, this is the solution as described by Vue's creator: https://github.com/vuejs/vue/issues/1032

rafaelpimpa: Thanks for the heads up. Didn't know that.

rafaelpimpa:
Closing this since next release will cover all the last thing it was proposed here.

Issue #8, opened 2017-04-12
Rich modal with custom component


ciaasteczkowy:
Would you consider adding another modal, that can be open programmatically, where you could pass a component? This would allow to create rich modals (eg. with forms).

rafaelpimpa:
Sure, it's actually on my [backlog](https://github.com/rafaelpimpa/buefy/projects/1). It won't be totally programatically (like Snackbar & Toast) but something like Notification but with methods.

ciaasteczkowy:
That's great! I was wondering if it could look like this:

```javascript
import UserForm from '~components/UserForm'

export default {
  methods: {
    openUserForm () {
      this.$dialog.custom({
        component: UserForm
      })
    }
  }
}
```

rafaelpimpa:
Hmm, I see, seems like it's not that complicated to do this, will take a look later

korpa:
@rafaelpimpa: I could have a look into this and add a component injection like I did for tables. In my opinion this is the most flexible way.

What do you think?

dyreby:
I think this would be a great addition that would really expand the capabilities of buefy's dialog/modal feature.  Beyond forms I could also imagine dynamic content like progress indicators or a small modal-based UI that would allow a developer to pop up a quick control panel or settings editor.  I'm not sure how everything would work with a component that requires props, but using vuex behind the scenes would allow a modal component complete access to the state and vuex mutations / actions.

(And selfishly this is exactly what I need for a project I'm working on right now...)

@korpa I wish I could be some help, but I'm still learning JavaScript development and git workflows.  If you do have time to implement this feature I would really appreciate it.

@rafaelpimpa Thanks for your work on buefy!!

rafaelpimpa:
Actually I'm not really sure if the Dialog should be mixed into this, I was thinking about creating another component (maybe Modal) to do this kind of stuff, inject components, show images and so on.

What you guys think?

ciaasteczkowy:
I think it's a good idea. Dialogs would be a set of predefined modals with custom content inside a `modal-card`. That would make sense, if you allow users to put custom components in place of `modal-card`. Of course if `modal-background` and `modal-close` will be configurable with options, I won't complain :smile:

korpa:
Hi guys,
I missed the discussion :( while I added a component injection into dialog :)  #27 
Let's use it as a further discussion basis

rodymol123:
[...]

PR #27, opened 2017-04-19
Dialog: component injection added

korpa:
This is a proposal how to inject a component into dialogs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,13 @@
 * **New component: Select**
 * **New component: Dropdown**
 * **BREAKING:** Rename Table property ``html`` to ``render-html``
-* Fix #6 — Docs working on IE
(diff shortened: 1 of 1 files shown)


rafaelpimpa:
I'm sorry I didn't see your message before asking in #8, it went unnoticed :(

Would you separate this on a new ``Modal`` component?

It's more of a semantic issue for me, if you're injecting an entire form or anything else for example, it's not a Notice/Dialog anymore, should be treated differently.

Then I'll create a new branch for this so I can clone to see how's it 👍

korpa:
I will look into this. I think it even could be easier to do this as a separate component, as I may can use "vue slots"

korpa:
I reseted my master and therefor the PR is closed. I will open a new one, as soon as I have a new component.

The component injection into dialog can be found here: https://github.com/korpa/buefy/tree/27-component-in-dialog

rafaelpimpa:
Thank you

PR #31, opened 2017-04-20
Add Modal component

### LINKED ISSUE #8: Rich modal with custom component

korpa:
fixes #8

### DIFF
--- a/docs/pages/documentation/notices/Modal.vue
+++ b/docs/pages/documentation/notices/Modal.vue
@@ -1,0 +1,93 @@
+<template>
+    <div>
+        <h1 class="title is-spaced">Modals</h1>
+        <h2 class="subtitle">Modals inform users</h2>
+        <hr>
+
+        <div class="columns">
+            <div class="column">
+                <h3 class="subtitle">Modal</h3>
+
+
+                <button class="button is-primary" @click="showMyRichModal = true">Open modal</button>
+
+
+                <b-modal v-if="showMyRichModal" @close="showMyRichModal = false">
+                    <header slot="header" accesskey=""class="modal-card-head">
+                        <p class="modal-card-title">Header title</p>
+                    </header>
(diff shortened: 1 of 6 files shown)


rodymol123:
I would love to see this have the same body appending functionality like the dialog has, because it currently is a nightmare with z-index and absolutely positioned containers.

Should be a simple copy and paste trick from the Dialog component, but could save a lot of headaches.
And as always, thanks for being the amazing person you are by contributing to something I like ;)

Example

rafaelpimpa:
Exactly what I thought, making it programatically like Dialog (creating dynamically direct to the body element) would be better.

rodymol123:
Ditching the card for generic elements would make it more in line with the options Bulma set out with the modal, like image modals. And it would allow for the card modals too.

rafaelpimpa:
Yep, the thing is, if you're inject a component, that component would be the card you want. Just have to think about a way to close the modal within the injected component.

rodymol123:
Maybe emit a close event that can be handled by the parent?

korpa:
Hi guys,
the idea was to use slots and that you can define the modal directly were you want use it. But I see the problem. Will give it another try.

@rodymol123 Could you give me the code of the example dashboard your screenshot shows? Then I can test it directly in the use case.

rodymol123:
@korpa Can you run Laravel? If you can't, I will detach it from my Laravel setup for you ;)

korpa:
Yes, Laravel is fine.

rodymol123:
@korpa I sent you an invite to my private repo containing the Laravel project and VueJS app :)

rafaelpimpa:
Closing this because I'm adding it myself, release goes live probably this week.

PR #39, opened 2017-04-24
Make bDropdown work server side

iwonapiotrowska:
Hey, it's a great library, thanks!!!!!! We use it in our Project. Small improvement:
Window doesn't exists server side so we check if it exists before running code that depends on it. We need this for nuxt.js support.

### DIFF
--- a/src/utils/SelectMixin.js
+++ b/src/utils/SelectMixin.js
@@ -373,16 +373,20 @@ export default {
         }
     },
     created() {
-        document.addEventListener('click', this.clickedOutside)
-        window.addEventListener('resize', this.calcDropdownInViewportHorizontal)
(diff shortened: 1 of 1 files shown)


rafaelpimpa:
Thank you, probably have to make the same with Dialog.

I believe it won't work fully with SSR because (the click-outside won't work).

shentao:
Hey @rafaelpimpa! Thanks for merging this! 

The click-outside functionality might not be relevant when doing SSR. All it has to do on the backend is to render the component’s HTML and data. Once it’s on the client side, Vue should pick it up and set up the listeners as it would normally.

rafaelpimpa:
That's great! Didn't know it was that smart.

PR #53, opened 2017-05-08
Add `type="button"` to buttons to prevent submit

shentao:
Otherwise, it breaks usage within `<form>` elements due to automatically triggering the form submit.


rafaelpimpa:
Thank you

shentao:
Sure thing! Thanks for this amazing library. :D

PR #59, opened 2017-05-11
Fix the documentation/table demo code

JasonMillward:
Ran into this today while testing the demo table code. Just thought it would be nice to provide a working example to future devs.


rafaelpimpa:
Woa thanks, forgot to update that :X

PR #71, opened 2017-05-24
add class to checked rows

mikejavier:
It may seem simple, but it was very useful for me because I could visually differentiate a marked row.

image


rafaelpimpa:
Nice! Thank you

PR #76, opened 2017-05-26
Update Modal onCancel function

gongzza:
when Modal is closed, i want to receive arguments.

### DIFF
--- a/src/components/modal/Modal.vue
+++ b/src/components/modal/Modal.vue
@@ -64,7 +64,7 @@
              * Emit events, and destroy modal if it's programmatic.
              */
             close() {
-                this.onCancel()
+                this.onCancel.apply(null, arguments)
                 this.$emit('close')
                 this.$emit('update:active', false)
 


gongzza:
for example
``` js
// ModalForm.vue
this.$emit('close', 'gongzza')

// Parent.vue
this.$modal.open({
  component: ModalForm,
  onCancel: name => {
    this.name = name
  }
})
```
when Modal is closed, `this.name`'s value will be `gongzza`.

rafaelpimpa:
Hi, thanks for contributing!

Does the ``apply`` is needed here? We're not passing another ``this`` 🤔

gongzza:
i'm sorry. i can't understand what you said that "We're not passing another `this`"?

i looked this. so i used `apply`
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Spread_operator

rafaelpimpa:
Nevermind, you're right.

PR #77, opened 2017-05-26
add index to table

UrVerySpecial:
Hello, Rafael

Thanks a lot. your components are very simple, and nice.

I needed index in table, so i have added that.
If there is no problem, could you merge?

best regards, kang


rafaelpimpa:
Sure! Thank you

PR #78, opened 2017-05-26
fix table sort

gwvt:
Sort all falsy values except 0 to end, leave 0 values in normal sort order. Fixes 0 values being sorted to end, e.g., a list of numbers sorted as [1, 2, 0], where one would expect [0, 1, 2]. Null values still sorted to end, e.g., [0, 1, 2, null].


rafaelpimpa:
Great! Thank you.

PR #80, opened 2017-05-28
Fix bug in dropdown 

mikejavier:
Fixing a bug in the dropdown component. This only happens on mobiles with IOS. You try to close the dropdown when clicking outside it and nothing happens. To correct it simply add a `cursor: pointer` in the style of `.dropdown.brackground`


rafaelpimpa:
Thank you!

PR #90, opened 2017-06-04
:bug: Adding flex-direction to tab-content

mirshko:
This prevents long elements such as pre and code blocks that have a long string of text in them from overflowing off the container of the tab-content

## Example

See example _code block was set to width 100%_

screenshot 2017-06-04 15 53 08

This should be restricted to the container width. To fix this I added the flex-direction column rule to the `.tabs-content` class.

## Fix

This makes the code block the proper width and is constrained to the size of the tab-content container rather than overflowing

screenshot 2017-06-04 15 53 37

### DIFF
--- a/src/scss/components/_tabs.scss
+++ b/src/scss/components/_tabs.scss
@@ -3,6 +3,7 @@
         position: relative;
         overflow: hidden;
         display: flex;
+        flex-direction: column;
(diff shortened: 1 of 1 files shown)


rafaelpimpa:
Thanks a lot!

PR #101, opened 2017-06-15
update quick start codes in README with semicolons for easy copy-pasting

gczh:
Added semicolons to make it easy to copy-paste. Prevents making careless mistakes while copy-pasting over to projects. Makes things easier for people to pick up and implement buefy from the get-go.


rafaelpimpa:
Nice! Thanks :)

PR #105, opened 2017-06-16
Fix numeric 0 input

### LINKED ISSUE #104: [b-input] numeric zero values are not showing up
### Overview of the problem

**Buefy** version: [0.4.0]
**Vuejs** version: [2.3.3]
**OS/Browser**: Ubuntu 16.04/Chrome 57.0.2987.110

### Description
I just noticed that text input component has a similar problem as  https://github.com/rafaelpimpa/buefy/issues/82

0 numeric values are not showing up.
screenshot_2017-06-17_01-36-14

### Steps to reproduce
Bind input to a numeric number with a value of 0.

### Expected behavior
Input value should be 0, not empty.

iwan-wijaya:
Issue: #104

### DIFF
--- a/src/components/input/Input.vue
+++ b/src/components/input/Input.vue
@@ -67,7 +67,7 @@
             @click.native="togglePasswordVisibility">
         </b-icon>
 
-        <small class="help counter" v-if="maxlength">{{ this.newValue.length }} / {{ maxlength }}</small>
+        <small class="help counter" v-if="maxlength">{{ valueLength }} / {{ maxlength }}</small>
     </p>
 </template>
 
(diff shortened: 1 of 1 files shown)


src/components/input/Input.vue
```diff
@@ -109,9 +109,7 @@
         },
         data() {
             return {
-                // This is necessary because, for some reason, textarea
-                // returns undefined if there's no v-model or value.
```
iwan-wijaya: @rafaelpimpa I can't reproduce this problem
```
// This is necessary because, for some reason, textarea
// returns undefined if there's no v-model or value.
```
I've tested textarea type and it seems to be working fine.

rafaelpimpa: Let me test this again before merging, it was a pretty weird behavior if I remeber correctly.

rafaelpimpa:
Thanks a lot

Issue #117, opened 2017-06-23
Table row wierdness with input - row loop needs key?


jaybeecave:
### Overview of the problem

**Buefy** version: [X.X.X]
**Vuejs** version: [X.X.X]
**OS/Browser**:

### Description
I think we need a way to specify a key for the loop using one of the props. In this case I would use the JobID on each of the rows. The data looks like this.
```
[{"JobID":73,"JobNumber":"61817AS","JobDate":"2017-05-12T00:00:00+12:00","JobType":"Bulk Order","DeliverTo":"Jon Devonshire","InvoiceTo":"Malcolm Stewart","DateCreated":"2017-05-12T13:27:49.122+12:00","DateModified":"2017-06-23T15:15:24.527163+12:00","ClientID":2,"Name":"Wow","PurchaseOrderNumber":"P60621","InvoiceDate":"2017-05-12T00:00:00+12:00","JobLayouts":null,"Status":"Done"}]
```
### Steps to reproduce
use an input inside of a table row and then change the records. This will cause the input event to trigger incorrectly because It thinks that the row value has changed rather then the whole object for the row.

![example](https://image.prntscr.com/image/F3xRV9MfR9C8qaffXbsn_w.png)

### Expected behavior
The data changes without triggering input

### Actual behavior
The data changes and triggers input for each row

jaybeecave:
So I am on to something with the key. I added the following prop
```
 <b-table data-key="JobID" :data="records" :striped="true" :mobile-cards="true" :default-sort="['JobDate', 'desc']">

        props: {
            dataKey: {
                type: String,
                required:true
            }, 
        // ...
```

And I no longer have the issue.

 Problem is that its a breaking change... I am just trying to make the key optional now, but any thoughts are welcome.

Maetiz:
I think this issue still exists. The input event is triggered when the table is sorted.

I fixed this locally by allowing the row key to be specified instead of using the index from the v-for.

PR #120, opened 2017-06-23
Added datakey prop to table for a key on the table loop

jaybeecave:
Fix for #117


jaybeecave:
I thought about doing that also... But will it work reliably? I didn't test but in my head they still have the same index?

rafaelpimpa:
It will be unique for each row

jaybeecave:
Can you pass the index through the row props also? its useful to have

rafaelpimpa:
It already does (in the scoped slot)

Issue #130, opened 2017-06-28
Checkbox not emitting change event


gongzza:
### Overview of the problem

**Buefy** version: [0.4.2]
**Vuejs** version: [2.3.4]
**OS/Browser**: Windows/Chrome

### Description
i used the b-checkbox component like this:
`<b-checkbox v-model="todo.done" @change="update(todo)"></b-checkbox>`

i could get the change event when i clicked the checkbox
but when i pressed space or enter key, i couldn't get the change event.

`todo.done` is changed well.

### Steps to reproduce

1. Focus the checkbox (using tab key or click).
2. Press space or enter.
3. Didn't call the update function.

### Expected behavior

Get the change event when the v-model is changed

### Actual behavior

Didn't get the change event

gdutwyg:
@gongzza  if i want to focus the checkbox and Press enter , the checkbox should not  be selected(but space can select), how  should i do?

gdutwyg:
@rafaelpimpa

gdutwyg:
i want to preventDefault(), but it don't work

PR #131, opened 2017-06-28
Update Checkbox.vue

gongzza:
when the enter key or space key is pressed,
set the value and emit change event.


gongzza:
#130

rafaelpimpa:
Nice catch!

PR #145, opened 2017-06-30
Enter key press on Autocomplete option into a Form

jtommy:
Hi Rafael,
i have just fixed a issue about enter key press on autocomplete option into a Form.
In the last release 0.4.4, enter key press on option not work because the option (hover) isn't selected and the form is submitted.
I have changed from keyup to keydown and i have just tried this fix in all browsers (IE10+, Chrome, Firefox, Safari).


rafaelpimpa:
Thanks for this, didin't know that the form submit was on keydown

PR #150, opened 2017-07-03
"Lazy" binding Input component

jtommy:
Hi Rafael,
i have just fixed a issue about Input component. 
At the moment it's imposibile have a "lazy" binding; according to Vue documentation v-model.lazy is equals to :value and @change: but now the input method related to Input component emit a 'change' event and so it's impossibile have only the final value (on change).


rafaelpimpa:
Well this is a breaking change, there must be people using the @change event.

I'll wait until Bulma 0.5.0 release for the breaking changes.

PR #155, opened 2017-07-04
"Lazy" binding Input component

jtommy:
PR #150

### DIFF
--- a/src/components/input/Input.vue
+++ b/src/components/input/Input.vue
@@ -174,7 +174,6 @@
                 const value = event.target.value
                 this.newValue = value
                 this.$emit('input', value)
-                this.$emit('change', value)
                 !this.isValid && this.checkHtml5Validity()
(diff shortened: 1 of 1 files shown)


rafaelpimpa:
Actually that is not a breaking change since already emit on @change 👍 

Thanks!

rafaelpimpa:
I've tested and still doesn't work, look at https://github.com/vuejs/vue/issues/5743

``v-model`` modifiers doesn't work on custom components.

jtommy:
Yes, it's ok. 
"v-mode.lazy" doesn't work on custom component but the equivalent form is <b-input :value="myVar" @change="value => myVar = value" />. The problem was about 'change' event into input method and a lazy solution (value + change) didn't work.

rafaelpimpa:
oh I see, it's fine then 👍

PR #156, opened 2017-07-05
Add module

ForsakenHarmony:
See: https://github.com/rollup/rollup/wiki/pkg.module


rafaelpimpa:
The non compiled files are gonna work with that?

ForsakenHarmony:
oh right, it contains vue files, so probably not
I just use `import Buefy from 'buefy/src/index';`

PR #173, opened 2017-07-12
Fix Autocomplete docs (Async example) and double emit 'input' event

jtommy:
@rafaelpimpa 
I have just fixed the docs (now @input => real time) and a issue about double emit 'input' event on select. As you can see the watch function on newValue emit a 'input' so it's not required to emit an other event after the assignment.


rafaelpimpa:
Thanks!

PR #174, opened 2017-07-13
Fotello Support

AliLozano:
I added fontello support

### DIFF
--- a/src/components/icon/Icon.vue
+++ b/src/components/icon/Icon.vue
@@ -43,7 +43,7 @@
         done: 'ok',
         info: 'info-circle',
         check_circle: 'ok-circle',
-        warning: 'exclamation-triangle',
+        warning: 'attention',
         error: 'attention-circled',
         arrow_upward: 'up-big',
         chevron_right: 'angle-right',
--- a/src/components/icon/Icon.vue
+++ b/src/components/icon/Icon.vue
@@ -1,6 +1,6 @@
 <template>
     <span class="icon" :class="[type, size]">
-        <i :class="[newPack, newPack === 'fa' ? `fa-${newIcon}` : null ]">{{ newPack === 'mdi' ? newIcon : null }}</i>
+        <i :class="[newPack, iconClass]">{{ newPack === 'mdi' ? newIcon : null }}</i>
     </span>
 </template>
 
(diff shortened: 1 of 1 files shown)


src/components/icon/Icon.vue
```diff
@@ -33,6 +33,17 @@
                 } else {
                     return this.icon
                 }
+            },
+            iconClass() {
+              if (this.newPack === 'fa') {
```
rafaelpimpa: I think that a switch/case here is better to read than a if/else chain.

src/components/icon/Icon.vue
```diff
@@ -33,6 +33,17 @@
                 } else {
                     return this.icon
                 }
+            },
+            iconClass() {
```
rafaelpimpa: Missing comment (see contribution guide)

AliLozano:
Fontello can customize icon names and be like font awesome, but for avoid user have to customize their icons with font awesome's names, we can leave that they configure their packs  with their name, a dictionaries of mdi equivalent, prefix, and their style (mdi: name in element and fa: name in class), see the code:

```
// default pre configurated fonts
faDefault = {name: 'fa', prefix: 'fa fa-', buefyicons: { done: 'check' .. others icons ..}, style: 'fa'};
fontelloDefault = {name: 'fontello', prefix: 'icon-', buefyicons: { done: 'check' .. others icons ..}, style: 'fa'};
mdiDefault = {name: 'mdi', prefix: 'mdi', buefyicons: { done: 'done' .. others icons ..}, style: 'mdi'};
packs = {fa: faDefault, fontello: fontelloDefault, mdi: mdiDefault}
```

uses.. 
```
Vue.use(Buefy, { defaultIconPack:  'fontello'}); // read default configuration from packs
```

and configurating their custom pack:
```
Vue.use(Buefy, { defaultIconPack:  {name: 'mycustomicons', prefix: 'custom-', buefyicons: { done: 'check' .. others icons ..}}}); 
```

But i don't know how to do, without break something..

rafaelpimpa:
I didn't know fontello I thought it was a icon lib, I see the problem. 

I'm not sure if we should support things like that in Buefy for now. I plan to use SVGs as default internal icon in the future and then we can add other sources to use only with ``b-icon`` (like what you did).

AliLozano:
@rafaelpimpa I added generic packs support, to be more flexible, packs could be configured with something like `packs = {packs, ...config.customPacks }`, also packs could be in another file, but depends of you.

rafaelpimpa:
As I understand Fontello, users selects anything they want, they don't have to select that icons you have used as base, I don't think it's a good implementation.

AliLozano:
I'm disagree with you closed this push, Its a good solution, this supports any icon fonts library(no only fontello), only it's needing create which is their equivalent icon, being more flexible. In the future if you want to include icons only as svg, you only need to add a new pack and change the default pack to svg, leaving the users to use the buefy pack, material, fontware, fontello or any icon fonts, even their custom pack without force user to use svgs or mdi.

If fontello is the problem, maybe you can delete fontello, and create a configuration to add custom packs with `packs = {packs, ...config.customPacks }`

Issue #178, opened 2017-07-14
Feature Request: Menu


tsctao:
I'm writing menu component in my private project. I see your menu component is planned in projects page. If you haven't finished it yet, can I contribute this one.

PR #180, opened 2017-07-16
Add Menu component

tsctao:
#178 

I take some time during you doesn't respond my issue yet to make it. I'm not sure if you like this one.

I make it to support multi-level menu and router-link component. The difficulty is how to manage states across these components, so the data flow is a bit ugly.

Edit: I forgot it, toggle animation is also difficult part. I take a lot of time to make it work every browser and my mobile device. You may review it again to make sure it works.

### DIFF
--- a/docs/pages/documentation/general/Menu.vue
+++ b/docs/pages/documentation/general/Menu.vue
@@ -1,0 +1,426 @@
+<template>
+    <div class="container">
+        <h1 class="title is-spaced">Menu</h1>
+        <h2 class="subtitle">
+            A simple menu, for any type of vertical navigation
+        </h2>
+        <hr>
+
+        <div class="columns">
+            <div class="column is-3">
+                <b-menu>
+                    <b-menu-list label="General">
+                        <b-menu-item>Dashboard</b-menu-item>
+                        <b-menu-item>Customer</b-menu-item>
+                    </b-menu-list>
+                    <b-menu-list label="Administration">
+                        <b-menu-item>Team Settings</b-menu-item>
+                        <b-menu-dropdown>
+                            <template slot="trigger">Manage Your Team</template>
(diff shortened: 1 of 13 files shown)


rafaelpimpa:
Hey @tsctao I really like it! It's well documented and following the guidelines, thank you very much!

The only thing that is not good is the animation, using height for transitions won't have a good performance, in fact if we change the ``<transition>`` (of MenuDropdown.vue) to ``<transition-group>`` and animate the ``<li>`` instead, [if I'm not mistaken] Vue automatically uses the [FLIP technique](https://aerotwist.com/blog/flip-your-animations/) to animate, could you try that?

Ref: https://vuejs.org/v2/guide/transitions.html#List-Transitions

rafaelpimpa:
I'm gonna wait for you fix before updating the project structure (I'm updating vue-cli and fix the eslint).

tsctao:
After I tried to understand the concept and implement it, I have no idea how to get it works. I need some time rethink it thoroughly. Can you give me more hint?

rafaelpimpa:
OK, I'll accept the PR and make some adjustments, thanks!

rafaelpimpa:
After some testing I couldn't make it work with the FLIP technique, my guess is for that to work we have to have every list item as a transition-group to work properly. I'm just gonna remove the animation to not cause lag/fps drops.

tsctao:
Yeah, I thought I need to wrap them all with `transition-group` but I tried it and it doesn't work. Maybe it needs many hacks.

Another way, I will try to apply `translate3d` by calculate the positions manually.

PS. `translate3d` is GPU animation, is it right?

rafaelpimpa:
Didn't see before your question, yep all ``transforms`` properties, ``opacity`` and ``filter`` are GPU-based.

Issue #164, opened 2017-07-07
B-table dont execute default sort if enter direct on component by route


prsolucoes:
Im using last version.

B-table dont execute default sort if enter direct on component by route.

Default sort only works if i navigate from one component toother, if i access direct from URL it dont works.

Can you help me?

rafaelpimpa:
Are you populating the table with an ajax call? Try to call the ``initSort()`` method after the request.

prsolucoes:
Como eu acesso o componente que está no template, dentro dos métodos do Vue? Algo como "minhaTable.initSort".

rafaelpimpa:
**English** in comments please, to access a component instance use the [``ref`` attribute](https://vuejs.org/v2/api/#ref):

```html
<b-table ref="table"></b-table>
```

```javascript
this.$refs.table.initSort()
```

prsolucoes:
Hi,

I doesnt work. I made a video:
https://youtu.be/erQE-GiBbno

My code:
https://github.com/prsolucoes/gohc-web-app/blob/vuejs/src/components/HealthcheckList.vue

Download video:
https://www.dropbox.com/s/jjzupe59ym6d9tn/error-btable.mov?dl=1

Thanks.

rafaelpimpa:
is that ``hcList`` async or promise? If so you should then chain it to the request, but if you change to this on the mounted it may work also:

```javascript
this.$nextTick(() => this.$refs.hcTable.initSort());

**Comment by prsolucoes** [comment]
Hi,

I put:

```
mounted() {
    this.$nextTick(() => this.$refs.hcTable.initSort());
}
```

But doesn't work.

The hcList come from vuex store dispatch. It come from the getters. I dont know if is async or promise.

https://github.com/prsolucoes/gohc-web-app/blob/vuejs/src/vuex/store.js

**Comment by rafaelpimpa** [comment]
Actually... I have no idea why is that happening, Buefy docs is alright and so is my side project.

**Comment by jtommy** [comment]
Hi @prsolucoes, 
I have just seen your project and I think that the problem is about App.vue.
The first time (reload or enter link) you could not have "sync" list before render the page.
You could try for example (App.vue): 

```
import store from '@/store'
...
export default {
   beforeRouteEnter (to, from, next) {
        axios.get('/api/...', 
        }).then(response =>
          store.commit('list', response.data. ...)
          next()
        }).catch(e => {
        })
  },
  methods: {
     ...
  }
}

[...]

PR #182, opened 2017-07-17
B-Table: InitSort async data

jtommy:
#164 
The main problem is about async data in other words when the b-table is ready but data aren't loaded (ajax or other). 
Backend sorting it's ok (server sort) but I can't see what is the default sort column (arrow down or up) because when initSort is called the columns aren't ready (empty array). Same issue no backend sort but always async data.
I don't know if it's the best solution but in the current logic the columns array is required to find the default column and sorting.

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -162,7 +162,8 @@
                 isAsc: true,
                 mobileSort: {},
                 currentPage: 1,
-                _isTable: true // Used by TableColumn
+                _isTable: true, // Used by TableColumn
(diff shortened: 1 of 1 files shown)


rafaelpimpa:
Did it work OK? I was trying to figure a way to do that, I think it's good enough. I would just remove the underscore prefix (I'm using it just to identify the components since vue won't keep the 'name').

jtommy:
Yes, it's work. I have just simulated to load the data json in docs after a timeout about x seconds.

rafaelpimpa:
Thanks!

PR #192, opened 2017-07-19
add editorconfig

davidcostadev:
With this file we can better configure the specifications of the use spaces and etc.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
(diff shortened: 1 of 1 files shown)


rafaelpimpa:
Thought it wasn't necessary anymore with ESLint. Thanks!

davidcostadev:
Not necessarily, the config editor does cover all file types. http://editorconfig.org/

PR #194, opened 2017-07-21
WIP: Datepicker

bartboy011:
Here's the first pass at the datepicker element, let me know what you think!


bartboy011:
Here's a screenshot of the element for quick reference:

screen shot 2017-07-21 at 1 49 49 pm

In that image, July 19 is selected, and July 21 is today's date (hence the outline).

mariolopjr:
@bartboy011 Looks great, I hope it gets mainlined! DatePicker is probably one of the hugest components to forms (who doesn't need a datepicker nowadays LOL?).

I am currently (myself) working on a min/max slider since Bulma/Buefy have none built-in (which I think is also a highly utilized component!). Thanks for your PR, I may cherry pick your patch for my own projects for now. : )

jtommy:
@bartboy011 well done!
I have some requests to complete the component:
- i18n months and days (maybe is better 3 words then one);
- month and year selectable (configurable by props);
- footer buttons (today, clear, close, ..., configurable by props)

bartboy011:
@jtommy Thanks! I'll start working on those edits soon!

wanxe:
This is going to be very nice.
Only a small suggestions: 😸 
 - The possibility of disabling past dates.
 - Add a range of selectable years, for example start 2017 and ends in 2020.

bartboy011:
To do list:

- [x] i18n date names
- [ ] month and year selectable
- [ ] footer buttons
- [ ] earliest and latest selectable dates

bartboy011:
Hey all, just made the requested additions, take a look when you have a chance!

@jtommy the only thing I wasn't able to add was the ability to 'clear' the date out. Because so much of the component is built around the value prop being a Date, if the value is set to null it breaks the component. If you have a suggestion on a workaround, let me know!

src/components/datepicker/Datepicker.vue
```diff
@@ -1,6 +1,16 @@
 <template>
 <b-modal :active.sync="isActive" :width="640">
-    <b-datepicker-table v-model="dateSelected" :pack="packName"></b-datepicker-table>
+    <b-datepicker-table
+      v-model="dateSelected"
+      :pack="packName"
```
jtommy: You don't need a pack prop. You can see Panel or Table component

src/components/datepicker/Datepicker.vue
```diff
@@ -12,10 +22,53 @@ export default {
     props: {
         value: Date,
         active: {
-          type: Boolean,
-          default: false
+            type: Boolean,
+            default: false
         },
-        iconPack: String
+        iconPack: String,
```
jtommy: You don't need a pack prop. You can see Panel or Table component

src/components/datepicker/DatepickerTable.vue
```diff
@@ -22,27 +22,36 @@
...
         <div class="datepicker-table">
             <div class="datepicker-header">
-                <div class="datepicker-cell">Su</div>
-                <div class="datepicker-cell">M</div>
-                <div class="datepicker-cell">Tu</div>
-                <div class="datepicker-cell">W</div>
-                <div class="datepicker-cell">Th</div>
-                <div class="datepicker-cell">F</div>
-                <div class="datepicker-cell">S</div>
+                <div v-for="day in dayNames" class="datepicker-cell">{{day}}</div>
```
jtommy: :key bind is required

src/components/datepicker/DatepickerTable.vue
```diff
@@ -22,27 +22,36 @@
...
                 </b-datepicker-table-row>
             </div>
         </div>
     </div>
+    <div class="level">
+        <div class="level-left"></div>
+        <div class="level-right">
+            <b-field grouped class="level-item">
+                <p class="control" v-if="footerToday">
+                    <button class="button is-primary" @click="focusToday">Today</button>
```
jtommy: i18n (like month and days)

src/components/datepicker/DatepickerTable.vue
```diff
@@ -22,27 +22,36 @@
...
+                </p>
+                <p class="control" v-if="footerClose">
+                    <button class="button is-primary" @click="$emit('close')">Close</button>
```
jtommy: i18n (like month and days)

src/components/datepicker/DatepickerTable.vue
```diff
@@ -57,28 +66,34 @@ export default {
...
+        monthNames: Array,
+        footerToday: Boolean,
+        footerClose: Boolean,
+        earliestDate: Date,
+        latestDate: Date
     },
     data() {
         let iconNames
 
         if (this.pack === 'mdi') {
```
jtommy: See pack prop comment

src/components/datepicker/Datepicker.vue
bartboy011: Would you suggest just using `config.defaultIconPack` as found on line 74 of this file?

src/components/datepicker/DatepickerTable.vue
```diff
@@ -0,0 +1,217 @@
...
+import bDatepickerTableRow from './DatepickerTableRow'
+export default {
+    name: 'bDatepickerTable',
+    components: {
+        bDatepickerTableRow
+    },
+    props: {
+        value: {
+            type: Date,
+            required: true,
```
jtommy: Value isn't required (clear action)

src/components/datepicker/DatepickerTable.vue
```diff
@@ -0,0 +1,217 @@
...
+            iconNames = {
+                left: 'chevron-left',
+                right: 'chevron-right'
+            }
+        }
+
+        return {
+            iconNames,
+            focused: {
+                month: this.value.getMonth(),
```
jtommy: If value is null check current date

src/components/datepicker/DatepickerTableRow.vue
```diff
@@ -0,0 +1,82 @@
...
+</div>
+</template>
+
+<script>
+export default {
+    name: 'bDatepickerTableRow',
+    props: {
+        selectedDate: {
+            type: Date,
+            required: true
```
jtommy: SelectedDate isn't required (clear action)

src/components/datepicker/DatepickerTableRow.vue
```diff
@@ -0,0 +1,82 @@
...
+         */
+        classObject(day) {
+            function dateMatch(dateOne, dateTwo) {
+                return (dateOne.getDate() === dateTwo.getDate() &&
+                    dateOne.getFullYear() === dateTwo.getFullYear() &&
+                    dateOne.getMonth() === dateTwo.getMonth())
+            }
+
+            return {
+                selected: dateMatch(day, this.selectedDate),
```
jtommy: If selectedDate is null then selected is equals to false

src/components/datepicker/Datepicker.vue
```diff
@@ -0,0 +1,111 @@
...
+            this.isActive = false
+        },
+
+        /*
+         * Emit input event with dateSelected as payload to ensure that a date
+         * is provided (accounting for inital value prop being null) then
+         * emit update:active event with value of isActive to sync with trigger
+         */
+        isActive() {
+            this.$emit('input', this.dateSelected)
```
jtommy: You have to remove it because emit input event on open and close modal

src/components/datepicker/Datepicker.vue
jtommy: You can remove config dependecy. As you can see Panel or Table component there is "both" prop that we'are using about custom component

bartboy011:
@jtommy thanks for these!

Month and year selectable: I was addressing this through the passed in value prop. The date passed in is the date used for month and year. I think your suggestions for how to address the default date are better and I will adjust everything accordingly. 

Today button: you're right, that is a better way for that button to work. 

Do you think a second event is necessary? Your suggestions tend to revolve around removing the need for a `v-model` attribute on the component. Do we need to have another event emitted then that is something like `select`?

src/components/datepicker/DatepickerTable.vue
bartboy011: Completed!

src/components/datepicker/Datepicker.vue
bartboy011: Perfect!

jtommy:
Month and year selectable: Default date it's ok but it isn't required (so you can implement Clear button). In general i think something to select year and/or month quickly (think about select your birthday).

About new event, I think it's ok at the moment if you follow my suggestions. I just noticed the issue into watch of isActive prop.

bartboy011:
@jtommy I believe these changes address your entire review! Let me know if there's anything else you see.

src/components/datepicker/DatepickerTable.vue
```diff
@@ -0,0 +1,208 @@
+<template>
+<div class="box datepicker">
+    <div class="level is-marginless">
+        <div class="level-item">
+            <h4 class="has-text-centered datepicker-month level-item">
+                <span @click="decrementMonth" class="datepicker-decrement">
+                    <b-icon icon="chevron_left"
+                      both
+                      type="is-primary is-clickable"
+                      :pack="pack">
```
jtommy: You can remove it

src/components/datepicker/DatepickerTable.vue
```diff
@@ -0,0 +1,208 @@
...
+                    </b-icon>
+                </span>
+                <strong class="has-text-primary">{{nameOfMonth(focused.month)}} {{focused.year}}</strong>
+                <span @click="incrementMonth" class="datepicker-increment">
+                  <b-icon icon="chevron_right"
+                    both
+                    type="is-primary is-clickable"
+                    :pack="pack">
```
jtommy: You can remove it

bartboy011:
@jtommy That's a good question. I'm not certain how I feel about the clear button also closing it. On one hand, not closing the modal on clear encourages the user to choose a date, but on the other, the clear button is an opt-in prop, so maybe I'm making decisions for the developer? What's your opinion?

jtommy:
@bartboy011 I think that clear button is a shortcut to reset the value. Now if you agree with me you could edit current behaviour and do reset+close (like today)

bartboy011:
Great, sounds good to me. Changes will be up in a moment.

jtommy:
@bartboy011 Well done, great work!
The only thing to improve is about a quick way to select a month and year. What do you think about ?

bartboy011:
@jtommy I think I've found an easy way to do this, via selects, however I'm encountering issues getting everything to fit on mobile. Here are a couple of screenshots:

Here's the new look with the selects on desktop:
screen shot 2017-08-08 at 9 40 35 am

And again on mobile:
screen shot 2017-08-08 at 9 41 57 am

Thoughts on how to make this work on mobile?

jtommy:
@bartboy011 Great!
About mobile, no touch o swipe behaviour at the moment. Ok, about remove arrow icons 👍 

Question: How many elements into year select ? Automatic slide up and down by delta ?

bartboy011:
Currently I have it set that, if there is a `latest-date` prop set, the year from that date is the highest the year selector will go. Otherwise, it offers up to 3 years forward. For the earliest date, same thing with the `earliest-date` prop for the year, otherwise it goes back to 1900 (a pretty standard earliest year I see in a lot of date pickers). Yes to the automatic slide.

Here's a screenshot:
screen shot 2017-08-08 at 10 31 35 am

jtommy:
Ok, when you are ok we review it. I think it's ready!

bartboy011:
I just had a thought - with the new b-dropdown options (in v 0.5.0) it's possible to have the datepicker in a dropdown rather than a modal... Should this be given the option to be used as a standalone table, and not require the modal? Something like an `as-modal` prop?

rafaelpimpa:
Hey @bartboy011 thanks for contributing!

Looks like a great WIP, I have few observations:

* As you said, I think it's better to use the dropdown and put the date on a ``b-input`` to really make it like a form control (like autocomplete), I don't think it's necessary an option to open like modal because the dropdown itself behaves like a modal when it's touchscreen.
* The footer buttons labels/buttons could be a slot rather than prop, easier to stylize with icons and others (like the new upload component or dropdown trigger).
* The selectable range functionality could also disable the prev/next buttons don't you think?

Other than that I think it's great, good work on doing without dependency 👍

bartboy011:
@rafaelpimpa awesome notes, thank you! I'll do some overhauling and push commits soon!

bartboy011:
Overhaul done! @jtommy @rafaelpimpa, the only thing I feel like that needs to be done now is allow developers to set the format in which the date is emitted/displayed. Currently if the value passed in is a date object, a date object is emitted, and the same with a string, but that doesn't address `mm/dd/yyyy` formatting. Any thoughts on how to best address this?

prsolucoes:
Only to say that it is a nice library. Good job ppl.

rafaelpimpa:
@bartboy011 Great job! Now it looks like a Datepicker for sure 😄 

About the format, we can pass a function prop for that, and use the ``toLocaleDateString()`` (https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString) as default.

jtommy:
@bartboy011 In addition to @rafaelpimpa you have to check formatValue function because there is issue about timezone. For example when i select a date i don't see the selected date but the previous day.
A possibile solution could be:

```
const toDate = new Date(value)
const date = new Date(toDate.getTime() - (toDate.getTimezoneOffset() * 60000)).toISOString()
```

bartboy011:
@rafaelpimpa Good notes, thanks. I ended up not using `b-dropdown` as I was having some issues with the dropdown opening and then immediately closing when calling both `this.$refs.dropdown.$data.isActive = true` and `this.$refs.dropdown.toggle()`, not really sure why. It also was only occurring on desktop-sized browsers. Regardless, it ended up working out better to simply use standard HTML rather than a component.

I think the reason formatValue is being called so many times is that, in order to allow both a Date object and a String to be passed in as value, the value has to be passed through formatValue in order to appear properly in the input field. Currently, the b-input has `:value="formatValue(value)"` as a prop so that if a Date is passed in, it still is displayed as a string. Additionally, when the value is emitted, `dateSelected` is parsed to a string via formatValue if a string was passed in. So, it's being called once on the emitting of the selected value, once when the b-input component is re-rendered, and then once more when dateSelected is changed. Because it is used to compute the string version of more than one date, it can't be a computed value. So right now I don't really see a way to change the behavior to reduce usage.

I added the `:date-formatter` function prop with the `.toLocaleString()` fallback and it's working great! However, in my timezone, your fix @jtommy caused the same issues you were having without it. I'm not entirely sure how to address that issue at this time... I guess other than having developers pass in a `dateFormatter()` that handles this issue. I've gained a whole lot of respect for the work being done over at Moment.js after this!

Additionally, I went through and made the datepicker fully screen-reader friendly. The table is entirely tab-navigable, aria-labels have been added on any non-standard buttons, and all inputs have keydown listeners attached.

rafaelpimpa:
Thanks @bartboy011!

I've made a few tests using ``b-dropdown`` and it worked good, what I did was replacing every ``isActive`` to ``$refs.dropdown.isActive`` (dropdown ref being the ``b-dropdown`` itself) and remove the ``this.isActive = true`` from the focus event.

If you manage to do this then it's pretty much done ;)

bartboy011:
I've got it working, but it requires moving the b-input into the b-dropdown itself, which causes styling issues with the input. The issue I was having previously, I believe, is that since I was not placing the b-input inside the b-dropdown so that the styling of the b-input was unaffected, the click event listener inside the b-dropdown was closing the dropdown menu when the b-input was clicked. I confirmed this by focusing on the b-input via tab, which opened the b-dropdown without an issue. If you're fine with the styling of the b-input being affected by the b-dropdown then I'll leave this as is and push the commit.

rafaelpimpa:
What kind of style was affected? I didnt't notice.

bartboy011:
Mostly the input width since dropdown containers don't grow to fill the available width the way a standard input does. It's not the biggest issue in the world, but may annoy a few users.

rafaelpimpa:
We can fix that with this I guess:
```sass
.datepicker {
  .dropdown, 
  .dropdown-trigger {
    width: 100%;
  }
}
```

bartboy011:
Works for me! Committing now!

rafaelpimpa:
Thanks a lot for this @bartboy011!

bartboy011:
@rafaelpimpa Absolutely, thanks for all the review! You as well, @jtommy!

jgerbes:
Just a thought regarding the mobile version.... Wouldn't it be better to fallback to the native date input UI on touch screen devices? That's the direction that [flatpickr](https://github.com/chmln/flatpickr) have gone with and I feel that it makes a lot of sense.

jtommy:
@jgerbes We have just added that feature!

jgerbes:
@rafaelpimpa It appears that the datepicker example is the docs isn't live: https://buefy.github.io/#/documentation/datepicker

bartboy011:
The docs on the live site are only through v0.5.0 and the datepicker will be included in release v0.5.1, once that release is out they'll be live.

LiamDotPro:
Is there a chance we could have a time element added to this? Sadly all of the other options with a time element all use bootstrap and I'd prefer to just use bulmar and buefy for everything inside my project.

jtommy:
@LiamDotPro We are working on it.

PR #217, opened 2017-08-02
Fix default value for b-table data prop

Tirke:
To avoid the Vue warning : `[Vue warn]: Invalid default value for prop "data": Props with type Object/Array must use a factory function to return the default value.`


rafaelpimpa:
Thanks, forget that one.

PR #218, opened 2017-08-02
add card component with collapsible

prograhammer:
If you look at Bulma documentation, they have an example of a Card that looks to be collapsible (even though of course the JS is not there). So I thought having collapsible ability in cards (just like panels) would be useful. I'm using this in my own project right now, and it's great. The shared collapsible functionality between cards and panels could be refactored to a mixin later.

Hope you like this, and keep up the great work on Buefy!


prograhammer:
screenshot1
screenshot2

prograhammer:
Forgot to do a feature branch. Closing and opening a new one...

PR #219, opened 2017-08-02
Add card component with collapsible

prograhammer:
If you look at Bulma documentation, they have an example of a Card that looks to be collapsible (even though of course the JS is not there). So I thought having collapsible ability in cards (just like panels) would be useful. I'm using this in my own project right now, and it's great. The shared collapsible functionality between cards and panels could be refactored to a mixin later.

Pull down and check out the docs page. Here's what the two examples on the docs page look like:  

screenshot1
screenshot2  

Hope you like this, and keep up the great work on Buefy!


src/components/card/Card.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+    <div class="card">
+        <header class="card-header" :class="{'is-collapsible' : collapsible}" @click="toggle">
+            <p class="card-header-title" v-if="header" v-html="header"></p>
+            <slot v-else name="header"></slot>
+            <a class="card-header-icon">
+                <span class="icon">
+                    <b-icon v-if="collapsible"
```
bartboy011: Make sure to import `b-icon` and register it below in case developers aren't importing the whole Buefy library.

prograhammer: @bartboy011 Ahh right...Done. See below. Thanks!

bartboy011: Perfect, thanks!

prograhammer:
@bartboy011 merged in latest upstream dev, resolved conflicts.

rafaelpimpa:
Hey I didn't have time to look at this yet but I'll take a look soon.

jtommy:
@prograhammer We're sorry for delay.
Your PR is great but it is very similar to Panel component and we don't provide to map all Bulma components as Buefy components. 
We have just made new component `Collapse` [2edac0816712855e3d1eb74a26aa9e130e3face9](url) that allow to use the Bulma style that you prefer.

PR #221, opened 2017-08-04
Add support for backend pagination to Table

cappuc:
This add support for backend paginated data to b-table.
I know that this can be done using separated b-pagination component but i think it's nice to have support directly in the b-table component.
I added 2 props to b-table:
- backendPaginated: Boolean that enable backend pagination
- total: total number of items (from backend)
If backendPaginated is true the number of items is read from total otherwise it's the length of data

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -116,7 +116,7 @@
             <div class="level-right">
                 <div v-if="paginated" class="level-item">
                     <b-pagination
-                        :total="newData.length"
+                        :total="newDataTotal"
                         :per-page="perPage"
                         :simple="paginationSimple"
                         :current="currentPage"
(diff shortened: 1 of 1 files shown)


src/components/table/Table.vue
```diff
@@ -192,12 +198,26 @@
...
+                    this.newDataTotal = value.length
+                }
+            },
+
+            /**
+             * When Pagination total change, update internal total
+             * only if it's backend-paginated.
+             */
+            total(newTotal) {
+                if (this.backendPagination) {
```
rafaelpimpa: Change to 
```javascript
if (!this.backendPagination) return

this.newDataTotal = newTotal
```

It's a good practice to return the function as soon as possible.

cappuc:
I changed the function as requested

rafaelpimpa:
Thanks!

PR #233, opened 2017-08-10
Moved sort $emit up, allowing custom sort to be used

Tirke:
There is currently a shortcoming of the customSort behavior : we don't have access to the sort order which is crucial if we want to implement a proper sort.
The only way to get the current sort order is by listening to the sort event, but this event is not emitted before the actual sort takes place.
This pull request simply move the $emit call before the actual sort happens, effectively solving the problem.
BUT : 
We may need some refactoring because some people may expect the event to be emitted after the actual sort ...

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -352,11 +352,11 @@
                         ? !this.isAsc
                         : this.isAsc = true
                 }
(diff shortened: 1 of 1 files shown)


rafaelpimpa:
Well, sort event is used mostly for backend sorting so I don't think we gonna have problem moving up, thanks.

PR #234, opened 2017-08-10
Added types (colors) to Switch component

Tirke:
I thought a type prop could be useful on `b-switch` so here it is ... 😃 
If it's not a good idea just close this.


jtommy:
@Tirke Yes, it could be useful! Thanks

PR #242, opened 2017-08-11
Change vuejs required version to 2.4+ (docs & readme)

iwan-wijaya:
`$attrs` & `inhseritAttrs` requires vue [2.4](https://github.com/vuejs/vue/releases/tag/v2.4.0)

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Browse the [online documentation here](https://buefy.github.io).
 
 ## Quick Start
 
-You need [Vue.js](https://vuejs.org/) **version 2.3+**.
+You need [Vue.js](https://vuejs.org/) **version 2.4+**.
 
 ### 1 Install via npm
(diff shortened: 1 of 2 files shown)


rafaelpimpa:
But when you use the built js is it still required?

iwan-wijaya:
Yes it is, I was using vue 2.3.3 for my project and did an upgrade to Buefy 0.5. input props like placeholder doesn't work anymore because of this:
https://github.com/rafaelpimpa/buefy/blob/dev/src/components/input/Input.vue#L16
https://github.com/rafaelpimpa/buefy/blob/dev/src/components/input/Input.vue#L27

In vue 2.3 placeholder will be applied to the wrong element (root element: `div.control`)

rafaelpimpa:
Alright, thanks for this!

PR #249, opened 2017-08-13
Fix sort and pagination errors on detailed table by removing the :key…

Tirke:
… prop on tr.

There is currently a conflict when the table use the `detailed` props for displaying a collapsible row under each row. The same keys are used on two different elements the real `tr` and the detailed one. 

The fix I PR'ed is to remove the `:key` prop, but in fact another potential fix that you might prefer is to use `:key="index + visibleData.length"`to make sure that there are no conflicting keys.

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -86,7 +86,6 @@
                         </tr>
 
                         <tr v-if="detailed && isVisibleDetailRow(row)"
-                            :key="index"
                             class="detail">
                             <td :colspan="columnCount">
                                 <div class="detail-container">


wanxe:
Hmm. Let me try something about that. Because if you do any request and refill the table you loose the index. I think that the best way is implement any internal index. On my projects, normally use the item.id or another unique keys. With this way, you never loose the index even is the request is refreshed.

Tirke:
My opinion is in line with my PR, keys are not always needed because the default behavior is efficient.
As stated on Vue doc : 

> This default mode is efficient, but only suitable when your list render output does not rely on child component state or temporary DOM state (e.g. form input values).

I don't really think that the detailed row relies on child state and/or temp DOM state but I might be wrong. It's quite common to just put the `key` prop whenever you use `v-for`but in my opinion we should be more cautious about our usage and only use it when necessary.

jtommy:
@wanxe You are the main developer of this feature. What do you think about @Tirke fix ?

wanxe:
@rafaelpimpa Was who attached the :key prop not me. I think that the better solution is attach the unique key via prop for the details. For example:

<b-table detailed :detailKey="item.id">

I have that on my custom table and working really well.
Don't know what you think about that.

jtommy:
@wanxe Best solution it is have a unique id as :key value. 
'detail-key' prop it's ok but it could be not enough because not only one field could be a unique key of data

wanxe:
@Tirke, this solution:
`:key="index + visibleData.length"`
isn't the best, because if you refetch your data or filter, lose the index. We need to generate an unique id. And the best way is set that via prop. 
Right now, I do not see any other way.
I hope this weekend send a PR with the change.

PR #251, opened 2017-08-14
Add extra slot to data table - allow level-left to be overwritten

tgosp:
Allow this number - http://d.pr/i/Msi7MB to be replaced with action buttons.
Here is how the table will look with action buttons - http://d.pr/i/lcF6uo

Please ping me if I have to do some extra changes here.


tgosp:
@rafaelpimpa - Please review this when you have time.
Ping @jtommy

jtommy:
@tgosp Thanks. We're updating docs as you can see. We'll review as soon as possible.

tgosp:
Thank you so much @jtommy 

I'm missing few more things around the data component. 
Expect more requests soon.

rafaelpimpa:
Hi @tgosp, thanks for contributing!

We gonna make it a lot simpler, just a simple named slot so the dev can add anything he/she wants, that's why this will be closed, thanks for the original idea!

PR #261, opened 2017-08-17
prevent bRadioButton to emit update event when it is not being clicked

### LINKED ISSUE #260: update event is emitted multiple times on single click on a bRadioButton
### Overview of the problem

**Buefy** version: 0.5.0
**Vuejs** version: 2.4.2
**OS/Browser**: various

### Description

When several bRadioButton is uesd on one v-model.

Clicking on one of them will cause other bRadioButton to emit an update event.

Expected: there should be only one update emitted from the bRadioButton actually being clicked.

Live example: https://www.webpackbin.com/bins/-KrjoDDJhGzIV-MxnReU

frantic1048:
This fixes #260

### DIFF
--- a/src/components/radio/RadioButton.vue
+++ b/src/components/radio/RadioButton.vue
@@ -47,7 +47,11 @@
              * Emit input event to update the user v-model.
              */
             newValue(value) {
-                this.$emit('input', value)
+                // only trigger input event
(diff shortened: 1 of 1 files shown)


rafaelpimpa:
Thanks!

PR #273, opened 2017-08-21
fix Autocomplete for SSR

LANSELOT:
fix Autocomplete for vue SSR **"ReferenceError: window is not defined"**


jtommy:
@LANSELOT Thanks!

Issue #284, opened 2017-08-27
ability to change pagination icons (rtl)


moeenzarrinfam:
### Overview of the problem
ability to change pagination icons
**Buefy** version: 0.5.1
**Vuejs** version: 2.4.2
**OS/Browser**: chrome/firefox

### Description
sometimes (for example in rtl pages) we need to change the pagination next and previous icons
![screenshot from 2017-08-27 12-48-04](https://user-images.githubusercontent.com/13137143/29748221-0d0d2412-8b27-11e7-8ed5-c5894f16b193.png)

jtommy:
@moeenzarrinfam About rtl there is PR to Bulma project https://github.com/jgthms/bulma/pull/650.
I think it's better a base solution that only single component feature.

moeenzarrinfam:
this is a very simple idea to solve this problem
you can mirror navigator items with CSS:
`
  .pagination-next , .pagination-previous{
    -moz-transform: scale(-1, 1);
    -webkit-transform: scale(-1, 1);
    -o-transform: scale(-1, 1);
    -ms-transform: scale(-1, 1);
    transform: scale(-1, 1);
  }
`

jtommy:
@moeenzarrinfam Good to know, thanks!
We'are waiting for Bulma progress about RTL to think something for all components.

Tovli:
Following @jtommy comment about RTL support
Is there a plan for this? maybe a constructor option?

jtommy:
@tovil as you can see bulma author is working on it https://github.com/jgthms/bulma/commits/rtl-support

Tovli:
That's great
thanks a lot.
I really need to improve my search skills :-(

PR #285, opened 2017-08-27
#284 changeable pagination icons added

moeenzarrinfam:
problem fixed by adding props to **b-table** and **b-pagination**
default values didn't changed

#284 
tnx

### DIFF
--- a/src/components/pagination/Pagination.vue
+++ b/src/components/pagination/Pagination.vue
@@ -1,10 +1,10 @@
 <template>
     <div class="pagination" :class="[order, size, { 'is-simple': simple }]">
         <a class="pagination-previous" @click="prev" :class="{ 'is-disabled': !hasPrev }">
-            <b-icon icon="chevron_left" both></b-icon>
+            <b-icon :icon="iconPrev" both></b-icon>
         </a>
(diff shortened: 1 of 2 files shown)


jtommy:
@moeenzarrinfam I think it's a good idea. Please resolve conflicting files.

moeenzarrinfam:
@jtommy resolved

jtommy:
@moeenzarrinfam I'm looking your commit and the problem is about 'both' prop to b-icon component. This prop is used to internal component to manage different pack names with the same icon name. 
Your commit allow to use all 'mdi' icons only and the current equivalent icon (see Icon component source)

moeenzarrinfam:
@jtommy oops!... I didn't know that 
last commit contains my idea to solve the problem .
please check it.
tnx

jtommy:
@moeenzarrinfam I think that it's only a workaround because now there are 'mdi' and 'fa' icon pack but in the future?
You should enable/disable both prop if the current icon is equals to default or not, or another solution but it have to general.

jtommy:
@moeenzarrinfam We thought about this PR and at the end most of the components (panel, table, datepicker, ..) using icons don't allow to change these. So we prefer follow this way, at the moment.
Anyway let me thank you for the support and the spent time :thumbsup:

PR #291, opened 2017-08-30
Make closing options configurable for "Modal"

jwahdatehagh:
Hi @rafaelpimpa - first of all thanks for this amazing library - we totally love it!

This PR enables passing passing an Array like `['escape', 'x', 'outside']` as the `canCancel` prop without changing it's original API. Passing a Boolean still works as expected.

Now you can e.g. prevent closing the modal by clicking on the background overlay but still enable closing it via escape or clicking on the X: `:canCancel="['escape', 'x']"`.


rafaelpimpa:
I like that, thanks a lot!

PR #296, opened 2017-08-31
Fixes the issue of the PR #249 on detailed table with pagination and sort

wanxe:
Created an unique :key based on the index and Date() for the detailed row.
@Tirke Please, can you tell me please if you can reproduce the same error now? Thanks so much!

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -89,7 +89,7 @@
                         </tr>
 
                         <tr v-if="detailed && isVisibleDetailRow(row)"
-                            :key="index"
+                            :key="index + new Date().getUTCMilliseconds()"
                             class="detail">
                             <td :colspan="columnCount">
                                 <div class="detail-container">


Tirke:
### 1.

I'm not entirely OK with the fact that we put a `key` because nobody clearly proved me that we need one. I didn't research it, I don't know if there are any benchmarks on the `:key` usage in Vue, but Vue docs clearly state that keys are not always needed, and you can really see that everything works great with no `key`.

---

### 2.

 I know that it's common practice to use Date() to generate unique ids in JS but I'm not really a big fan of that.
With that out of the way, if we want to keep the fix this way,  `new Date().getUTCMilliseconds()`
isn't really a good solution here. It produces int between 0 and 999, if we are unlucky we could fall in an easy trap :
```javascript
// index = 1 , new Date().getUTCMilliseconds() = 29
// .... Later
// index = 30, new Date().getUTCMilliseconds() = 0
```
💣 😢 
`Date.now()` is I think faster than `new Date().getUTCMilliseconds()` + it produce more unique ints.

Did I say I'm not a big fan of using Date for producing unique ids ? 😆

jtommy:
I am agree with @Tirke about `new Date()` but I'm not so sure about `key`.
Yes, it's not required but all depend on content of single detail.

wanxe:
Good explanation: https://vuejs.org/v2/guide/list.html#key
We aren't iterating directly over components.

So I'm with @Tirke . I prefer remove it to avoid the issues.

jtommy:
@wanxe we accept @Tirke PR and close that.

PR #309, opened 2017-09-08
Support setting native id in checkbox and native for in checkbox and …

tmepple:
I needed to get automated browser testing to work which required the ability to set the native `id` attribute and the corresponding label `for` attribute in a rendered checkbox.  Also added the ability to set arbitrary attributes on checkboxes using the `v-bind="$attrs"` as is done in `input`.

Also added the ability to set the label `for` attribute in `field` which allows users to click on the label name to give focus to the form element.

I guessed at the convention for property names you would like -- but please let me know if anything should be changed.

Thanks for Buefy btw... it's really great!!

### DIFF
--- a/docs/pages/documentation/form/checkbox/api.js
+++ b/docs/pages/documentation/form/checkbox/api.js
@@ -16,13 +16,6 @@ export default [
                 default: '—'
             },
             {
-                name: '<code>native-id</code>',
-                description: 'Same as native <code>id</code> also set as <code>for</code> on the label',
-                type: 'String',
-                values: '—',
-                default: '—'
-            },
-            {
                 name: '<code>true-value</code>',
                 description: `Overrides the returned value when it's checked`,
                 type: 'Any',
--- a/docs/pages/documentation/form/checkbox/api.js
+++ b/docs/pages/documentation/form/checkbox/api.js
@@ -16,6 +16,13 @@ export default [
                 default: '—'
             },
             {
+                name: '<code>native-id</code>',
+                description: 'Same as native <code>id</code> also set as <code>for</code> on the label',
+                type: 'String',
+                values: '—',
+                default: '—'
+            },
+            {
                 name: '<code>true-value</code>',
                 description: `Overrides the returned value when it's checked`,
                 type: 'Any',
--- a/src/components/checkbox/Checkbox.vue
+++ b/src/components/checkbox/Checkbox.vue
@@ -2,11 +2,14 @@
     <label class="checkbox"
         :class="{ 'is-disabled': disabled }"
         ref="label"
+        :for="nativeId"
         :disabled="disabled"
         :tabindex="disabled ? false : 0"
         @keydown.prevent.enter.space="$refs.label.click()">
         <input v-model="newValue"
             type="checkbox"
+            v-bind="$attrs"
+            :id="nativeId"
             :disabled="disabled"
             :name="name"
             :value="nativeValue"
@@ -20,10 +23,12 @@
 <script>
     export default {
         name: 'bCheckbox',
+        inheritAttrs: false,
         props: {
             value: {},
             nativeValue: {},
             disabled: Boolean,
+            nativeId: String,
             name: String,
             trueValue: {
                 type: [String, Number, Boolean, Function, Object, Array, Symbol],
--- a/src/components/checkbox/Checkbox.vue
+++ b/src/components/checkbox/Checkbox.vue
@@ -2,14 +2,11 @@
     <label class="checkbox"
         :class="{ 'is-disabled': disabled }"
         ref="label"
-        :for="nativeId"
         :disabled="disabled"
         :tabindex="disabled ? false : 0"
         @keydown.prevent.enter.space="$refs.label.click()">
         <input v-model="newValue"
             type="checkbox"
-            v-bind="$attrs"
-            :id="nativeId"
             :disabled="disabled"
             :name="name"
             :value="nativeValue"
@@ -23,7 +20,6 @@
 <script>
     export default {
         name: 'bCheckbox',
-        inheritAttrs: false,
         props: {
             value: {},
             nativeValue: {},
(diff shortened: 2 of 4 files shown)


src/components/checkbox/Checkbox.vue
```diff
@@ -2,11 +2,14 @@
...
         :class="{ 'is-disabled': disabled }"
         ref="label"
+        :for="nativeId"
         :disabled="disabled"
         :tabindex="disabled ? false : 0"
         @keydown.prevent.enter.space="$refs.label.click()">
         <input v-model="newValue"
             type="checkbox"
+            v-bind="$attrs"
+            :id="nativeId"
```
jtommy: You can remove it because `v-bind="$attrs"` add custom attribute and i think `:for` it's not required if checkbox is bounded by `'<b-field>`.

src/components/checkbox/Checkbox.vue
```diff
@@ -2,11 +2,14 @@
     <label class="checkbox"
         :class="{ 'is-disabled': disabled }"
         ref="label"
+        :for="nativeId"
```
jtommy: See the next comment

docs/pages/documentation/form/checkbox/api.js
```diff
@@ -16,6 +16,13 @@ export default [
                 default: '—'
             },
             {
+                name: '<code>native-id</code>',
```
jtommy: You can remove it. See checkbox component review

src/components/field/Field.vue
```diff
@@ -15,6 +15,7 @@
         props: {
             type: String,
             label: String,
+            labelFor: String,
```
jtommy: I don't know if it's ok or add `v-bind="$attrs"` to `<label>` tag (like checkbox component). 
At the moment custom attribute go to wrapper `<div>` but i don't think that it helps in this tag

tmepple:
Hi @jtommy,

Thanks for the very fast review.  I will implement all of your feedback but wanted to ask another question first.

To support automated browser testing (simulating clicking checkboxes) I need to target the `label` created by the `Checkbox.vue` template using a css selector.  I can remove the `for` attribute from the checkbox label as requested and instead access it as a child of the `Field.vue` `div`.  This would require me to add a new prop to `Field.vue` allowing the ability to set an id on the `div` or if you like I can use `v-bind="$attrs"` on the `div` to accomplish the same thing.  Are you ok with that?

jtommy:
@tmepple Why do you want add an id to Div? As you see from my review, you might add for prop to Field and apply it to Label tag.

tmepple:
Hi @jtommy, The reason is I'm pretty sure the `label` tag for `<b-checkbox>` is rendered by the template inside `Checkbox.vue` instead of relying on `Field.vue` like `<b-input>`, `<b-select>`, etc does.  Adding the `for` to Field will work great for `<b-input>`, etc but not for checkboxes.

If I pass the `label` and `for` props to the the encompassing Field around a checkbox, I end up with two rendered `label` tags for the same field which doesn't look good.

I guess rather than adding a new prop to render an id to the div inside Buefy's Field I can just wrap another div around the whole thing and put an id on that to target the children for automated testing.  Do you like that idea better?

jtommy:
@tmepple I'm sorry but where do you want add a div ? Anyway I think that Label into `Field` with `for` attribute works with checkbox

tmepple:
Hi @jtommy, I think an example would really help the discussion...

```html
<b-field>
  <b-checkbox name="user[agree]" native-value="true" v-model="agreeTOS">
    I agree to the Terms and Conditions.
  </b-checkbox>
</b-field>
```
Is rendered by `Checkbox.vue` as:
```html
<div class="field has-addons">
  <label tabindex="0" class="checkbox">
    <input type="checkbox" name="user[agree]" true-value="true" value="true">
    <span class="check"></span>
    <span class="control-label">I agree to the Terms and Conditions.</span>
  </label>
</div>
```
In the case above we wouldn't want to pass `label` or `for` as attributes to `<b-field>` because then by my testing two `<label>` tags would be rendered.  One by `Field.vue` and one by `Checkbox.vue`.  Instead, if we were rendering a `<b-input>` or `<b-select>` then I would use `<b-field label="Name"><b-input value="John Smith"></b-input></b-field>` which would work great.

I was thinking in addition to updating my PR making the other changes that you suggested above, if I could add an `id` prop to `<b-field>` so the div could be rendered as `<div class="field has-addons" id="user_agree">` then I could easily target any internal element for automated click-testing with css `#user_agree > label` or `#user_agree > label > span.check`.

Alternatively, if you don't want me to add the new `id` prop on `<b-field>` then I can wrap everything in another div as below:

```html
<div id="user_agree">
  <b-field>
    <b-checkbox name="user[agree]" native-value="true" v-model="agreeTOS">
      I agree to the Terms and Conditions.
    </b-checkbox>
  </b-field>
</div>
```
It's a little more verbose but it would also work as I could then target the rendered html with css like `#user_agree > div > label > span.check`.

Please just let me know... if you are not comfortable with my adding an `id` to `<b-field>` that's perfectly fine at least I have a workaround and can push another commit with the other approved changes.

jtommy:
Id to Field it's ok but you don't update nothing because all native attributes to Field are applied to  root div.

src/components/field/Field.vue
tmepple: Thanks @jtommy, I tried renaming this prop to `for` but since it's a reserved word (I think) the `        <label class="label" :for="for" v-if="label">{{ label }}</label>
` in the template refuses to build.  Should I leave it as `labelFor` or is there a way in Vue to escape the reserved word and use `for` anyway?

jtommy: @tmepple Yes, you are right! I don't like but i think it's the only solution.

src/components/checkbox/Checkbox.vue
```diff
@@ -24,6 +24,7 @@
             value: {},
             nativeValue: {},
             disabled: Boolean,
+            nativeId: String,
```
jtommy: Please, remove it

tmepple: Apologies for missing that and thanks for all of your work on this excellent project!!

jtommy: Thanks!

PR #316, opened 2017-09-12
Fix checkbox shrinking when text is very long

Tirke:
When you use `b-checkbox` with long labels in restricted width context, the checkbox shrinks.

Repro with the fix commented :
https://jsfiddle.net/Tirke/sz8Lpk6y/1/


jtommy:
@Tirke Thanks!

PR #318, opened 2017-09-12
Correct checked-rows prop name.

jivemonkey2000:
Quick one to correct a prop name in the docs.


jtommy:
@jivemonkey2000 Thanks!

Issue #288, opened 2017-08-27
Indirect `Field` component has unwanted `is-clearfix` class


yarbshk:
### Overview of the problem

**Buefy** version: 0.5.1
**Vuejs** version: 2.4.2
**OS/Browser**: Opera 46

### Description

In latest release you changed the `parentField` computed property in the `FormElementMixin` and it's pretty cool! However some elements remained unnoticed as for me.  

Using an `Input` component wrapped by an indirect `Field` component leads to unwanted `is-clearfix` class (see screenshot below):

![Screenshot](https://image.ibb.co/npXwbk/Screenshot_from_2017_08_28_00_15_51.png)

 Can you sync the `hasMessage()` computed property in the `Input` component? For example as follows:

```javascript
hasMessage() {
    return this.parentField.$data._isField && this.parentField.newMessage
}
```
Thank you a lot!

PR #321, opened 2017-09-14
Check if have any message prop from parent if it's a Field

yarbshk:
Fix the #288 issue. Check message prop from a parent if it's a (**in**)direct Field.


Tirke:
Can you use double negation `!!` over `Boolean()` for casting, as it is both equivalent and faster ?

yarbshk:
@Tirke thanks for reply and hint, of course i can.

PR #329, opened 2017-09-21
v-sortable directive and datakey for alternative to using index

jaybeecave:
not sure how to implement the v-sortable directive but it is a good feature.

I have put a pull request over before about the datakey prop to allow using a different loop key than index. This is **needed**. Some situations where the underlying list may change data, but the index remains the same we get unusual results. It allows for use of a primary key or something unique, which I think is very useful and it is a non breaking change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -212,7 +212,7 @@
     "arr-flatten": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/arr-flatten/-/arr-flatten-1.1.0.tgz",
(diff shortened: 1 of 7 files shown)


jaybeecave:
something still not right here. I am working on this.. not ready for merge yet sorry.......

jaybeecave:
Okay  this is fine. I was having a seperate issue where I was using the index from the props.index and not getting the original index, but rather the current index based on sort. This is different from previous versions, so It was unexpected behaviour, but not related to this PR.

jtommy:
@jaybeecave Thanks but we think there are too many changes in the PR.
Anyway we don't want to add any (possible) integrations with other lib directly in a Buefy component like v-sortable.

PR #350, opened 2017-10-04
Accept callback as component

ivalkenburg:
Able to use async components without throwing a vue warning.


jtommy:
@ivalkenburg Thanks!

PR #354, opened 2017-10-05
Handle tab in autoselect

tmepple:
Previously, pressing `tab` on an autocomplete field would move to the next tabbable element on the page but leave the dropdown open.  This PR closes the dropdown while permitting native `tab` handling to move to the next element.

Additionally, if there is a hovered option it will select it (the same as pressing `enter`) and move to the next element.  This provides for even faster data entry especially when used in conjunction with the `keep-first` option.


jtommy:
@tmepple It's good! Thanks!

tmepple:
Thanks for the extremely fast review @jtommy!

PR #387, opened 2017-10-31
expose Table's  currentPage as prop

### LINKED ISSUE #386: How to set initial currentPage for paginated b-table ?
### Overview of the problem

**Buefy** version: 0.5.4
**Vuejs** version: 2.5.2
**OS/Browser**: Various

### Description

I am trying to sync route param's pageNumber with b-table. e.g. Accessing `/rabbithouse/menu?page=20` shows the 20th page of menu data with `b-table`.

So far, I achieved showing the correct page of data with b-table with `backend-pagination` and set new route query & update data on `page-change` event. But the page number of `b-pagination` inside `b-table` is always `1` on first load. Even I directly open `/rabbithouse/menu?page=1204`, the table is correctly showing the 1204th page of data, while the page number showing is `1`.

### Expected behavior

I think we could expose the `currentPage` prop with a default value `1` to fix this.

frantic1048:
Fixes #386, enables controlling initial page number of b-table.


jtommy:
@frantic1048 Thanks!

PR #404, opened 2017-11-12
Ensure expanded prop works on tabs

chasegiunta:
Depending on where the tabs component is nested, the `expanded` prop may not work. This simple change takes one more step to try & ensure that. When trying to implement tabs into a panel (example: https://bulma.io/documentation/components/panel) I ran across this issue. I apologize for this being split up into two commits.


rafaelpimpa:
Thanks! No problem the number of commits, I'll squash :)

PR #406, opened 2017-11-13
Table detailed enhancements:

wanxe:
- Optional preselected opened details via prop: detailedRows.
- Adds detailKey (required when detailedRows is set).
- Fixes the detail row style with the tables.


jtommy:
Thanks @wanxe

Issue #423, opened 2017-11-19
When two or more detail rows are open, errors occur by sorting table headers or clicking pagination buttons


sanderheilbron:
Since Buefy version 0.6.1 issue #246 is back again.

**Buefy** version: 0.6.1
**Vuejs** version: 2.5.5
**OS/Browser**: MacOS 10.13.1 / Chrome 62

Although a detail-key is available, the following errors show up when opening the first detail row and sorting a column:
* [Vue warn]: It seems there are duplicate keys that is causing an update error. Make sure each v-for item has a unique key.
* [Vue warn]: Error in nextTick: "TypeError: Cannot read property 'key' of undefined"
* TypeError: Cannot read property 'key' of undefined

A second issue that occurs is that rows disappear after sorting columns.

rafaelpimpa:
@wanxe can you take a look?

wanxe:
Of course @rafaelpimpa I think is fixed on  #426

wanxe:
As you can see on the [video](https://mega.nz/#!UkoEEDSA!BdO7K9QBxXXI6E_XlCo-RH8ujsX4oTfWRZbIfSSeYoU) I can't reproduce the error, is possible to attach any fiddle? @sanderheilbron

fdrews:
Hello in my mind is not the correct solution to remove the :key from detail row. 

Possible solution is to add a pre or postfix to the keys. Here is the closed Vue issue [#7323](https://github.com/vuejs/vue/issues/7323) with the same explained problem.

Best regards Florian

wanxe:
Yes is a possible solution. Let me test it. Thanks sow much @fdrews :)

fdrews:
Your welcome and thank you for your great work. 

If you test it, you will get another errors for duplicate keys in the docs components.

fdrews:
Hello @wanxe i have detected the other duplicate key errors. The ApiView in docs/components have also a duplicate key. The duplicate key is responsible for all docs errors where the apiview have a title attribute in seperates components api folder.

Regards flo

wanxe:
Can you provide any link? I think is the new Vue version. Is more sensible with the component :keys. 
But you can do a PR if you feel confident 😄

fdrews:
If I understand correctly, there was no error message in an earlier version (< 2.5.10). But the error already existed. :D

wanxe:
[...]

PR #426, opened 2017-11-20
Fixes the table details issues #423

wanxe:
We don't need the :key on the detailed row because cause conflicts with the main <tr> loop.

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -92,7 +92,6 @@
                         </tr>
 
                         <tr v-if="detailed && isVisibleDetailRow(row)"
-                            :key="index"
                             class="detail">
                             <td :colspan="columnCount">
                                 <div class="detail-container">


sanderheilbron:
Unfortunately I noticed this fix is not released with version 0.6.2. When can we expect this fix to be released?

wanxe:
Can you bring me a codepen or fiddle with that error?

sanderheilbron:
The issue is that :key is still available on the detailed row in release 0.6.2.

```html
<tr v-if="detailed && isVisibleDetailRow(row)"
                            :key="index"
                            class="detail">
                            <td :colspan="columnCount">
                                <div class="detail-container">
                                    <slot name="detail" :row="row" :index="index"></slot>
                                </div>
                            </td>
                        </tr>
```

sanderheilbron:
@wanxe Can you confirm this fix isn’t released? Thanks!

wanxe:
Was fixed on my last PR but I think @rafaelpimpa added the :key again.

rafaelpimpa:
Hey sorry guys, I'll remove the ``key`` and add a comment to not forget!

sanderheilbron:
@rafaelpimpa Thanks!

PR #429, opened 2017-11-21
fix #428, make b-table react to currentPage prop

### LINKED ISSUE #428: B-table - how to change current-page from js?
### Overview of the problem

**Buefy** version: 0.6.1
**Vuejs** version: 2.5.3
**OS/Browser**: Any

### Description

According to https://github.com/rafaelpimpa/buefy/issues/386 - it is possible now to set **initial** current-page for `b-table` component.

But how can I set current-page on some js event?
For example, we have paginated list of rows and simple add form on the same page.
I want to set current-page to 1, after successful submit. But I can't see how can I do it.

Tried `.sync`, didn't help.
>:current-page.sync="page"

Thanks!

frantic1048:
Trying to fix #428 .

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -311,6 +311,14 @@
             */
             openedDetailed(expandedRows) {
                 this.visibleDetailRows = expandedRows
+            },
+
+            /**
+             * When prop currentPage updated
+             * react to it
+             */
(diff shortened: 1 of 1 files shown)


rafaelpimpa:
Thanks, will also adda sync option to currentpage :)

PR #439, opened 2017-11-25
Add support for "is-horizontal" in Field

electrolinux:
Add a horizontal prop to the Field component (along with a new FieldBody component to handle multiple controls in the slot.)

Let me know if you need more for reviewing this pr.

Here is a screenshot of the Field documentation page related to Horizontal
buefy-horizontal

### DIFF
--- a/src/components/field/Field.vue
+++ b/src/components/field/Field.vue
@@ -1,17 +1,33 @@
 <template>
     <div class="field" :class="[fieldType, newPosition, {
         'is-expanded': expanded,
-        'is-grouped-multiline': groupMultiline
+        'is-grouped-multiline': groupMultiline,
+        'is-horizontal': horizontal
     }]">
-        <label class="label" :for="labelFor" v-if="label">{{ label }}</label>
-        <slot></slot>
+        <div class="field-label is-normal" v-if="horizontal">
+            <label class="label" :for="labelFor" v-if="label">{{ label }}</label>
+        </div>
+        <template v-else>
+            <label class="label" :for="labelFor" v-if="label">{{ label }}</label>
+        </template>
+        <field-body v-if="horizontal">
+            <slot></slot>
+        </field-body>
+        <template v-else>
+            <slot></slot>
+        </template>
         <p class="help" :class="newType" v-if="newMessage" v-html="formattedMessage"></p>
     </div>
 </template>
 
 <script>
+    import FieldBody from './FieldBody.vue'
+
     export default {
         name: 'bField',
+        components: {
+            FieldBody
+        },
         props: {
             type: String,
             label: String,
@@ -21,6 +37,7 @@
             groupMultiline: Boolean,
             position: String,
             expanded: Boolean,
+            horizontal: Boolean,
             addons: {
                 type: Boolean,
                 default: true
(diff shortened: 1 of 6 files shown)


src/components/field/Field.vue
```diff
@@ -1,17 +1,33 @@
 <template>
     <div class="field" :class="[fieldType, newPosition, {
         'is-expanded': expanded,
-        'is-grouped-multiline': groupMultiline
+        'is-grouped-multiline': groupMultiline,
+        'is-horizontal': horizontal
     }]">
-        <label class="label" :for="labelFor" v-if="label">{{ label }}</label>
-        <slot></slot>
+        <div class="field-label is-normal" v-if="horizontal">
```
electrolinux: The "is-normal" class add the required `padding-top` for aligning the label to the control text

src/components/field/FieldBody.vue
```diff
@@ -0,0 +1,14 @@
+<script>
+    export default {
+        name: 'fieldBody',
+        render(h) {
+            return h('div', { attrs: { 'class': 'field-body' }}, this.$slots.default.map(v => {
+                // skip returns and comments
+                if (!v.tag) {
+                    return v
+                }
+                return h('div', { attrs: { 'class': 'field' }}, [v])
```
rafaelpimpa: Lets use template here instead of render function to keep a pattern.

electrolinux: I'll be happy to do that, if only I knew how...
The whole (and unique) purpose of this FieldBody component is to:
* enclose everything inside a `<div class="field-body">` tag
* wrap each `<slot/>` entries in a `<div class="field">`

Not doing the second point means we can't have more than one control in a horizontal field, while the [Bulma doc](https://bulma.io/documentation/form/general/#horizontal-form) show it is possible.

In fact, I've started without any FieldBody, using this template:

    <div class="field-body" v-if="horizontal">
        <div class="field"><slot></slot></div>
    </div>


But it doesn't works with more than one control in slot.

rafaelpimpa: What's the difficult on doing that with template syntax? Not sure if I got the point.

electrolinux: All I got with a template solution was:

    <div class="field-body">
        <div class="field">
            <input/>
            <input/>
            ...
        </div>
    </div

But what is really reeded is:

    <div class="field-body">
        <div class="field">
            <input/>
        </div>
        <div class="field">
            <input/>
        </div>
    </div>

And I really don't know how to do that with template only. 

Found this render solution on [stackoverflow](https://stackoverflow.com/questions/43834785/how-add-parent-for-each-slots-in-vuejs)

rafaelpimpa: You are right, this is not easily done without render functions, my mistake.

src/components/field/FieldBody.vue
```diff
@@ -0,0 +1,14 @@
+<script>
+    export default {
+        name: 'fieldBody',
```
rafaelpimpa: Update component name to ``bFieldBody``, and call like ``<b-field-body>`` :)

electrolinux:
Requested changes applied :)

rafaelpimpa:
Thanks!

PR #441, opened 2017-11-26
fix icon size when pack is font-awesome in input sizes

mikejavier:
captura de tela 2017-11-26 as 10 31 27

When I tried to use the input sizes with the 'font-awesome' pack, found this bug.
this happened because the fa-2x and fa-3x classes were being set when it was not needed, according to the bulma doc.

https://bulma.io/documentation/form/input/#sizes


rafaelpimpa:
I'll update it on FormElementMixin to fix every form element, thanks anyway!

PR #449, opened 2017-11-29
Datatable custom row + development improvements

bitkris-dev:
- **First commit** (73bcc6e): I've started to use Buefy and I've noticed I wasn't able to add a custom component as a row for `<b-table>`. This commit wants to solve this missing feature.
I've also added `.package-lock` to the `.gitignore` file, as best practice.
- **Second commit** (68f4130): After the fork, running the **dev** branch threw errors about a missing file for the tooltip examples. I had to delete the code causing the issue, but I didn't have the time to solve this too. **You can ignore it.**
- **Third commit** (83ff167): I'm working in production and it wasn't possible for me to add my fork as a `package.json` dependency without throwing errors. The problem was in the missing `/lib` directory which contains (as far as I know) all the compiled Buefy `.vue` components. I've removed it from `.gitignore` so any fork now contains also the compiled code ready to be used.


rafaelpimpa:
There are a lot of files in there getting re-commited, should fix that before I review.

Also, try to make small changes, easier to review and your PR will be merged faster.

bitkris-dev:
A lot of files getting re-commited? Sorry, but I don't see any file getting commited multiple times. Can you be more specific?

gkinsman:
Thanks for contributing!

It's usually a good idea to split changes like this into separate PR's, so that they can be evaluated and merged individually, and so one change doesn't hold up the rest :).

I suspect @rafaelpimpa means that the `/lib` folder shouldn't be included in source control, as it is a compiled artefact. If it's not always updated when changes are made, that will lead to drift between the lib folder and the source files, which becomes a bit of a nasty problem when using a CI pipeline to auto-build on merge, as the compiled files can't be updated at the same time :). The solution to using it as a local fork would probably be to build it yourself and include as a webpack static asset, until the master is updated. 

Also, the reason package-lock is source controlled ([docs](https://github.com/npm/npm/blob/v5.0.0/doc/files/package-lock.json.md)) is to make tracing of dependencies easier over time.

It looks like you've made some valuable contributions, but to make it easier on the maintainers and ensure your changes are included, it's nice to follow the [contribution guidelines](https://github.com/buefy/buefy/blob/dev/.github/CONTRIBUTING.md).

carlobeltrame:
@gkinsman Since I'd like to see this feature as well, I created a new PR #1127 containing just the first commit.

bitkris-dev:
> @gkinsman Since I'd like to see this feature as well, I created a new PR #1127 containing just the first commit.

Don't hope for it, they often ignore Pull Requests. I've stopped to contribute to Buefy (I did other contributions, always ignored, losing a lot of time) when I've notice this stubborn closure from the maintainers.
Also, they could've cherrypicked the first commit without leaving all those efforts suspended.

jtommy:
Close https://github.com/buefy/buefy/pull/1127

PR #459, opened 2017-12-04
quick and easy typo fixes for building docs on my ext4 fs

electrolinux:
Yep! 
You can't build docs on ext4 fs (or any other case sensitive fs, I suppose) without fixing this typos.
Hope this help.


rafaelpimpa:
It should be ``Taginput`` on every reference

rafaelpimpa:
Fixed

PR #475, opened 2017-12-14
fix, not click same page

BlackMix:


jtommy:
@BlackMix Thanks for PR but what's the issue ? Pagination emits the events but if the value is the same doesn't happen; anyway we could add that if check.
However`disabled ` attribute to pagionation link change the style and that's no good.

BlackMix:
u can remove disabled, my bad

But I tested and last time I could to click in same page e sent request for api
I need to do it

thx

PR #485, opened 2017-12-19
 Feature: Adds "label-classes" prop to adjust label styling in b-field

docnoe:
With the "label-classes" prop one can adjust the label of a <b-field> to fit the styles of the controls inside it.

E.g. if the b-field contains an input with size "is-small", the label is not vertically centered unless "is-small" is added to it, too. Doing so is not possible at the moment, but easy with the changes of this PR

### DIFF
--- a/docs/pages/components/field/Field.vue
+++ b/docs/pages/components/field/Field.vue
@@ -64,6 +64,14 @@
             <p>Add the <code>horizontal</code> prop for aligning label and control in horizontal forms.</p>
         </Example>
 
+        <Example :component="ExLabelClasses" :code="ExLabelClassesCode" title="Label classes">
+            <div class="tags has-addons">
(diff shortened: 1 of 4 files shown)


jtommy:
@docnoe What do you think about a new name of the prop? For example labelSize ?

Thanks for the support!

docnoe:
@jtommy We could add more props that describe different styles, but I think a more generic prop is more flexible. This way you can set any other class as well. If, for example, the label should be white, just adding `label-classes="has-text-white is-small some-other-class"` would add the respecting bulma class and anything you want.

jtommy:
@docnoe Ok, then we prefer custom-class prop name to keep consistency.
Thanks

docnoe:
@jtommy makes sense, changed it...

jtommy:
Thanks!

PR #490, opened 2017-12-21
Support to icon on Table label

BlackMix:
Hi guys.

thats example:

```html
<b-table-column field="time" icon="clock-o" width="20" centered>
  2m
</b-table-column>
```


BlackMix:
heh.. I got bug.. later I sent this pr again with fix

PR #492, opened 2017-12-21
fix #491 add isAsc parameter to custom-sort function of b-table

### LINKED ISSUE #491: custom-sort function in b-table is not getting `isAsc` status
### Overview of the problem

**Buefy** version: 0.6.1, 0.6.2
**Vuejs** version: 2.5.4
**OS/Browser**: various

### Description

I'm using `custom-sort` of `b-table` to sort a complex object list.

I combined `custom-sort` and `sortable` on `b-table`. When I click on the `sortable` column head, the table is sorted. But I click again on the column head, the table does not change, which will reverse on a sortable column without `custom-sort` function.

Finally, I found the `custom-sort` function [is missing a parameter](https://github.com/rafaelpimpa/buefy/blob/v0.6.2/src/components/table/Table.vue#L355) `isAsc` to determine whether the data should be sort ascend or descend.

### Expected behavior

The table should *be reversed* on the second click on the `sortable` head

### Actual behavior

The table *does not change* on the second click on the `sortable` head

frantic1048:
This PR add an `isAsc` parameter to `custom-sort`. Fix issue described in #491.

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -401,7 +401,7 @@
                 let sorted = []
                 // Sorting without mutating original data
                 if (fn && typeof fn === 'function') {
-                    sorted = [...array].sort(fn)
+                    sorted = [...array].sort((a, b) => fn(a, b, isAsc))
                 } else {
                     sorted = [...array].sort((a, b) => {
                         // Get nested values from objects


jtommy:
Thanks @frantic1048 !

PR #508, opened 2018-01-02
Adds a key into detailed rows with prefix #423:

wanxe:
- Removes console.log on ExAsyncData component example.
- Change the data source of the ExDetailedRow to data_test.json

### DIFF
--- a/docs/pages/components/table/examples/ExAsyncData.vue
+++ b/docs/pages/components/table/examples/ExAsyncData.vue
@@ -81,7 +81,6 @@
                         }
                         this.total = currentTotal
                         data.results.forEach((item) => {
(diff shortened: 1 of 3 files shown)


jtommy:
@wanxe Is it still valid ? We decide to don't add `key` attribute to detail row, isn't it ?

wanxe:
Too much time ago... let me check

wanxe:
I think that you can add the key with prefixes if you want. Also I was changed the source data of the example and some stuff... So I know that now it can conflicts with the new changes, but with this PR seems better to me.

jtommy:
Ok about example but at the moment no prefixes, thanks!

PR #509, opened 2018-01-03
Feature: Adds autocomplete slot support to taginput

gkinsman:
Brings feature parity between taginput and autocomplete by passing taginput's slots through to the underlying autocomplete.

image

image

### DIFF
--- a/docs/pages/components/taginput/Taginput.vue
+++ b/docs/pages/components/taginput/Taginput.vue
@@ -13,6 +13,12 @@
             </p>
         </Example>
 
+        <Example :component="ExTemplatedAutocomplete" :code="ExTemplatedAutocompleteCode" title="Templated Autocomplete" vertical>
+            <p>
+                Slots are available for autocomplete items and the empty message, like with the <router-link to="/documentation/autocomplete">Autocomplete</router-link> control.
+            </p>
+        </Example>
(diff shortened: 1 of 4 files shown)


jtommy:
@gkinsman Great idea! To be honest i'm thinking a way to avoid that code:
```
defaultSlotName() {
    return this.hasDefaultSlot ? 'default' : 'dontrender'
 },
 emptySlotName() {
      return this.hasEmptySlot ? 'empty' : 'dontrender'
 },
```
I don't like have a fake slot name, what do you think ?

gkinsman:
I agree, although without the dummy slot name, the replacement slot will always be rendered instead of the autocomplete default. One alternative would be to duplicate the default from autocomplete but I decided against that. 

There doesn't seem to be a way to tell a nested component like this whether it should use the default slot or not - except for this method. In that sense, I think using 'dontrender' makes some sense.

jtommy:
@gkinsman Thanks but it could be changed if we find an elegant way

yakovyarmo:
why can't you use something like:
`                <template v-if="hasDefaultSlot">
                    <template slot-scope="props">
                        <slot
                          :option="props.option"
                          :index="props.index" />
                    </template>
                </template>`

PR #510, opened 2018-01-03
feat: checkboxes can be disabled

jvmonjo:
Ability to disable checkboxes in the Table component


jtommy:
@jvmonjo Thanks for the support but it's no the best way to add that feature.
We can't do assumptions about property data name.

PR #513, opened 2018-01-05
Required is optional for Prompt

jhoff:
This makes the "required" attribute on a prompt input overridable through the inputAttrs object.

Our use case is that on a user action we want to prompt the user to input an optional comment.

Also, for what it's worth, according to [the arrow function docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Returning_object_literals), using `() => {}` actually returns undefined. I believe the intention is to have `inputAttrs` default to an empty object, so you need to wrap the object literal in parenthesis `() => ({})`. There are a few other instances of `() => {}` being used within the project but I'm not sure what their intention is.

### DIFF
--- a/src/components/dialog/Dialog.vue
+++ b/src/components/dialog/Dialog.vue
@@ -31,7 +31,6 @@
                                         v-model="prompt"
                                         class="input"
                                         ref="input"
-                                        required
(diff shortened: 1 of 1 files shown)


jtommy:
@jhoff Thanks!

Issue #500, opened 2017-12-29
Message/help text of horizontal field is placed in wrong element


Herteby:
### Overview of the problem

**Buefy** version: 0.6.2
**Vuejs** version: 2.5.13
**OS/Browser**:Linux Mint/Chrome

### Description

According to Bulma, message/help text for horizontal forms should be placed like this:
```html
<div class="field is-horizontal">
  <div class="field-label is-normal">
    <label class="label">Email</label>
  </div>
  <div class="field-body">
    <div class="field">
      <div class="control"></div>
      <p class="help">HELP TEXT</p> <!-- In the inner field -->
    </div>
  </div>
</div>
```
But in Buefy they are placed like this:
```html
<div class="field is-horizontal">
  <div class="field-label is-normal">
    <label class="label">Email</label>
  </div>
  <div class="field-body">
    <div class="field">
      <div class="control"></div>
    </div>
  </div>
  <p class="help">HELP TEXT</p> <!-- In the outer field -->
</div>
```
This causes major issues in horizontal forms.

### Steps to reproduce

```html
<b-field horizontal label="Label" message="helper text">
  <b-input/>
</b-field>
```
### Expected behavior

Bulma:
![image](https://user-images.githubusercontent.com/6474685/34433592-b800974a-ec80-11e7-9b34-1589c6ce5ba1.png)

### Actual behavior
Buefy:
![image](https://user-images.githubusercontent.com/6474685/34434031-9aec5c9a-ec83-11e7-9bbc-d3f7c2e993d1.png)

Herteby:
This workaround seems to work for now:
```html
<b-field horizontal label="Label">
  <b-field message="helper text">
    <b-input/>
  </b-field>
</b-field>
```

ghprod:
confirmed ..

example from docs https://buefy.github.io/#/documentation/field

![image](https://user-images.githubusercontent.com/1922652/34483656-80d776e2-eff2-11e7-8a17-2e234abb5278.png)

ghprod:
@Herteby wrapping with nested `<b-field>` seems like the best solution for now.

![image](https://user-images.githubusercontent.com/1922652/34483730-00bf4fd8-eff3-11e7-816e-5e27c81b93bc.png)


Thanks

rafaelpimpa:
Can you take a look @electrolinux ?

electrolinux:
Yep! Working on it right now...

electrolinux:
See PR: #521

PR #521, opened 2018-01-09
render the bfield component inside FieldBody (fix: #500)

electrolinux:
Instead of rendering a simple `<div class="field"/>`, the FieldBody now render the proper buefy Field component.


jtommy:
@electrolinux I don't see any difference, the issue isn't fixed.
For example: 
```
<b-field horizontal label="Subject" type="is-danger" message="Error">
      <b-input name="subject"></b-input>
</b-field>
```
The error message is on the right as original issue.

electrolinux:
Oops! Not even tested this :(

I've had just verified that the inner control place the error message correctly when it come from the control validation, for example.

So here is a new take on this. 
Also added a Field with message in Example.

PR #525, opened 2018-01-10
Added paginationSize on table

nateleavitt:
Right now I'm using pagination component in addition to the table component to allow for size customization. Would love a simple size option for the pagination at the bottom of the table.

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -217,7 +217,10 @@
                 default: 20
             },
             paginationSimple: Boolean,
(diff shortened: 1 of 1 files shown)


src/components/table/Table.vue
```diff
@@ -216,6 +217,7 @@
                 default: 20
             },
             paginationSimple: Boolean,
+            paginationSize: '',
```
jtommy: You have to set type of prop, `String`

PR #527, opened 2018-01-11
Update _notices.scss

ayouidaniel:
align-self: flex-start is missing for classes .is-top-left and .is-bottom-left


jtommy:
@ayouidaniel Thanks!

PR #581, opened 2018-02-05
Disable checkboxes programatically through custom method

jvmonjo:
Changed the PR #510 with this new one, which makes no assumption about client data and uses custom method


jtommy:
@jvmonjo Thanks! PR is great but to be honest i don't like the prop name. What do you think of `isRowCheckable` ? Did you have other proposals ?

jvmonjo:
I think `isRowCheckable ` makes total sense. I'll change it right now.

jtommy:
@jvmonjo Please resolve conflict and update api docs

jvmonjo:
Actually. Giving it a second thought. Shoudn't `isRowCheckable` return `false` if the checkbox is disabled?

I think it must be the other way around. It has to be a question which when you respond `true`it means that is disabled.

jvmonjo:
So maybe `isRowDisabled` ?

jtommy:
You might change default value. 
I don't like `isRowDisabled` because it's generic about row and not about checked mode.

jvmonjo:
Don't merge yet. I need to double check the new default.

jvmonjo:
I've just fixed the logic to the new prop default. I think it's now safe to merge if you want.

src/components/table/Table.vue
```diff
@@ -160,6 +160,7 @@
 </template>
 
 <script>
+/* eslint-disable */
```
jtommy: Why that ?

src/components/table/Table.vue
```diff
@@ -295,7 +296,7 @@
              */
             isAllChecked() {
                 const validVisibleData = this.visibleData.filter(
-                        (row) => this.isRowCheckable(row) !== true)
+                        (row) => this.isRowCheckable(row) == true)
```
jtommy: Is not enough only `this.isRowCheckable(row)` ?

jvmonjo:
Sorry about eslint flag. I had to do it to run de dev server. I've already remove it

src/components/table/Table.vue
jvmonjo: Sorry. I had to do it to be able to run my dev server without eslint complaining. I've already remove it.

src/components/table/Table.vue
jvmonjo: How would you do it? Feel free to rewrite it in a simpler way

jtommy:
Please check review and fix eslint rule error

src/components/table/Table.vue
jvmonjo: I get it. I already changed it

PR #589, opened 2018-02-07
Tabs Component - Emit change event only when tab is actually changed

adrlen:
Let me know if I need to change something :)


adrlen:
Rebasing buefy/adrlen dev branch

PR #598, opened 2018-02-13
update close button in modal

nikugogoi:
specifying button attribute `type` to `'button'`. Acts as a form submit button inside a form if not specified.


jtommy:
@nikugogoi Thanks!

PR #604, opened 2018-02-14
TagInput | add prop allowing to remove last tag when pressing backspace

adrlen:


gkinsman:
I wonder whether this should be a toggleable behaviour or whether it should just be the way the thing works. When would you ever not want to be able to remove a tag on backspace?

adrlen:
At first, I set the default value to `true` but then went back to `false`. I don't recall why actually, maybe not to broke the current behaviour.

jtommy:
@adrlen Ok, no breaking change. I'm thinking about prop name and if correct have a fixed char code.

jtommy:
@adrlen It's only a my personal idea but what do you think about a `removable` prop that enable removing by backspace and close tag icon ?

adrlen:
Then removing by backspace would be a default behavior since `removable` should be true by default I guess ?

I'm not sure I cant think of a case where you wouldn't want to be able to remove tag in an input. Since you already cant remove them with state `disabled`.

Maybe we could have prop `remove-on-keys` with backspace key by default ?

jtommy:
@adrlen Ok for `remove-on-keys`. 
With `removable` i mean a tag list without close icon and with these two props we can combine these, for example `removable` prop equals to true and `remove-on-keys` equals to empty array, in a way to have the current behaviour.

adrlen:
I hardly see a case where someone would override this prop actually but I made the changes 🙂

PR #606, opened 2018-02-15
Autocomplete | emit select event only if option selected actually change

adrlen:
For the moment, when no option is selected, `select` event is emitted on each input


apmontgomery:
this is important for me.  thanks!

jtommy:
@adrlen Thanks!

Issue #425, opened 2017-11-20
Feature Request: Loading component at the parent level


adrlen:
Would it be possible to have the loading component not overlaying all the page but just its parent ?

```
<div class="columns">
  <div class="column">
    First column
    <b-loading :active.sync="true" :canCancel="true"></b-loading>
  </div>
  <div class="column">
    Second column
  </div>
</div>
```

In this example, just the first column would be overlayed by the loader.

Thoughts ?

nick-beers:
I've found that the fixed position loader is a little overkill for simple use cases that @adrlen is bringing up as well. If there was a way to have an absolute positioned loader in some capacity vs the fixed position (additional class?) so that we can use a relative class on the parent to force the loader to only overlay the parent that should accomplish this request.

adrlen:
@jtommy Would you accept a PR on this ?

jtommy:
@adrlen Yes! If possibile no a breaking change and we'll review your PR. 
Thanks.

jtommy:
Close https://github.com/buefy/buefy/commit/f42665a338b442f91bd5551174e3214796bd3ac8

PR #609, opened 2018-02-15
Loading | Add prop to limit loading component to its container #425

adrlen:
PR regarding issue #425


jtommy:
@adrlen Great! Thanks a lot

PR #622, opened 2018-02-18
TagInput | add min width to input #613

### LINKED ISSUE #613: TagInput | Control field min-width
### Overview of the problem

**Buefy** version: [0.6.3]
**Vuejs** version: [2.5.13]
**OS/Browser**: MacOS/Chrome64

### Description

When having multiple tags in TagInput, it can occur that the control field has a small width limiting what is visible.

### Steps to reproduce

Add multiple tags in tag input without having the control going back on the line

### Expected behavior

Have better vision on what is typed

### Actual behavior

screen shot 2018-02-16 at 17 45 29

Just setting a `min-width` should resolve this. PR ?

adrlen:
Fix #613

### DIFF
--- a/src/scss/components/_taginput.scss
+++ b/src/scss/components/_taginput.scss
@@ -35,6 +35,7 @@
                 padding-bottom: 0;
                 border: none;
                 box-shadow: none;
+                min-width: 10em;
                 &:focus {
                     box-shadow: none !important;
                 }


adrlen:
Is it better to use `em` instead of `px` actually ?

jtommy:
@adrlen Height is in `em`, so i think it's better use the same unit.

adrlen:
fixed

rafaelpimpa:
Isn't this already inherited from ``input`` class?

adrlen:
It's set as `max-width: 100%;`

PR #632, opened 2018-02-21
Field | fix rounded with dynamic addons

adrlen:
When having addons dynamically displayed, rounded field behavior isn't correct.

Example : https://jsfiddle.net/j9jp0kq0/1/


jtommy:
@adrlen Thanks!

PR #641, opened 2018-02-23
Add support for classes is-* in tables

brunoosilva:
Look this the problem: https://codepen.io/brunoosilva/pen/LQBwex

Only `table` have the class `is-striped`

### DIFF
--- a/src/scss/components/_table.scss
+++ b/src/scss/components/_table.scss
@@ -140,8 +140,8 @@
                     &:not(:last-child) {
                         margin-bottom: 1rem;
                     }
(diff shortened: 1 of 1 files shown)


jtommy:
@brunoosilva Did you resolve it with `!important` ?

brunoosilva:
@jtommy Could, but I do not find a good solution. I think it's better to do this the way I did, so keep running what exists and also for different `row-status`, do you understand me ?

jtommy:
@brunoosilva Sure, I understand it.

PR #642, opened 2018-02-24
Update the link of docs/pages/Home.vue

smasato:
Update the link of github-btn.
By doing so, the number of stars will be displayed.


jtommy:
@smasato Thanks!

PR #644, opened 2018-02-24
Add clearOnSelect prop to Autocomplete.vue

raniesantos:
**REASON TO ADD THIS:**

In my project I have a search bar to find users where I use `b-autocomplete`.

In this case, autocomplete handles an array of user **objects**.

When the `select` event is triggered, the page goes directly to the profile of the selected user.

**Preserving/completing the input value is unnecessary in this situation, I just want to reset it to '' (empty string).** Also I think this is a fairly common use-case.

If I don't specify a `field` to be displayed from the object, autocomplete emits (input event) `undefined` as the new value.

*This causes errors in my code because I expect my search value to always be a string.*

### DIFF
--- a/docs/pages/components/autocomplete/api/autocomplete.js
+++ b/docs/pages/components/autocomplete/api/autocomplete.js
@@ -23,6 +23,13 @@ export default [
                 default: '<code>value</code>'
             },
             {
+                name: '<code>clear-on-select</code>',
+                description: 'Clear input text on select',
+                type: 'Boolean',
+                values: '—',
+                default: '<code>false</code>'
+            },
+            {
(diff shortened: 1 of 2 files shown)


rafaelpimpa:
Could you document this change as well?

raniesantos:
@rafaelpimpa I just want to say... the way the docs are written are genius. I didn't have to add any HTML table row markup.

rafaelpimpa:
Thanks! Hah, there are better ways to do the docs but it's working for now :)

raniesantos:
Crap, it just dawned on me that I have to wait for the next release to use the feature I added, haha.

PR #608, opened 2018-02-15
Add aria attributes to Dropdown and Dropdown item components #576

LucasSegersFabro:
Hi!
This is my first time contributing to a project.
This issue (#576 ) inspired me to study accessibility!
I just started to look into the components, but I think I can help!

Could someone please tell me if the changes I made in the Dropdown component are relevant? If so, I'd like to improve my knowledge and improve accessibility on other components (maybe improve dropdown one for keyboard usage)!

Next step is to try a screen reader.

Thank you for your time :)

Issue #576, opened 2018-02-02
Accessibility needs serious attention in every one of these components


marcysutton:
### Lacking Accessibility in This Component Library

I found this library searching for Vue.js component examples, and sadly it was built with little to no accessibility. This excludes users with disabilities and so many others simply because there is no keyboard or screen reader support.

### Description

This library would work much better for users if accessibility requirements like keyboard operability (reachable, operable, perceivable), focus management, screen reader labeling and text alternatives, programmatically determinable DOM relationships, color contrast, and more. 

Components needing attention:

- Collapse: Associate button with collapsible content and indicate its state. You could try aria-haspopup on the button, aria-controls pointing the button to the content, and aria-expanded to indicate state. 
- Dialog: There are no semantics in use, consider using `role="dialog"` or the `<dialog>` element. Disable background content with `inert` and/or use a focus trap to prevent navigating behind modal dialogs.
- Dropdown: The items inside are not keyboard interactive: anchor tags need `href` attributes to be reachable, or you could use `button` elements instead. Hover menus are completely inaccessible to keyboard users and shouldn't be used at all without a focus counterpart. The menu triggers and items inside need to be programmatically associated like the collapse component above.
- Form Checkbox/Radio: By hiding the native input, you remove the ability for a screen reader to announce that it is an input, and that it is checked/unchecked. Try using a different hiding technique that preserves the inputs for assistive technologies.
- Tabs: There are no tab roles for this component, and the `<a>` elements without `href` means nothing is focusable inside for keyboard users. I'd highly recommend reading up on accessible tabs to understand all of the requirements.

I could easily keep going–nearly every single one of your components was built without a care for accessibility. Quality software is a lot more than about how it looks–please consider the barriers you're putting in place for users of every one of these components.

### Solutions and resources
- Here are some accessible examples of patterns similar to some of yours: http://heydonworks.com/practical_aria_examples/
- Here's the ARIA Authoring Practices Guide, including semantics and navigation conventions for UI components: https://www.w3.org/TR/wai-aria-practices-1.1/
- Here's an article on accessibility in JavaScript apps, including UI component libraries: https://www.smashingmagazine.com/2015/05/client-rendered-accessibility/

### Steps to reproduce

Try using a keyboard or a screen reader to access any of these components.

### Expected behavior

You should be able to interact with all of the components without having to use a mouse or a trackpad. If you turn your screen off, you should still be able to use the components with a screen reader like NVDA or Voiceover.

### Actual behavior

[...]

PR #647, opened 2018-02-26
Add aria attributes to Tabs and TabItems components

mateuswetah:
Hi there!

Continuing the discussion in Ref. #576 and Ref. #608, I'm planning to help a bit with Aria attributes and keyboards navigation. Here, as a _first-try_, I add `role`, `aria-controls`, `aria-selected` and  `tabindex` to allow proper navigation and feedback for Tabs; and `role`, `aria-hidden` and an `id` for TabItem.

The `id` is the tricky part, I'm not sure if it goes well with the way you're planning your library. As I described in the docs, it would be an optional `props` for TabItem. Maybe we can be more specific labeling it `ariaId`? I'm not sure. Fact is, if you do plan on providing accessibility for other components, we would need to offer this ID solution in other components, as it is the way ARIA navigates through the accessibility tree.

I'm not an expert in the accessibility issue, but I'm planning on studying and implementing other cases if you're interested, as I'm using Buefy in a big project and this is a key feature for us. Thanks in anyways for all you've being doing here, it's great!

### DIFF
--- a/docs/pages/components/tabs/api/tabs.js
+++ b/docs/pages/components/tabs/api/tabs.js
@@ -90,6 +90,13 @@ export default [
                 type: 'Boolean',
                 values: '-',
                 default: 'true'
+            },
+            {
+                name: '<code>id</code>',
+                description: 'Optional id for <i>aria-control</i>',
+                type: 'String',
+                values: '-',
+                default: '-'
             }
         ]
     }
--- a/src/components/tabs/TabItem.vue
+++ b/src/components/tabs/TabItem.vue
@@ -1,6 +1,11 @@
 <template>
     <transition :name="transitionName">
-        <div v-show="isActive && visible" class="tab-item">
+        <div
+            v-show="isActive && visible"
+            class="tab-item"
+            role="tabpanel"
+            :id="id"
+            :aria-hidden="!isActive || !visible">
             <slot/>
         </div>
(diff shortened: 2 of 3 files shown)


LucasSegersFabro:
0144511f7adcebf5c8a2b951ba9fb6281036ca50 makes DropdownItem listen for isActive changes on its Dropdown parent to change tabindex, making items tab-accessible.

Circle CI tests not passing though. Think I'll have to rebase.

LucasSegersFabro:
Hi @mateuswetah !
What if 'id' has a default value of `this._uid` if it isn't set?
Ref: https://forum-archive.vuejs.org/topic/1635/how-to-set-a-unique-id-for-each-component-instance/7

mateuswetah:
@LucasSegersFabro, wooha, didn't knew about that, seems a good approach! 

I think it would allow handling the ARIA stuff more automatically to users. Still, I'm not sure how Buefy's team would like to deal with this. They may offer it more as a _ready-boxed_ solution, or expose attributes such as Bootstrap, which is a really complicated but a more appropriate way, in my opinion. 

Maybe @rafaelpimpa can give us any inside if he's planing to put this on roadmap, and if so, how we could help him. ;)

mateuswetah:
This is outdated and will be refactored latter.

Issue #656, opened 2018-03-01
b-autocomplete v-model not accepting number (int/float)


jeliasson:
### Overview of the problem
**Buefy** version: 0.6.3
**Vuejs** version: 2.5.7
**OS/Browser**: Windows 10/Chrome

### Description
When model is a number (int/float), I'll get the error ```TypeError: e.replace is not a function```. This seems to be related to the value being parsed to method ```escapeRegExpChars```, in which function .replace() is not available on int/float types. 
https://github.com/buefy/buefy/blob/dev/src/components/autocomplete/Autocomplete.vue#L264

What about having ```escapeRegExpChars``` return the orginal value, if type is int/float?

### Steps to reproduce
http://jsfiddle.net/95epuL5u/2/
### Expected behavior
```v-model``` to parse number.

### Actual behavior
 ```TypeError: e.replace is not a function```.

PR #657, opened 2018-03-01
Do not parse value with type number in helper escapeRegExpChars

jeliasson:
Method .replace() does not exist of type number.
https://github.com/buefy/buefy/issues/656


jtommy:
@jeliasson I think it's better manage that issue in Autocomplete component and not in the common function.

jtommy:
https://github.com/buefy/buefy/commit/4fbbc0a09603432970d5d32c7e18251cc0e7e096

PR #661, opened 2018-03-04
Convert a value to string before getting its length

andrzejkupczyk:
The `maxlength` property (character counter) is not working correctly when you add the number modifier to your `v-model` managed input. Effect:

```
<b-input type="number" v-model.number="age" maxlength="2"></b-input>
```
zrzut ekranu z 2018-03-04 13-37-37


andrzejkupczyk:
Dayum, this change does not fix anything.

PR #664, opened 2018-03-05
TagInput | fix onblur event if tag is added onblur

adrlen:
When leaving input not empty, it adds automatically a tag. However, the `onblur` event is emitted before adding this tag hence it doesn't have all the tags.

### DIFF
--- a/src/components/taginput/Taginput.vue
+++ b/src/components/taginput/Taginput.vue
@@ -173,10 +173,10 @@
             },
 
             customOnBlur($event) {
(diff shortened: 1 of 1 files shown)


adrlen:
It might be nice to find a way not to emit `onblur` event when removing a tag with close icon.

jtommy:
@adrlen Thanks! About removing a tag i think it's no easy at the moment

adrlen:
Thought about it for a minute but quickly realized it won't be easy ahah

PR #682, opened 2018-03-11
Table/disabled detailed

wanxe:
The `disabled-detailed` prop controls the visibility of the trigger that toggles the detailed rows.

- The prop supports an function to evaluate the visibility: 
By example: `
	disabledDetails: (row) => {
 		return [3, 5].includes(row['id'])
	}
`

### DIFF
--- a/docs/pages/components/table/api/table.js
+++ b/docs/pages/components/table/api/table.js
@@ -171,6 +171,13 @@ export default [
                 default: '<code>[]</code>'
             },
             {
+              name: '<code>disabled-detailed</code>',
+              description: 'Controls the visibility of the trigger that toggles the detailed rows.',
+              type: 'Function',
+              values: '—',
+              default: '<code>true</code>'
+            },
+            {
                 name: '<code>detail-key</code>',
                 description: 'Use an unique key of your data Object when use detailed or opened detailed. (id recommended)',
                 type: 'String',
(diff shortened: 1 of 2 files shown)


jtommy:
@wanxe Well done, thanks. What do you think to change the prop name ? hasRowDetail or an other name that is more like it.

wanxe:
@jtommy I think that the prop name need to be the opposite than 'has', may be something like `has-disabled-detailed`?

jtommy:
But disabled is quite trivial because that function allows to hide the detail (when detailed is true)

wanxe:
But if you pass a function that evaluates which trigger need to be hidden. The prop name need to match with the action that performs, I think. We need to think about it, suggestions are allowed 😸

jtommy:
@wanxe I think `has-visible-detailed` is ok, no ?

wanxe:
Ok, the prop is changed!

adrlen:
English isn't my primary language but `has-detailed-visible` seems more natural to me 😛

jtommy:
@adrlen Yes, or simple detail. For me there'arent problems, you can decide the correct name for syntax and meaning

wanxe:
Haha then? We have a winner? `has-detailed-visible`?

jtommy:
Ok! You can change it and I'll merge it

PR #683, opened 2018-03-12
Add value attribute to switch

smasato:
Add a `value` attribute that binds `nativeValue` to switch. The document has a `native-value` property, but it was not used in the switch component.


jtommy:
Thanks!

PR #684, opened 2018-03-12
Autocomplete enter and tab should act the same

jbender:
Right now pressing `enter` doesn't automatically close the dropdown, which feels like an oversight. As the `tabPressed` method appears to handle this as expected, I simply extracted that as a helper and called it in both.


jtommy:
@jbender Thanks for the support. 
I dont' think is a overisght because for me when no option are selected you can press for example "esc".
I think it's a common behaviour of autocomplete components.

PR #702, opened 2018-03-20
Add "fal" to values for "pack" to support font awesome light icons

adrianrussell:
Hi first contribution to a project, very minor one at that!

Not 100% sure it is worth it, but it would have helped me from a documentation point of view.


jtommy:
@adrianrussell Thanks!

PR #706, opened 2018-03-21
Fire modal cancel on escape only when modal is active

geoloep:
Right now the cancel method of a modal is called every time the user presses escape somewhere in the application. Even when the modal isn't even visible. Furthermore the onCancel callback function of the modal will be called, which is unnecessary and might lead to unexpected side effects. This PR makes sure the modal is active before if decides to cancel.


jtommy:
@geoloep Thanks

PR #707, opened 2018-03-21
doc added Object as parameter in tagsinput events

Apo-S:
Parameter for add, remove events, a ream can be not only String, but also an Object, if the v-model was an Array of Objects type


jtommy:
@Apo-S Thanks

PR #711, opened 2018-03-21
fix #710 switch with true/false values - regression from #671

### LINKED ISSUE #710: Update 0.6.4 break switch component
### Overview of the problem

**Buefy** version: [0.6.4]
**Vuejs** version: [2.5.15]
**OS/Browser**: Linux/Chromium,Firefox

### Description

After upgrading to 0.6.4, the switch component breaks. Downgrading to 0.6.3 resolves the issue.

### Steps to reproduce

1. update `package.json` with "buefy": "^0.6.4",
2. `npm install && npm run dev` (with webpack)
3. the switch stops working

### Expected behavior

Clicking on the switch component should toggle its state.

### Actual behavior

The switch "wobbles" but returns to the same state, instead toggling.

buefy-switch-issue

adrlen:


jtommy:
@adrlen This fix isn't enough

PR #714, opened 2018-03-22
Fix for duplicated visible days names

wafto:
Sometimes people like me use the same name for displaying custom days names, when 2 has the same key vue.js throws a warning for duplicated keys, the example calendar correspond to a Spanish one in which **Tuesday** is represented with a 'M' _(Martes)_ and **Wednesday** with a same 'M' _(Miércoles)_

screen shot 2018-03-22 at 4 53 12 pm


jtommy:
@wafto Thanks!

PR #737, opened 2018-03-29
Add stop in click for using inside b-table with selection

brunoosilva:


jtommy:
@brunoosilva Thanks

jtommy:
@brunoosilva We have to restore the previous behaviour. Maybe we could add a prop to resolve your issue.

Issue #700, opened 2018-03-19
Allow Datepicker to be set with a string value


rcubitto:
**Buefy** version: 0.6.3
**Vuejs** version: 2.5.7

### Description

I was trying to seed the datepicker with a stringy date like `2018-03-19` but it is failing.

According to the docs, the `v-model` property only accepts a `Date` type. Can we extend it to also accept Strings, and use the `date-parser` to convert it to a `Date` type?

If you are interested, I could give it a try and PR a solution. Thanks.

agriboz:
It would be great in my opinion.

jtommy:
@rcubitto Ok, but you have to pay attention about that function because its behaviour is to parse a string from input text and the pattern could be different from a start date (for example if it's coming from a server)

rapidexact:
Hi. Is anyone working on this feature ?

Sillanium:
Hi. Please check my solution https://github.com/buefy/buefy/pull/744

davidosuna1987:
Hello, I have the same problem...when I pass the date from server (Y-m-d string) I'm getting the following error: Expected Date, got String.
I don't know how to solve it.

I try to use defaultDateParser but nothing happens and I get the same error message.

I'm just a begginer with Vue and Buefy and maybe is my mistake.

Thank you!

rapidexact:
@davidosuna1987 Hi. You should pass instance of Date object instead of string to datapicker. Dateparser function only for parsing date that user typing.

davidosuna1987:
@rapidexact thank you for your answer. Ok then, if in my database I'm storing date with Y-m-d date format, how can I retrieve this data into datepicker component? That is my real question...

I don't know how to do this dinamically.

davidosuna1987:
@rapidexact But when the date fields are generated dinamically (I have a contacts table and the user can add more contacts uploading a csv file), I can't use a computed property, right?

That is the reason I would like to use database string dates into datepicker component...

Any idea?

Thank you again my friend!

davidosuna1987:
@rapidexact Unfortunately the last days I have been trying to do it but I have not succeeded. Could you give me some guidelines on how to do it?

Thanks again.

Sztukmistrz:
maybe it will help something...
    <b-datepicker
    :value='new Date(Date.parse("{{$datetimeField['value']->format('m-d-Y')}}"))'
        editable>
    </b-datepicker>

johnpuddephatt:
[...]

PR #745, opened 2018-04-03
Fix Datepicker. Date now can be set from a string

Sillanium:
Hi all. It is my solution for issue https://github.com/buefy/buefy/issues/700

### DIFF
--- a/src/components/datepicker/Datepicker.vue
+++ b/src/components/datepicker/Datepicker.vue
@@ -162,7 +162,7 @@
         mixins: [FormElementMixin],
         inheritAttrs: false,
         props: {
-            value: Date,
+            value: [Date, String],
             dayNames: {
                 type: Array,
                 default: () => {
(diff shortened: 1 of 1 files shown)


jtommy:
@Sillanium `dateParser` has an other scope as I described for https://github.com/buefy/buefy/issues/700

Sillanium:
@jtommy in my case date comes from server (MySQL DB). It works fine.
Maybe add "modelParse" method with default value be "dateParser" value ?

jtommy:
@Sillanium Yes maybe it could be a good way but I should think about names.

Sillanium:
@jtommy Ok. Tell me about your resolution, please. I can contribute

jtommy:
We should rethink the component to add somenthing about

elmatou:
I understand this has not been merged and the possibility to send Date as string is not yet implemented.

Any reason, since it is a quit old issue (#700) ?

Issue #753, opened 2018-04-09
BTimepicker with Clock


SharpSeeEr:
### Overview of the problem

**Buefy** version: 0.6.5
**Vuejs** version: 2.5.13
**OS/Browser**: Windows 10/Chrome 65.0.3325.181

### Description

Would you be open to implementing a timepicker with a clock interface rather than number dropdowns?  Something like this:
![image](https://user-images.githubusercontent.com/4459909/38520763-f4665556-3c00-11e8-9e60-68eddf94bc81.png)
![image](https://user-images.githubusercontent.com/4459909/38520796-0b3386dc-3c01-11e8-8cbc-5c68c018fbe6.png)

(Images captured from [Vuetify.js Docs](https://vuetifyjs.com/en/components/time-pickers))

I have already been working on it for my own use, but would love to have it included in buefy.  If so I'll finish it up as a buefy component in my fork.

SharpSeeEr:
My initial effort is committed on my fork in a clockpicker branch:
[Buefy Clockpicker](https://github.com/SharpSeeEr/buefy/tree/clockpicker)

Screen shot from the docs:
![image](https://user-images.githubusercontent.com/4459909/38649629-fc5056f8-3db4-11e8-8dcd-a303f9902549.png)

Inline:
![image](https://user-images.githubusercontent.com/4459909/38649691-63e0b6be-3db5-11e8-93e9-e91a58e93aef.png)

rafaelpimpa:
@SharpSeeEr I like that! Mind creating a PR?

SharpSeeEr:
Not at all!  I am still working on getting support for colors - `is-primary`, `is-info`, etc, and can create a pull request as is or after I get that in, whichever you prefer.

rafaelpimpa:
Submit when you feel it's ready :) thanks a lot

Ansty93:
Any update on this issue? I would rather to use a clockpicker on desktop version instead of dropdown.

jtommy:
@Ansty93 I'll try to reuse the PR to add clockpicker

SharpSeeEr:
@Ansty93 @jtommy I'll update the PR to the latest.  Maybe tonight, probably tomorrow (3/27).

Ansty93:
@SharpSeeEr Sweet! Iam really looking forward to it

SharpSeeEr:
I have updated the pull request - it's up to date with dev now.

PR #770, opened 2018-04-17
Clockpicker

SharpSeeEr:
This is a new Clockpicker component, as discussed in https://github.com/buefy/buefy/issues/753.  This is a full implementation with full docs and tests.

I will note that the scss does use some exact pixel values for padding, height, and width, contrary to the contributing guidelines for buefy.  This is because it was already very difficult calculating lengths and coordinates for the hand and clock face numbers without having to figure in relative values.

Please let me know if I need to make any changes.


jtommy:
@SharpSeeEr Quick note: could you change class names following Buefy style ? (for example main class `b-clock-picker` and wrap all into this ).
Anyway, great!

SharpSeeEr:
@jtommy I'll get those changes made ASAP!

Any idea why the codecov/patch is failing?  And how to potentially fix it?

SharpSeeEr:
I have made the requested changes, and added proper animation for the clock hand.

SharpSeeEr:
Ping!  I'm not sure what the process is for accepting pull requests, and I'm not trying to push the issue, I'm just excited to see this in the official distribution, mostly so I can start using it directly in my projects.  :)  I'll take no response as "We'll merge it when we are damn good and ready, stop pestering us!"

Thanks!

jtommy:
@SharpSeeEr Thanks for changes. We'll review as soon as possible.

jtommy:
@SharpSeeEr Color example is great but I think it's better have a `type` prop like other components.
Why do you import `TimepickerMixin` in that way ? I think you should use for example `import TimepickerMixinfrom '../../utils/TimepickerMixin`.

SharpSeeEr:
@jtommy I added the type prop to the Clockpicker and updated the docs to use that.  I initially didn't use the property because my preference for my projects is to use the Bulma classes, and only use a component prop when necessary.  But that is my preference, and I see the need to have this consistent across the library.  

I updated the import clause.  Not sure why I formatted it that way originally, but it now is simply `import TimepickerMixin from '../../utils/TimepickerMixin'`

While there I noticed the inline Clockpicker showed a 0 for the hour when 12 is selected, so I fixed that as well.

Please let me know if there is anything else I should change.

SharpSeeEr:
Any update on merging this?

SharpSeeEr:
I merged the upstream changes into my branch and resolved the conflicts.

jtommy:
@SharpSeeEr Thanks! 
We're testing new single components build so we have to think if add your component in the next release or after it.

jtommy:
@SharpSeeEr We're quite ready to review it but meanwhile you might resolve conflicts.

SharpSeeEr:
Updated pull request integrating current buefy/dev.

jtommy:
@SharpSeeEr Great! Thanks a lot!

PR #772, opened 2018-04-18
Update _pagination.scss

dailycommit:
original padding in bulma is 0.75em


jtommy:
@dailycommit I think we don't need of this change if is the default bulma size but simply remove the padding. Anyway I think it's too big because we have only arrows, isn't it ?

PR #781, opened 2018-04-26
Fix table mobile card view for IE11

rperki8:
margin-right: auto; does not work in IE11 (see [https://stackoverflow.com/questions/37534254/flex-auto-margin-not-working-in-ie10-11/37535548](url) and [https://stackoverflow.com/questions/31903734/ie11-flexbox-max-width-and-marginauto](url)) without width: 100%. Adding "width: 100%;" causes wrapping on long data though so instead using "justify-content: space-between;" gives the desired effect (left-justifying the label and right-justifying the data) without the quirks of margin*: auto;.

Screenshot before (IE 11):
image

Screenshot adding width: 100% (IE 11):
image

Screenshot removing margin-right: auto; and changing justify-content: space-between (IE 11):
image

Screenshot before (Chrome 65):
image

Screenshot removing margin-right: auto; and changing justify-content: space-between (Chrome 65):
image

### DIFF
--- a/src/scss/components/_table.scss
+++ b/src/scss/components/_table.scss
@@ -158,13 +158,12 @@
                     td {
                         display: flex;
                         width: auto;
-                        justify-content: flex-end;
+                        justify-content: space-between;
                         text-align: right;
                         border-bottom: 1px solid $background;
                         &:before {
                             content: attr(data-label);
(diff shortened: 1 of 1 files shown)


jtommy:
@rperki8 Thanks!

PR #788, opened 2018-04-30
Fix the name of the king in the north

Meechyst:


brandonpittman:
This was an important change.

PR #794, opened 2018-05-02
Add new parameter in b-taginput of validation

brunoosilva:
You can validate the value before adding it to the tag list


jtommy:
@brunoosilva Please resolve conflicts and I think it's better change name of prop to `beforeAdding` or similar

jtommy:
Thanks but please remove changelog commit because we have to decide when release this PR.

brunoosilva:
@jtommy When it is done new release with this feature ?

jtommy:
As soon as possible. Stay tuned!

PR #802, opened 2018-05-08
missing default for `$info` color

Oliboy50:
I'm not too sure about this `!default` keyword, I though it was here to allow override the value if it was set before... but it looks like i'm already able to override it right now...

besides, could someone tell me why we don't set `$info-invert` as well?

### DIFF
--- a/src/scss/utils/_variables.scss
+++ b/src/scss/utils/_variables.scss
@@ -4,7 +4,7 @@ $speed-slower: 250ms !default;
 $primary: #7957d5 !default;
 $primary-invert: $white !default;
 
-$info: #167df0;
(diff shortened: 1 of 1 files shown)


Cl3MM:
For those of you facing this issue, you can add the following postinstall temporary fix in the scripts section of your `package.json` file (make sure you have `sed` installed):

    {
      "scripts": {
        "postinstall": "sed -i 's/#167df0;/#167df0 !default;/' node_modules/buefy/src/scss/utils/_variables.scss"
      },
      "dependencies": {
        "buefy": "^0.7.0"
      }
    }

It helps when using an automated build pipeline.

PR #825, opened 2018-05-21
Added a searchbar in docs sidebar

bitkris-dev:
I've added an autocomplete input which takes the same data used to create the sidebar column structure to search a component based on user's input. The array is filtered to include only the **UI components** pages and does a  `this.$router.push` when the `@select` event is emitted from the autocomplete component (so, it works immediately on click). After the route push, the input is cleaned.

### DIFF
--- a/docs/components/TheSidebar.vue
+++ b/docs/components/TheSidebar.vue
@@ -1,5 +1,20 @@
 <template>
     <aside class="sidebar menu is-hidden-touch">
+        <b-field label="Search in docs">
+            <b-autocomplete
+                v-model="search"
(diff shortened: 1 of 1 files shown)


bitkris-dev:
I think would be nice to at least receive a sort of reply from the repo mantainer.
There are efforts and time involved behind pull requests and ignoring them repeatedly has the simple effect to turn possible contributors away. And I would like to see Buefy speed up in its evolution, as I like this framework a lot.

jtommy:
The site will be moved into an other repo using vue-cli 3 too

PR #834, opened 2018-06-02
updated material design icons package

qchenevier:
Currently, we cannot use all the materials icons, due to lack of update of the material icons dependency. The icons currently available are from [mdi 2.0.46](https://materialdesignicons.com/cdn/2.0.46/) whereas there are a lot more available in the last version: [mdi 2.2.43](https://materialdesignicons.com/cdn/2.2.43/). I propose to update this dependency.

The package has changed name on npm, it is now called `@mdi/font` so I changed paths accordingly.

It seems to build properly when I do `npm install` then `npm run build`, but I have not tested my PR more than that.

> Note: I'm a total noob using node & javascript so I might have missed something.

### DIFF
--- a/docs/assets/scss/main.scss
+++ b/docs/assets/scss/main.scss
@@ -26,8 +26,8 @@
 @import "~highlight.js/styles/atelier-cave-light.css";
 
 // MDI
-$mdi-font-path: "~mdi/fonts";
-@import "~mdi/scss/materialdesignicons";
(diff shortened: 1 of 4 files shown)


shevchenkodenis:
Is there any updates on this issue?

jtommy:
https://github.com/buefy/buefy/pull/975

PR #853, opened 2018-06-12
fix dateUTC

jlsneto:


jtommy:
@jlsneto Thanks but why?

jlsneto:
When I click on any date, it is always returned -1 day. I noticed that passing a "schedule" as a parameter causes the dateUTC to return the correct date.

jtommy:
Could you provide us an example (jsfiddle or codepen) about your issue ?

jlsneto:
https://jsfiddle.net/jlsneto/rv82msxw/

wopian:
Neither seem to be correct to me.

Timezone | Date | UTC Date | Result | Note
-------- | ---- | -------- | ------ | ----
UTC-12	 | 01:00, 13th | 13:00, 13th |  | Both dates are off by -1
UTC-6	 | 07:05, 13th | 13:05, 13th |  | This PR is showing the correct UTC-adjusted date
UTC-1	 | 12:18, 13th | 13:18, 13th |  | This PR is showing the correct UTC-adjusted date
UTC-0 | | | | Both correct
UTC+6	 | 19:10, 13th | 13:10, 13th |  | This PR is now incorrect and a day ahead
UTC+11	 | 00:13, 14th | 13:13, 13th |  | Both dates are off by +1

jlsneto:
Very good @wopian , so I guess not this constant dateUTC is not necessary. For if you return date.toLocaleDateString(), it works quietly.

jtommy:
@jlsneto I don't see any issue in your fiddle

jlsneto:
@jtommy probably its location.

damien-biasotto:
If I may say, I'm not sure that the last change will actually fix the issue.

I unfortunately work a lot with dates (with different timezones and DST, and UTC conversions) and I think the UTC conversion was the right way to go, I 'm using a similar solution (borrowed from https://github.com/dalelotts/angular-bootstrap-datetimepicker/issues/219). 

For example I have no issue from where I'm based (Australia/Sydney timezone), the day selected is the right one.

 I'll try to write some unit tests to highlight the actual behaviour.

jtommy:
Take a look at https://github.com/buefy/buefy/commit/a06eed4b2753d31a8082be68397b643b922e2cd8

PR #860, opened 2018-06-14
Add table fullwidth prop. Issue https://github.com/buefy/buefy/issues…

quimarche:
…/858.


jtommy:
@quimarche Buefy Table is full width yet

quimarche:
No, it isn't. There needs to be a way to add the is-fullwidth attribute to the table within the table-wrapper element.

Issue #855, opened 2018-06-13
b-input support scroll event?


iwata:
### Overview of the problem

**Buefy** version: [0.6.6]
**Vuejs** version: [2.5.16]
**OS/Browser**: Mac/Chrome

### Description

I wanna set `@scroll.native` for textarea scroll, but it didn't invoke any callback functions.

### Steps to reproduce

I think the case of a simple component.
1. input some text with multi-line, it will show scroll bar.
1. scroll up or down in the `textarea`
1. position zero continues to be displayed.
```
<template lang='pug'>
  section.content
    b-field(label='body')
      b-input(v-model='body', rows='3', type='textarea', @scroll.native='onScroll')
    .result.is-size-3 {{ position }} position
</template>
<script>
export default {
  name: 'InputTextarea'.
  data: () => ({body: '', position: 0}),
  methods: {
    onScroll({target}) {
      this.position = target.scrollTop
    }
  }
}
</script>
```

### Expected behavior

The position variable is changed and displayed.

### Actual behavior

Zero continues to be displayed.

service-paradis:
@jtommy 
I think this issue can be closed

PR #866, opened 2018-06-18
scroll event about b-input

iwata:
`textarea` of `b-input` can handle a scroll event.
ref. #855


iwata:
How about this PR?
Any advices?

PR #882, opened 2018-06-27
Clarify slot property

caleb-allen:
It is unclear in the docs that the `slot` property is necessary


jtommy:
@caleb-allen Yes there is no Slots tab in API section. Please add it and all slots into in the same way of autocomplete component. 
Thanks

caleb-allen:
@jtommy will do. Where can I find all of the slots options?

jtommy:
In the component code 😁

jtommy:
Please add `default` slot for Dropdown and DropdownItem component too.

PR #894, opened 2018-07-07
added [].includes for 3 types checks empty

BlackMix:


jtommy:
@BlackMix We have no full support for IE but in this way select doesn't work.

BlackMix:
hmm that's weird, here usually works this, but ok

Issue #912, opened 2018-07-17
Accessibility issue with Pagination and Table components


rwb7041:
### Overview of the problem

**Buefy** version: [0.6.6]
**Vuejs** version: [2.5.2]
**OS/Browser**: Windows OS - Chrome

### Description

There is an accessibility issue with the Pagination and Table components in Buefy 0.6.6. Basically, the next and previous page buttons in the Pagination component and the Detail button in the Table component do not have any sort of text for screen readers to pick up on. According to https://dequeuniversity.com/rules/axe/3.0/button-name?application=axeAPI, all buttons must have some kind of discernible name. I'll be submitting a pull request to fix this issue.

PR #913, opened 2018-07-17
Added aria-labels to buttons on table and pagnination to fix accessibility issue

rwb7041:
ref #912

### DIFF
--- a/src/components/pagination/Pagination.vue
+++ b/src/components/pagination/Pagination.vue
@@ -4,6 +4,7 @@
             role="button"
             href="#"
             class="pagination-previous"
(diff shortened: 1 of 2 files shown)


mateuswetah:
Hi @rwb7041 just passing here, I know this is an issue that must be discussed with more people, but just providing `aria-label`s like that won't be a good solution as it is not translatable. I think we need props to the components like pagination, tags, etc, where inner buttons would need to be labelled.

jtommy:
@mateuswetah  We have to discuss about but anyway it's no a good idea have fixed text.
Let's continue to discuss in the issue

PR #914, opened 2018-07-18
custom input state icon

maetad:


service-paradis:
This could also be used in MessageMixins.js

PR #915, opened 2018-07-18
Add code to set activeTab on created

ais-one:
Currently, tab item is only rendered on mounted.

It can be a long wait before user sees anything if a lot of things need to be loaded.

This patch allows active tab to be rendered on created first.


jtommy:
Why long wait ? It will be set active when all is ready

PR #927, opened 2018-07-26
Fix dropdown menu overflow and add line truncate

8bu:
Got a weird overflow when I'm using autocomplete input. Because my dropdown-item were too long and made the dropdown-menu have to expand to contain the item. I don't think it's necessary in this case (because this is autocomplete input). So I have to write some css to fix that.


jtommy:
Thanks!

PR #929, opened 2018-07-30
Loading: onCancel should not be called on ESC key when inactive

ankurk91:
Hi,

I found that `onCancel` callback is being called whenever user press ESC key even if loading is not active.

See this example -
```html
<b-loading :active="false" :on-cancel="onCancel"  :can-cancel="true"></b-loading>
```
```js
methods : {
 onCancel() {
        alert("User cancelled the operation by pressing ESC key")
 },
}
```
I can setup a full JSFiddle if you want.

### DIFF
--- a/src/components/loading/Loading.vue
+++ b/src/components/loading/Loading.vue
@@ -52,7 +52,7 @@
              * Close the Modal if canCancel.
              */
             cancel() {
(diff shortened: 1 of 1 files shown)


jtommy:
@ankurk91 Thanks

PR #948, opened 2018-08-17
Prevent a switch with multi-line text from shrinking

rhwilr:
If a switch has a long label and the text wraps to the next line, the background for the switch will start to shrink.
image

This PR will prevent this by adding a `flex-shrink: 0;`
image


jtommy:
Thanks!

PR #1006, opened 2018-09-21
Fix #940 - Radio or Radio-Button click event is triggered twice

### LINKED ISSUE #940: Radio or Radio-Button click event is triggered twice
### Overview of the problem

**Buefy** version: 0.6.6
**Vuejs** version: 2.9.6
**OS/Browser**: Any

### Description
When clicked on a radio or radio button, click event is emitted twice instead of once.

Similar issue: Switch | prevent click event to be emitted twice (https://github.com/buefy/buefy/pull/671)

### Steps to reproduce

Bind a native click event to a radio option:
@click.native="$emit('click-occured')"

### Expected behavior
Click event should occur only once

### Actual behavior
Click event is occurred twice

diboy2:
This happens because radio button is embedded in label. So, it triggers both label and radio button. To prevent this, I used click.prevent event in label.

### DIFF
--- a/src/components/radio/Radio.vue
+++ b/src/components/radio/Radio.vue
@@ -5,7 +5,8 @@
         :class="[size, { 'is-disabled': disabled }]"
         :disabled="disabled"
         :tabindex="disabled ? false : 0"
-        @keydown.prevent.enter.space="$refs.label.click()">
+        @keydown.prevent.enter.space="$refs.label.click()"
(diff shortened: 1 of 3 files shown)


jtommy:
@diboy2 It isn't working, are you sure is ok ?

jtommy:
@diboy2 I don't see any changes in your last commit

PR #1020, opened 2018-10-06
Improve tests for FieldBody.vue

afifsohaili:
Couldn't get `Field.vue`'s spec to 100% because the code coverage isn't telling me the correct lines my spec hasn't hit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
   },
   "lint-staged": {
     "*.{vue,js}": [
-      "eslint --fix",
+      "lint --fix",
       "git add"
     ]
   },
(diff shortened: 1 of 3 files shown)


jtommy:
@anteriovieira Could you take a look ?

jtommy:
@afifsohaili Please split in two PRs. One for test and one for FieldBody, thanks.
About test what's the issue using eslint ?

anteriovieira:
@afifsohaili follow the guidelines of @jtommy, doing this, it's okay.

afifsohaili:
> About test what's the issue using eslint?
It's not for tests. It's a general improvement that I discovered I could do as I was working on the project.

Using `lint --fix` will go back to the `lint` script declared in `package.json`. So it will also pass the options and the directories. That means, on precommit it will run `eslint --ext .js,.vue src docs` instead of just `eslint --fix`.

afifsohaili:
I'll resolve conflicts by end of tomorrow. Currently a bit preoccupied.

jtommy:
@afifsohaili Thanks

PR #1029, opened 2018-10-14
Unit Tests: upgrade vue-test-utils to latest version, correct deprecation warning

lordfuoco:
This PR aims to upgrade the version of vue-test-utils to the latest one (which include proper support for scopedSlot testing) and to remove all the warning generated by the now deprecated function `shallow`

There is only one touch to the the actual code in the dialog component `beforeMount` hook, where I wrapped the appendChild in a next tick function.
The code was working even without  the `nextTick` ( in the browser ) but only because of timing but it was not passing in the tests ( and is conceptually wrong: https://vuejs.org/v2/api/#beforeMount since in beforeMount the render function is not yet called $el can't be present yet. )

see this fiddle  for a better example: https://jsfiddle.net/0n5fmch3/ 
stringifying `this.$el`show empty object, while inside the `nextTick` the element exist ( is not possible to stringify it when is a real dom since is a circular structure )

Hope this helps!

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
   "devDependencies": {
     "@fortawesome/fontawesome-free": "5.3.1",
     "@mdi/font": "2.7.94",
-    "@vue/test-utils": "1.0.0-beta.15",
+    "@vue/test-utils": "^1.0.0-beta.25",
     "autoprefixer": "7.1.1",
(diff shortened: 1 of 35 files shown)


jtommy:
@lordfuoco great, thanks!

PR #1034, opened 2018-10-17
Improve tests for Field component

afifsohaili:
Per @jtommy 's comment [here](https://github.com/buefy/buefy/pull/1020#issuecomment-429667561), I'm splitting the Field and FieldBody spec improvements into two PRs.


jtommy:
@afifsohaili Thanks

Issue #1021, opened 2018-10-08
Addition of novalidate prop to b-input to toggle HTML5 validation checking


lukio3:
### Overview of the problem

Whilst the native HTML5 validity-checking function is great in most cases; there are edge cases where you want to disable this in place of doing your own validation on the blur/keyup events or similar. 

### Examples

1. On the 'number' type input; you may wish to have a step of "10" for example, but for purposes of validation you are happy for users to enter a value to a higher specificity. E.g.

[stackoverflow example](https://stackoverflow.com/questions/41522971/html-input-number-step-without-validation#comment78063820_41522971)

2. You may be using input fields in a group; and therefore the default render/placement of the HTML 5 message causes an issue, example:

input group:

![image](https://user-images.githubusercontent.com/21043078/46599392-2873de00-cadf-11e8-86da-dea4c3b05b7b.png)

after validation triggered: 

![image](https://user-images.githubusercontent.com/21043078/46599395-2f025580-cadf-11e8-88d3-464623a12d62.png)

### Expected behavior

I think it would be good idea to add a "novalidate" prop to the b-input component that toggles the checkHTML5validity function in the formElementMixin.

The "novalidate" naming convention would be consistent with the native HTML5 input box.

I have tried to hack this myself but cannot figure out a simple place to toggle the HTML5 validity checking. If you have any ideas on how I can get started i'm happy to open a PR.

I don't think this should be a global config for Buefy as you may want the HTML5 validation on some elements but not on others.

Let me know what you think

Cheers

Luke

jtommy:
@lukio3 Your idea is correct but it could be good no display error when there is a form with `novalidate`.
If you want you might make a PR in this way

PR #1056, opened 2018-10-26
prop to toggle HTML5 validation checking

lukio3:
Regarding: https://github.com/buefy/buefy/issues/1021

This change simply places an optional prop on FormElementMixin.js to allow/disallow the running of the CheckHtml5Validity() method (on blur).

In my specific case I needed this to be selectable by component rather than at the <form> level; as would be the case if we went for a solution akin to using 'novalidate', natively.

Anyway, suggestions welcome on what else I need to do to get this merged.

Luke


src/utils/FormElementMixin.js
```diff
@@ -10,7 +10,8 @@ export default {
         iconPack: String,
         // Native options to use in HTML5 validation
         autocomplete: String,
-        maxlength: [Number, String]
+        maxlength: [Number, String],
+        useHtml5Validation: Boolean
```
jtommy: Add as default the config value

src/utils/FormElementMixin.js
```diff
@@ -85,7 +86,9 @@ export default {
         onBlur($event) {
             this.isFocused = false
             this.$emit('blur', $event)
-            this.checkHtml5Validity()
+            if (this.useHtml5Validation || config.defaultUseHtml5Validation) {
```
jtommy: Remove config value check.
Are you sure is it enough ? Isn't it better add new prop check into checkHtml5Validity method?

PR #1057, opened 2018-10-27
Implemented indeterminate state for checkable table

imarlovic:
Set the header checkbox on a checkable table to indeterminate state if some (but not all) checkable rows have been selected.

buefy-table-checkbox-indeterminate


imarlovic:
Could I get some advice on how to approach fixing the code-coverage check failures?

Issue #973, opened 2018-09-06
Taginput doesn't add tags on mobile


parkersprouse:
### Overview of the problem

**Buefy** version: [0.6.6]
**Vuejs** version: [2.5.17]
**OS/Browser**: Chome 68.0.3440.91 on Android 8.1.0

### Description

I can't seem to be able to add any tags when testing my site on mobile. On my PC I can type in the input field and hit either `enter` or `comma` and the tag gets added, but hitting either of those on mobile does nothing. `enter` just moves to the next field while `comma` just types a comma in the field. I can't find a way to get the tag to actually submit on mobile.

williamabbott:
+1

williamabbott:
I think the issue is, at least on android chrome browser, if there are fields after the current one, it will show the user the tab key (keycode 9) instead of return, therefore Buefy not adding the keycode (Its only looking for enter 13 and comma 188)

jtommy:
@williamabbott I merged your PR but i don't think it resolves the android issue.

williamabbott:
Maybe when the conditions are met we can display a "Add tag" header which when clicked, adds the typed characters as a tag. I've tried to detect the Android tab key with no luck and this could just be a fool proof and clear way to do it.

tycrimm:
#1065 breaks the ability to tab from a TagInput to another form component for the sake of the mobile fix. Reverting the PR allows the user to once again tab between input components on a desktop. Can we find a better workaround for mobile so we don't break the desktop experience?

williamabbott:
Agreed, it's not a viable solution, I'd really like to see a "add 'typed term'" option appear first in the list as the user is typing, I was working on a pr for it:

![screen-recording-2019-02-26-at-19 44 59](https://user-images.githubusercontent.com/5918271/53438323-f583a580-39ff-11e9-9c8c-6b3349c56d01.gif)

jtommy:
@tycrimm you're right but anyway you might override default keys using `confirm-key-codes` prop at the moment

service-paradis:
We would probably need to replace the use of deprecated `KeyboradEvent.keyCode`:
https://github.com/buefy/buefy/issues/1485#issuecomment-526783899

jtommy:
Close #1940

PR #1065, opened 2018-11-05
Added tab keycode to confirmKeyCodes in Taginput.vue

williamabbott:
I think this resolves the issue. #973


jtommy:
@williamabbott What issue has been fixed ?

williamabbott:
See referenced issue.. Taginput on mobile doesnt add the tag. When in android for example the only confirm button you get on the keyboard actually tabs to the next field instead of entering. I added tab keycode.

williamabbott:
(when 'add new' is active)

tycrimm:
This breaks the ability to tab from a TagInput to another form component for the sake of the mobile fix. Reverting the PR allows the user to once again tab between input components on a desktop. Can we find a better workaround for mobile so we don't break the desktop experience?

PR #1073, opened 2018-11-08
Added an option to enable the use of the fontawesome 5 svg component

rhwilr:
This PR adds support for the [Font Awesome svg component](
https://www.npmjs.com/package/@fortawesome/vue-fontawesome).

With Font Awesome 5 they provide a vue component to render icons as svg. This allows us to only import the icons we need ans use tree shaking to reduce the bundle size.  

With this PR, I added a `defaultIconComponent` config that is used to render the icon, instead of the `<i>` tag.

```js
import Vue from 'vue'
import Buefy from 'buefy'
import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'

Vue.use(Buefy, {
  defaultIconPack: 'fas',
  defaultIconComponent: FontAwesomeIcon
})
```

### DIFF
--- a/src/components/icon/Icon.vue
+++ b/src/components/icon/Icon.vue
@@ -5,7 +5,7 @@
             :class="[newPack, newIcon, newCustomSize, customClass]"/>
 
         <component
-            v-if="useIconComponent"
+            v-else
             :is="useIconComponent"
             :icon="[newPack, newIcon]"
             :size="newCustomSize"
--- a/src/components/icon/Icon.vue
+++ b/src/components/icon/Icon.vue
@@ -1,6 +1,13 @@
 <template>
     <span class="icon" :class="[newType, size]">
-        <i :class="[newPack, newIcon, newCustomSize, customClass]"/>
+        <i
+            v-if="!useIconComponent"
+            :class="[newPack, newIcon, newCustomSize, customClass]"/>
+
+         <component
+            v-if="useIconComponent"
+            :is="useIconComponent"
+            :icon="[newPack, newIcon]"/>
     </span>
 </template>
 
@@ -25,17 +32,9 @@
              * internal icons are always MDI.
              */
             newIcon() {
-                if (!this.both) {
-                    if (this.newPack === 'mdi') {
-                        return `${this.newPack}-${this.icon}`
-                    } else {
-                        return `fa-${this.icon}`
-                    }
-                }
-
                 return this.newPack === 'mdi'
                     ? `${this.newPack}-${this.icon}`
-                    : `fa-${this.getEquivalentIconOf(this.icon)}`
+                    : this.getFAIconOf(this.icon)
             },
             newPack() {
                 return this.pack || config.defaultIconPack
@@ -77,9 +76,19 @@
                     case 'is-large': return largeSize
                     default: return defaultSize
                 }
+            },
+            useIconComponent() {
+                return config.defaultIconComponent
             }
         },
         methods: {
+            getFAIconOf(value) {
+                if (this.useIconComponent) {
+                    return this.getEquivalentIconOf(this.icon)
+                }
+                return `fa-${this.getEquivalentIconOf(this.icon)}`
+            },
+
             /**
              * Equivalent FA icon name of the MDI.
              */
--- a/src/components/icon/Icon.vue
+++ b/src/components/icon/Icon.vue
@@ -95,6 +95,11 @@
              * Equivalent FA icon name of the MDI.
              */
             getEquivalentIconOf(value) {
+                // Only transform the class if the both prop is set to true
+                if (!this.both) {
+                    return value
+                }
+
                 switch (value) {
                     case 'check': return 'check'
                     case 'information': return 'info-circle'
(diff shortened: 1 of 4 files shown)


shazha:
There's another way to use SVG icons with Font Awesome 5 and no need to change `Buefy`'s code. Add the following code into your `main.js` file:
```
import { library, dom } from '@fortawesome/fontawesome-svg-core'
dom.watch()
```
Then the only thing you need to do is importing icons used by Buefy components. For instance, import `angle-left` and `angle-right` if you are using `Datepicker`
```
import { faAngleLeft, faAngleRight } from '@fortawesome/free-solid-svg-icons'
library.add(faAngleLeft, faAngleRight)
```

rhwilr:
Nice! I did not see that when looking through the Font Awesome 5 documentation. 
This solves the problem for me. I'm leaving this PR open though, maybe this feature is still something we want in buefy?
I think we should definitely mention `dom.watch()` in the buefy documentation for all the FA5 users.

cesasol:
I did create a replacement bc I do use the vue component, since on nuxt the dom watch can be really expensive on resources. This merge could really facilitate the replication of my implementation for more use cases.

francoism90:
@rhwilr Think this should be merged, this way no one has to set `dom.watch()` and it will work OOB.

src/components/icon/Icon.vue
```diff
@@ -25,17 +34,9 @@
...
-                        return `${this.newPack}-${this.icon}`
-                    } else {
-                        return `fa-${this.icon}`
-                    }
-                }
-
                 return this.newPack === 'mdi'
                     ? `${this.newPack}-${this.icon}`
-                    : `fa-${this.getEquivalentIconOf(this.icon)}`
+                    : this.addFAPrefix(this.getEquivalentIconOf(this.icon))
```
jtommy: I think it isn't correct before when pack is FontAwesome you're adding always the equivalent icon.

jtommy: Why are you always loading `getEquivalentIcon` ?

rhwilr: I'm not sure I get what you mean. I only added `addFAPrefix`. The call to `getEquivalentIconOf` was there before.

rhwilr: Ah I see, you are asking why I removed the section L28-L35.
I remember removing it because I could not figure out what this does. In fact, it caused problems with icons that are named differently in FA.

For example, the Message has the `both` prop set to true and uses the `information` icon. But this icon is named `info` in FA. So why is this not renamed to the correct icon-name?

jtommy: Because the project uses mdi name and for internal icons we load the equivalent.

rhwilr: Exactly, but my point is that the equivalent icon is Not loaded when `both = true`. 
My change fixes this already.

rhwilr:
What is the state of this PR? Do you still want me to make some changes?

jtommy:
Check my review please

rhwilr:
I just went over all my changes again, and I still think I made the right change. :confused: 

Maybe I can try to explain my reasoning for removing the `both` check again:
If the used pack is `mdi`, the returned icon is the same regardless of the `both` prop.
If the pack is `fa` the correct FontAwesome icon should be returned. Since buefy uses the mdi name for icons in the code, we may have to rename the icon sometimes if the mdi name does not match the FA mane.

For example:
The [Input](https://github.com/buefy/buefy/blob/dev/src/components/input/Input.vue#L138) component uses an `information` status icon. But FA does [not have](https://fontawesome.com/icons?d=gallery&q=information) an icon called `information`. So we have to rename it to `info-circle` in the `getEquivalentIconOf` method.

Now the problem is that the Input component has the `both` [prop set for the status icon](https://github.com/buefy/buefy/blob/dev/src/components/input/Input.vue#L44).
Thus, the `information` icon will not be renamed to `info-circle` and no icon is displayed.

This is why I removed the check for the `both` prop. It makes it impossible to display status icons in inputs.
So ideally the `both` prop could be removed altogether.

Hopefully, I explained my reasoning well enough.
I'm sorry if I caused any trouble. This was not my intention. Please let me know what I've missed.

jtommy:
@rhwilr You're right but you should think about `bIcon` component.
Using your code I cannot use a FA icon called `information` because it will be changed to `info-circle`; at the moment there isn't a FA icon `information` but in the future there could be differents `info-circle` and `information` icons.

rhwilr:
Well, I would say it is very unlikely that that will happen and if that is the case, one could just use the FA component directly instead of wrapping it in bIcon.
But it this is a requirement to get this PR accepted I could add an `exact` prop that will skip any transformation done to the icon name.

jtommy:
Why don't you want maintain the same logic using both prop ? :)

jtommy:
@rhwilr `both` prop (maybe is not semantically correct) is only for internal use and the user (developer) shouldn't use it.
You might change the name of prop but the default value should be the same otherwise you'll introduce a breaking change and a unnatural behaviour for the user in my opinion as I told you above about `both` prop.

jtommy:
@rhwilr Thanks a lot!

MannikJ:
I am not getting it. Is it already part of any release? I am still facing reactivity issues with icons and font-awesome...

Issue #577, opened 2018-02-03
Close autocomplete dropdown programatically?


taylorhoward92:
### Overview of the problem

**Buefy** version: [0.6.3]
**Vuejs** version: [2.5.13]
**OS/Browser**: Chrome 64 on OS X

I am using the Autocomplete component for search suggestions. The idea is that a user can either select a suggestion from the autocomplete dropdown or can just do a free-text search. Then I use Vue router push to show the content for the suggestion, or a search results route for free-text search.

To allow submission of free-text search I am capturing the `@keydown.native.enter` event and taking appropriate action, including `blur()` on the input, but although the input gets deselected, the autocomplete dropdown menu still shows. I need a way to hide the dropdown menu or an alternate strategy for what I'm trying to achieve.

jtommy:
@taylorhoward92 There isn't a method, maybe only `setSelected` but I think it isn't ok for you. 
Anyway you might set `isActive` data to false, `this.$refs.myAutocomplete.isActive = false`

GarethTIGFreight:
I _sort of_ had a solution but with b-taginput with autocomplete but you should be able to adapt the code to work. See the *untested* code below. On the @typing event we filter the autocomplete results and append the filtered results with what you've typed. 

Alternatively you can use _ref_ on the autocomplete and directly access the inputted value (?). 

Good luck.

```
<template>
    <section>
        <b-field label="Insert a label">
            <b-taginput
                v-model="attr"
                :data="filteredTags"
                autocomplete
                placeholder="Insert a placeholder"
                @typing="getFilteredTags">
            </b-taginput>
        </b-field>
    </section>
</template>

<script>
    export default {
        name: 'meta-attribute-field',
        props: {
            def: {
                required: true
            },
            attr: {
                required: true
            }
        },
        data () {
            return {
                filteredTags: this.def.options,
                isSelectOnly: false
            }
        },
        methods: {
            getFilteredTags (text) {
                this.filteredTags = this.def.options.filter((option) => {
                    return option
                        .toString()
                        .toLowerCase()
                        .indexOf(text.toLowerCase()) >= 0
                })
                this.filteredTags.push(text)
            }
        }
    }
</script>
```

williamabbott:
Would you accept a pull request to have something like "close-on-select" attribute?

jtommy:
@williamabbott If ok, why not?

PR #1077, opened 2018-11-12
closeOnSelect to autocomplete, added to taginput docs

williamabbott:
#577

### DIFF
--- a/docs/pages/components/taginput/examples/ExAutocomplete.vue
+++ b/docs/pages/components/taginput/examples/ExAutocomplete.vue
@@ -4,6 +4,12 @@
             <b-switch v-model="allowNew">
                 Allow new tags
             </b-switch>
+            <b-switch v-model="openOnFocus">
+                Open dropdown on focus
+            </b-switch>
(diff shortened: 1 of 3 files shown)


jtommy:
@williamabbott I tested but I didn't found any difference when that prop is enabled or not to be honest

williamabbott:
@jtommy seeing some strange behaviour with autocomplete, by adding 'open-on-focus', sometimes the dropdown stays open for 2 or 3 selections. When it disappears, it's visibility is gone, but clicking in that area can select some options (this is from the dev branch with mods only to the taginput docs) gif:

autocomplete

jtommy:
Thanks. I think it's better to know when and why there is that strange behaviour before to add a new prop

williamabbott:
I have narrowed this bug down to the transition group that wraps around Autocomplete's dropdown-menu leaves the class `.fade-leave-to` ( opacity: 0 ) 

image

See the invisible selection on left. I think it could be related to this reported bug: https://github.com/vuejs/vue/issues/5800

Issue #1067, opened 2018-11-05
Switches do not work in IE11


danielclariondoor:
### Overview of the problem

**Buefy** version: 0.6.7
**Vuejs** version: 2.5.17
**OS/Browser**: Windows 8 / IE 11

### Description

Switches do not look like switches on IE11. The middle part is completely missing. It appears that the pseudo element "before" is not redered.

### Expected behavior

Should see switches with a slider in the middle.

### Actual behavior

There are just colored ovals.
<img width="2547" alt="screenshot at nov 05 15-25-20" src="https://user-images.githubusercontent.com/18540311/48033161-0fc30c00-e10f-11e8-8ec9-6552118554a4.png">

jtommy:
@danielclariondoor Are you sure ? I have no issue in IE11 (win 10).

danielclariondoor:
I downloaded Windows 8 with IE11 from MS and they are definitely not working, as per my screenshot.

morficus:
can confirm: IE11 on Window 8.1.... switches look busted. clicking on them does toggle them, so the issue is just visual.

IE11 on Windows 10... everything works as expected.

PR #1082, opened 2018-11-13
IE11 fix for switch

patrickabkarian:
fixes: https://github.com/buefy/buefy/issues/1067


jtommy:
@patrickabkarian Thanks

Issue #1080, opened 2018-11-13
Datepicker: Allow to pass property for custom Date object creation


jonagoldman:
### Overview of the problem
Currently, there is no way to specify how the Date() object is created inside the Datepicker component.

**Buefy** version: 0.7.0
**Vuejs** version: 2.5.13
**OS/Browser**: All

### Description
Looking at the source code of the Datepicker component, I see that `new Date()`  is called a few times inside `Datepicker.vue` and `DatepickerTableRow.vue` components.
I think providing a property to allow for custom Date() creation can be very useful in some cases, for example when working with timezones using an external date library (like Moment or Luxon).

For example in DatepickerTableRow.vue, `new Date()` is used to select today's date, but maybe the app is using a different timezone, and today's date inside the Datepicker will not match today's date in the application.

https://github.com/buefy/buefy/blob/25c2a2468b838d927a7a41b2a7e72d889b697dab/src/components/datepicker/DatepickerTableRow.vue#L157

### Proposal
Add a new optional property to the component similar to `date-formatter` and `date-parser` that allows to customize how the date is created:

```html
<b-datepicker :date-creator="dateCreator"></b-datepicker>
```

```js
// must return a Date() instance, and by default it just returns new Date()
function dateCreator() {
    let date = moment()
    /// ... do some work here and return a new Date instance
    return date.toDate()
}
```

PR #1083, opened 2018-11-13
Add 'todayInstantiator' property to allow custom today's Date creation

jonagoldman:
Fixes https://github.com/buefy/buefy/issues/1080 by adding a new `todayInstantiator` property.

This allows to pass a function like this:

```html
<b-datepicker :today-instantiator="todayInstantiator"></b-datepicker>
```

```js
todayInstantiator() {
    return new Date()
},
```

This fixes some issues when the browser default new Date() instantiation used inside the component does not match the way each application works with dates.

For example, in an app that uses Moment or Luxon to set a timezone, the today's date set by the app can be different from the today's date used inside the component.

With this property I have the freedom to do something like this:

```js
todayInstantiator() {
    return moment().tz('Australia/Sydney').toDate()
},
```

The `todayInstantiator` property is a function that must always return a regular JS Date object representing today's date.

**Please note** this function only applies when the component wants to get the today's date using `new Date()` and does not change anything else, ensuring that nothing breaks.

### DIFF
--- a/docs/pages/components/datepicker/api/datepicker.js
+++ b/docs/pages/components/datepicker/api/datepicker.js
@@ -23,6 +23,13 @@ export default [
                 default: '<code>(date) => new Date(Date.parse(date))</code>'
             },
             {
+                name: '<code>date-instantiator</code>',
+                description: 'Function used internally to create a new Date instance',
+                type: 'Function',
+                values: '—',
+                default: '<code>() => new Date()</code>'
+            },
+            {
                 name: '<code>min-date</code>',
                 description: 'Earliest date available for selection',
                 type: 'Date',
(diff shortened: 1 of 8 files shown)


jtommy:
@jonagoldman PR is ok, but i'm not sure about the prop name to be honest

jonagoldman:
@jtommy I'm not sure either :) First I called it `date-instatiator` but it's not accureate as it only instantiates today's date. Maybe `today-creator` or `date-creator` ?

jtommy:
`today-date` or `date-creator`, as you want

jonagoldman:
@jtommy done

jtommy:
@jonagoldman Thanks

zedrdave:
If I understand correctly, this would allow displaying the Datepicker in an arbitrary timezone instead of the user local by default?

Would it be possible to add the same parameter to `Timepicker`, so that it can be available to all components (I'm assuming it would then be possible to use `Datetimepicker` by providing both parameters to the subcomponents, right?)…

PR #1085, opened 2018-11-14
add docs for the table `striped` prop

PhilWaldmann:
`striped` is documented via an example, but is missing in the API table


jtommy:
@PhilWaldmann Thanks

PR #1090, opened 2018-11-16
Add table context menu event

gps-dev:


jtommy:
@gps-dev I had to remove prevent. I'll add an example in docs.

PR #1102, opened 2018-11-21
table: add multicolumn sort support

AlexKornitzer:
Hey, totally understand if you don't want to merge this but I have added multicolumn support to the table component as it was a requirement for what I needed. It works the same way as datatables where you hold shift to select multiple columns. The only breaking change I believe is that it now accepts a list of sortfields so this: `[field1, field2, field3]` or this `[[field1, desc], [field2, asc]]`. Hopefully it is bug free too, been testing it for the last week in backend mode.


jtommy:
https://github.com/buefy/buefy/pull/1694

PR #1127, opened 2018-12-17
Datatable: support for custom row

carlobeltrame:
This includes just the first commit from #449, manually re-applied to the current repository state.

### DIFF
--- a/docs/pages/components/table/Table.vue
+++ b/docs/pages/components/table/Table.vue
@@ -49,6 +49,13 @@
             </p>
         </Example>
 
+        <Example :component="ExCustomRow" :code="ExCustomRowCode" title="Custom rows">
+            <p>
+                By adding a prop named <code>customRow</code> with value true, you can skip the <code>tr</code> element and insert a custom component instead. <br>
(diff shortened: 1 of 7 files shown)


EvanBarbour3:
@jtommy I've been looking for something like this.

What this pull request would do is allow people to provide their own custom row(s) in the detail slot which would be really useful. I've attached an image of what this could be used for if it was included. This may now be out of date

EvanBarbour3:
@jtommy I've been looking for something like this.

What this pull request would do is allow people to provide their own custom row(s) in the detail slot which would be really useful. I've attached an example image of what this could be used for if it was included. 

This pull request may now be outdated, if you're interested in this feature I'd be happy to take a look.

I think passing the displayed columns into the slot would help, as people using the custom detail row would be able to manage the columns being displayed and hidden (currently handled in the available detail row via `:colspan="columnCount"`)

Screenshot 2019-04-17 at 10 11 17

jtommy:
@EvanBarbour3 I don't understand very well the example in the PR, can you provide an example to make me change my mind ?

jtommy:
In review #1336

PR #1132, opened 2018-12-18
Fixed tabindex on Radio and RadioButton components

Sacmanxman2:
Tabbing through radio buttons now only requires one tab press instead of two.


jtommy:
@Sacmanxman2 I think it's better add a new prop instead of set always "-1"

jtommy:
@Sacmanxman2 Thanks

PR #1133, opened 2018-12-19
Update input.js (add '@')

kelsonire:
Because most of Vue.js project you use '@click' instead of 'click'


jtommy:
Same thing in Buefy but in the docs we show only the event name

PR #1141, opened 2018-12-24
Docs | Add example of loading inside router hook

### LINKED ISSUE #312: Using bLoading inside VueRouter beforeEach - IT IS NOT A BUG
I know this is not a bug, but I do not know where to ask about something in Bulma or Buefy.

My problem is:

1. I want to use Buefy.Loading inside beforeEach, in my router.js to show centralize the loading of each page, how can I do this?

Thanks and Regards,

Rodolfo Azevedo

MatiasOlivera:
It is a common use case to use the `bloading `component inside a component navigation guard.

Fixes #312

### DIFF
--- a/docs/pages/components/loading/Loading.vue
+++ b/docs/pages/components/loading/Loading.vue
@@ -12,6 +12,10 @@
             <p>When you want to close the Loading, call the <code>close()</code> method from the component.</p>
         </Example>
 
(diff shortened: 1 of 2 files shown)


jtommy:
@MatiasOlivera You might use `import { LoadingProgrammatic } from 'buefy/dist/components/loading'` and then `LoadingProgrammatic.open()`, did you try this way ?

MatiasOlivera:
@jtommy I'm going to try it, it would be nice to have it in the documentation.

PR #1143, opened 2018-12-27
Add "parent" prop for Snackbar

cuong-nguyen:
This PR to make **snackbar** component work like modal component

# Description:
We are using **buefy** with **nuxt** which uses **vue-meta** which depends on `$parent` prop to setup meta tags on HTML head section.

# Problem:
Currently as soon as **snackbar** is created, nuxt checks for `$parent` prop then populate another prop (`$metaInfo`) then update the HTML head section accordingly. Because there's no parent prop passed in, `$metaInfo` prop is empty, nuxt remove some existing meta tags which causes broken page.


jtommy:
@cuong-nguyen Thanks

PR #1147, opened 2018-12-29
Add text-color property to BInput component

pflirae:
Adding a text-color property to the BInput component enable the use of any Bulma text color class or custom class to change the color of text within an input field.


jtommy:
@pflirae Thanks.
What do you think to change the prop name? customClass ?

pflirae:
Will change to custom-class and generate a new pull.

PR #1149, opened 2018-12-31
Add custom-class property to BInput component

pflirae:
Enable the custom-class on input, so CSS styles con be applied on input field.


jtommy:
Thanks

PR #1153, opened 2019-01-02
Fix BTable behavior when user wants to display all data in one page and paginated is true

pflirae:
The use case is as follow:

a combobox with different amounts of rows to show, like the example shown on documentation, including a last option that shows all the data (Todos), as can be seen on the following figure:

image

(Todos means All in spanish)

Current behavior:

When you select the Todos option, the list of pages is infinite, as shown in the following figure:

image

New behavior:

When the value 0 is assigned to perPage, in this exmple to the Todos option, the table will show all the data on a single page, allowing the option Todos to work correctly, that is, that the list of pages be displayed accordingly, which can be seen in the following figure :

image

The BTable documentation and the paging example have been modified to reflect this modification.

### DIFF
--- a/docs/pages/components/input/api/input.js
+++ b/docs/pages/components/input/api/input.js
@@ -72,10 +72,10 @@ export default [
                 default: '<code>true</code>'
             },
             {
-                name: '<code>text-color</code>',
-                description: 'Color to be used in input text, optional',
+                name: '<code>custom-class</code>',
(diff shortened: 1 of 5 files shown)


pflirae:
@jtommy I opted for your second proposal, thanks for the advise.
So, this pull request is closed without the need for a merge.
Regards

PR #1155, opened 2019-01-03
added control on chevron icon & column for detailed table

omer2500:
i didn't know what to do with the selecRow() function so i made this solution 

please tell me what you think and if i need to change something

Thanks!

### DIFF
--- a/docs/pages/components/table/api/table.js
+++ b/docs/pages/components/table/api/table.js
@@ -171,6 +171,13 @@ export default [
                 default: '<code>false</code>'
             },
             {
+                name: '<code>chevron</code>',
+                description: 'Allow shevron icon and column to be visible. if not will make the row click expand the row',
+                type: 'Boolean',
+                values: '—',
+                default: '<code>true</code>'
+            },
+            {
                 name: '<code>opened-detailed</code>',
                 description: 'Allow pre-defined opened details. Ideal to open details via vue-router. (A unique key is required; check detail-key prop)',
                 type: 'Array',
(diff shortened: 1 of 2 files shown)


omer2500:
i fixed what @wanxe  suggested.
about the third thing he mentioned
im not sure because of the selectRow() function.

wanxe:
Thanks so much for the contribution, now we need to wait what says @jtommy :D

docs/pages/components/table/api/table.js
```diff
@@ -170,6 +170,13 @@ export default [
                 values: '—',
                 default: '<code>false</code>'
             },
+            {
+                name: '<code>chevron</code>',
+                description: 'Allow chevron icon and column to be visible. if not will make the row click expand the row',
```
jtommy: Please change name to `showDetailIcon`

src/components/table/Table.vue
```diff
@@ -66,13 +66,13 @@
                             :key="index"
                             :class="[rowClass(row, index), {
                                 'is-selected': row === selected,
-                                'is-checked': isRowChecked(row)
+                                'is-checked': isRowChecked(row),
                             }]"
-                            @click="selectRow(row)"
+                            @click="detailed && !chevron ? toggleOnclick(row) : selectRow(row)"
                             @dblclick="$emit('dblclick', row)">
```
jtommy: Remove it. I want to let free to toggle detail row using directly the method `toggleDetails(row)`, so a developer could be open the detail as he wants.

omer2500: i'm not sure i 100% understand you.
soon as i restore it,
the `toggleDetails(row)` wont be called if you hide the chevron column & icon, so the table will be just a normal table.

docs/pages/components/table/api/table.js
omer2500: will fix! thanks

src/components/table/Table.vue
jtommy: Yes a normal table and the developer choose how toggle detail section. For example click all row or button or custom icon

omer2500: got it.
i will make the changes and commit soon

omer2500:
i fixed the issues based on your comments!

jtommy:
@omer2500 Thanks!!!

Issue #73, opened 2017-05-25
User individual components with font-awesome


darkylmnx:
### Overview of the problem
Vue ^2.2.6
Buefy ^0.3.2

### Description

How can i use font-awesome while using individual components ? As in you docs you say to do the following :

```
Vue.use(Buefy, {
    defaultIconPack: 'fa',
    defaultContentElement: '#content',
    ...
})
```
but by doing this it enables Buefy for all components not individual.

rafaelpimpa:
To use it with ``b-icon`` you must pass the ``pack`` attribute:

```html
<b-icon icon="search" pack="fa"></b-icon>
```

RuslanHasko:
Same question.

**Overview of the problem**

Vue ^2.2.6
Buefy ^0.3.2

**Description**

In my project I use only tables (with child components for columns and pagination).

As in your docs, I use:

`Vue.use(Buefy, {
    defaultIconPack: 'fa'
})`

But classes for icons have not changed (in pagination component and in table-column component):

`<i class="mdi">chevron_left</i>`

So, how can I fix this?
I do not want to change the default template.

rafaelpimpa:
@RuslanHasko did you import fontawesome to your project? It's nedded.

RuslanHasko:
Yes of course

rafaelpimpa:
I just tested and it's working fine for me 🤔 

``Vue.use(Buefy, { defaultIconPack: 'fa' })`` changes all icons of tables, pagination, message, notifications etc.

If your project is open I could take a look.

darkylmnx:
@rafaelpimpa 

> To use it with b-icon you must pass the pack attribute:

well it's quite cumbersome to do that as i know i'll only use font awesome to add it explicitly on EACH icon...

rafaelpimpa:
That's why you have the option of ``defaultIconPack``.

darkylmnx:
But if I use this options all buefy components are loaded, while I don't
want that.

In terms of component based architecture, I should be able to set a global
config, while still importing components on demand.

I think you should consider adding and option to the Vue install options,
which would be something like : loadAllComponents: false or true.

This gives the possibility to set global options while still keep the
possibility to load all components only of we want to.

Though here the right terme would be "define" components instead of "load"
m, but younger the point

Le 26 mai 2017 16:56, "Rafael Beraldo" <notifications@github.com> a écrit :

That's why you have the option of defaultIconPack.

—
[...]

PR #1157, opened 2019-01-04
Config for individual use of components

steveperrito:
When using components individually, one must first import the config object to set global options. Thought that info would be useful here.

### DIFF
--- a/docs/pages/installation/ConstructorOptions.vue
+++ b/docs/pages/installation/ConstructorOptions.vue
@@ -3,6 +3,12 @@
         <div class="example is-paddingless">
             <CodeView :code="usage | pre" lang="javascript" expanded/>
         </div>
+        
+        <b-message type="is-warning">
+            If using components individually, set global options like this:<br>
+            <code>import config from 'buefy/src/utils/config';</code><br>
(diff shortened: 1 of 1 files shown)


jtommy:
@steveperrito Are you sure it's right?
When individual component I think each component has own config

steveperrito:
Yeah, it's working in one of my current projects and is referenced from here: https://github.com/buefy/buefy/issues/73#issuecomment-304333025

steveperrito:
Of course, working doesn't necessarily mean right. So if there's a better way of doing this, feel free to disregard the pull req.

jtommy:
@steveperrito I should try it. Are you importing components from src?

steveperrito:
@jtommy Yes, here's what my main.js looks like:

```javascript
import config from 'buefy/src/utils/config';
import Field from 'buefy/src/components/field';
import Datepicker from 'buefy/src/components/datepicker';

config.defaultIconPack = 'fas';

Vue.use(Field);
Vue.use(Datepicker);
```

jtommy:
Ah ok now I understand. It's ok but we added a way to import individual component since 0.6.9 version without using src.
However you cannot use global config

PR #1159, opened 2019-01-04
New field attribute for visually-hidden labels (with example)

jorgepinon:


src/components/field/Field.vue
```diff
@@ -7,15 +7,16 @@
             <label
                 v-if="label"
                 :for="labelFor"
-                class="label" >
+                class="label"
+                :class="[labelSrOnly ? 'is-sr-only' : '']">
```
jtommy: Please use a computed property

src/components/field/Field.vue
```diff
@@ -7,15 +7,16 @@
...
                 {{ label }}
             </label>
         </div>
         <template v-else>
             <label
                 v-if="label"
                 :for="labelFor"
-                :class="customClass"
+                :class="[labelSrOnly ? 'is-sr-only' : '', customClass]"
```
jtommy: As above

jtommy:
It's my fault, you can use `custom-class` , for example `custom-class="is-sr-only"`

PR #1173, opened 2019-01-16
`preventDefault` dropdown trigger click event

kzkn:
- `preventDefault` dropdrown trigger click event
 - It avoids submitting a form when click dropdown trigger button in `<form>`


jtommy:
@kzkn Are you using a button as trigger ? You have to add `type="button"`.

kzkn:
@jtommy It works. Thanks!

Issue #568, opened 2018-01-30
option to keep table header visible when data length is 0


frantic1048:
### Overview of the problem

**Buefy** version: 0.6.2
**Vuejs** version: 2.5.2
**OS/Browser**: various

### Description

This issue is a feature request:

- an option for b-table, that keeps table being rendered with the table header
- or, make table's empty slot replace only the table body area
- or somewhat equivalent

### Context

I implemented some filter controls within custom table header. The table data is being filtered according to filter controls status in the table header.

When I set [a combination of] filter options that produces no data as a result. The table was entirely replaced by its `empty` slot. I could only refresh the page or go backward(if the filter state is bound on URL query string), to make the filter controls in table header back. Then I can try other filter options.

If the table header could be rendered when data is empty, thus I could still interact with filter controls on table header, without refreshing or going backward.
 
---

update

2018-01-31: fix grammar in the text

grebebird:
+1 for either fixing this issue, or updating Buefy's documentation. Oh please please do one or the other to save future developers' time. :)

The saddest thing about this issue is that Buefy's own example of this ([Sandbox with custom template](https://buefy.github.io/#/documentation/table)) is flawed, which led our team to a bunch of time spent working in circles to try to recreate what we had been seeing in the documentation. 

I CodePen'ed Buefy's documentation with the "isEmpty" flag set to TRUE **_first_** to demonstrate this issue. The headers don't show up, but if you toggle the "isEmpty" flag to FALSE and **_then_** TRUE, the headers are back, in the way we were expecting the first time through.

https://codepen.io/anon/pen/NYExzV?&editors=101

ivalkenburg:
This happens because the newColums property isn't being populated when there is no data. No data means there are no table column component created that register itself to the parent table component which populates the newColumns property.

willhoyle:
I'll look into the src tonight and see if it can be fixed but for now you can kind of do the same hack as the Documentation:

```js
<template>
    <b-table :data="fake ? fakeData : data">
    </b-table>
</template>

<script>
export default {
data() {
 return {
    fake: true,
    data: [],
    fakeData: [{id: 1, name: " "}] // fill in some fake empty values for your columns
},
  mounted() {
    this.$nextTick(() => {
      this.fake = false
    })
  }
}
</script>
```

It's far from ideal and extremely hacky but in a pinch this can work.

willhoyle:
Okay had time to take a proper look at the source.

[...]

PR #1180, opened 2019-01-18
Fix duplicate column headers

willhoyle:
Inside the some( ) function, column.newKey is undefined when the :data prop is empty. This causes the table to show duplicate headers.


willhoyle:
In the end, this is gonna need more work. I'll try to do something about it this weekend

jtommy:
https://github.com/buefy/buefy/issues/568#issuecomment-455715201

PR #1206, opened 2019-02-03
Accessibility for Collapse component. 

mateuswetah:
Puts aria-expanded. Creates a11y-id props to bused on div content. Adds aria-controls and role to triggers on documentation.

This is the first commit of a serious of upcoming related to accessibility. Naming conventions,such as the a11y for the special id prop are still being discussed.


jtommy:
Thanks!

Issue #1135, opened 2018-12-19
Modal component width issue


serzhiio:
When set up 'width' property of **modal** component, it sets **style="max-width: XXXpx"**.
Why not simple **width**?

P.S.: btw thx for outstanding work, well done!

yxngl:
The pull request failed to close multiple issues. For your question, simply using width is not responsive and will fail on mobile or in a narrower browser window.

Youdaman:
You can fix the width param not working by setting the width of .modal-card to auto, see https://github.com/buefy/buefy/issues/519#issuecomment-356127796

sookoll:
It is not working. I set modal width=1200 but it still produce modal with 640px and if I set .modal-card width=auto, it even more narrows modal width to 460px.

yxngl:
@sookoll You should use either `:width="1200"` or `width="1200px"`
It will be taken as a number only with a colon (as a JS expression).

PR #1219, opened 2019-02-08
Fix Modal width limit set by Bulma

### LINKED ISSUE #905: .modal-content width incorrect
**Buefy** version: [0.6.6]

The docs for modal components say the default width is 960 whereas the modal.sass has `.modal-content` width set to 640px and so modals display at 640 even when width is set higher than 640, and the width param only sets the max-width.

Possible fix is to override with:
```
.modal-content {
  width: auto !important;
}
```
To ensure width param of modal component is used.

yxngl:
This fixes the problem that setting the `width` prop of Modal over 640px does not take effect.

Bulma has a fixed variable defining the width of `modal-content` class, which defaults to 640px. Currently, setting `width` of Modal larger than 640px is not working (see "Launch image modal" in documentation).

By setting style of width 100% on `modal-content` also keeps modal component in card with `has-modal-card` prop working as before.

closes #905 #1135

### DIFF
--- a/docs/pages/components/modal/api/modal.js
+++ b/docs/pages/components/modal/api/modal.js
@@ -46,7 +46,7 @@ export default [
             },
             {
                 name: '<code>width</code>',
-                description: 'Width of the Modal, maximum of <code>960</code>',
+                description: 'Width of the Modal',
                 type: 'Number, String',
                 values: '—',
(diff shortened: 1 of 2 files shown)


jtommy:
@yxngl Thanks!

PR #1225, opened 2019-02-11
Fixing a field component bug

kpdemetriou:
Fixing a bug where custom label classes are not applied to horizontal fields


jtommy:
Thanks!

Issue #1228, opened 2019-02-13
Responsiveness of multiple fields: field has an unwanted left margin


edum18:
### Overview of the problem

**Buefy** version: 0.7.2
**Vuejs** version: 2.6.6
**OS/Browser**: Windows10/Chrome

### Steps to reproduce

codesandbox:
https://codesandbox.io/s/3ynz8zvo3q

### Expected behavior

By shrinking page, the second input shouldn't have the left margin.

### Actual behavior

By shrinking page, the second input has a left margin.

Should I just use columns then? Doesn't seem right since there is `grouped group-multiline` options.

yxngl:
Targeting wrapped elements seems hard. May be fixed by changing left margin of 0.75rem to right margin for nested .field.
https://github.com/buefy/buefy/blob/5737a8e1d567899539637ba69af9e24bed32a50e/src/scss/components/_form.scss#L5-L17

To:
```sass
.field {
    &.is-grouped {
        .field {
            flex-shrink: 0;
            margin-right: 0.75rem;
            &.is-expanded {
                flex-grow: 1;
                flex-shrink: 1;
            }
        }
    }
}
```

It is not the best, since there will be margins at the line and wrap ends.

yxngl:
Created a PR. But it will not align on the right when the last field is expanded. For your case, it is probably better to use columns.

PR #1233, opened 2019-02-19
Change left margin to right margin when spacing nested fields

yxngl:
Change to use right margins to space nested fields, which is then consistent with grouped controls in Bulma. It will solve the problem of a left margin in wrapped lines when group-multiline is applied in #1228 .


jtommy:
@yxngl Thanks

PR #1238, opened 2019-02-22
Show loading spinner on confirm button of dialog when callback returns a Promise

marc1404:
This PR would add Bulma's `is-loading` class to the confirm button of a Dialog when its callback returns a Promise. The spinner will be visible until the confirm Promise resolves.
The usage of the Dialog doesn't change as any result of the confirmation callback is "promisified" using `Promise.resolve()`. This means that any normal or void function will resolve immediately whereas an async function returns a Promise which will be awaited.

image

### DIFF
--- a/docs/pages/components/dialog/api/dialog.js
+++ b/docs/pages/components/dialog/api/dialog.js
@@ -91,7 +91,7 @@ export default [
             {
                 name: '<code>onConfirm</code>',
                 description: 'Callback function when the confirm button is clicked',
-                type: 'Function (value: String)',
+                type: 'Function (value: String): void, Promise',
                 values: '—',
                 default: '—'
             },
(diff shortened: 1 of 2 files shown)


jtommy:
@marc1404 The idea to accept a Promise I good but I don't like to add a custom logic in a generic component

marc1404:
@jtommy Thank you for your feedback. I understand that custom logic in a generic component isn't very clean.
I could alter the PR in one of two ways:
- Extracting the custom logic into a mixin
- Keeping a reactive `isLoading` property and making it possible to set the loading state from outside the component

Please let me know what you think. I'm mainly interested in making it possible to show the loading spinner on the confirm button.

jtommy:
I think both ways aren't good, you should simply use a modal component at the moment

PR #1242, opened 2019-02-25
Added preventDefault on mousedown when selecting an option

Livijn:
Adding this switches the order of the events when clicking/selecting an option. This is useful for when you auto-select an item on blur, etc.

**Before**
1. mousedown
2. blur (from input)
3. mouseup
4. click

**After**
1. mousedown
2. mouseup
3. click
4. blur


jtommy:
@Livijn Is it a random or documentend event sequence ? Anyway you should check your PR

Livijn:
Different browser handles to event order differently. Here is a discussion about the same thing: https://forum.vuejs.org/t/blur-before-click-only-on-safari/21598/6

PR #1244, opened 2019-02-26
Typo in usehtml5validation test

yereby:
Hi,
I believe there is a typo here
https://github.com/buefy/buefy/blob/7c658b1336b1a62f8e7140eca5449cb979267f5c/src/utils/FormElementMixin.js#L106

It should be `if (this.useHtml5Validation) return` otherwise the prop is not used and buefy continue to validate the component

Ref : this issue #1021

### DIFF
--- a/src/utils/FormElementMixin.js
+++ b/src/utils/FormElementMixin.js
@@ -103,7 +103,7 @@ export default {
          * and error message to parent if it's a Field.
          */
         checkHtml5Validity() {
-            if (!this.useHtml5Validation) return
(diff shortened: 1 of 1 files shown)


jtommy:
Because the default value is true otherwise it could be a breaking change.

yereby:
Ok but i don't see how to use it on a component then

jtommy:
`<b-input use-html5-validation="false"`

yereby:
Ho my bad i forgot the bind.
Thanks a lot !

Issue #1237, opened 2019-02-22
Toggle the date-picker on input focus/blur


imfaber:
### Overview of the problem

**Buefy** version: 0.7.2
**Vuejs** version: 2.5.17
**OS/Browser**: Windows/Chrome

### Description

Navigating the page with the tab button the datepicker doesn't show up when the input field has the focus. In other words it cannot be used through the keyboard.

### Steps to reproduce

Tab into a date field

### Expected behavior

The datepicker shows up either right away or after pressing **Enter** (a prop to control this behaviour would be nice).

### Actual behavior

The datepicker doesn't show up and it seems there is no way of triggering it without clicking it with the mouse

PR #1248, opened 2019-02-26
Improve datepicker accessibility

imfaber:
Make datepicker accessible through keyboard only navigation (#1237 ).

**New behaviour**
- Tab into a datepicker field.
- Press `Enter` to open datepicker
- Use `Tab` + `Enter` to select a date or press `Esc` to close it

A new prop called `open-on-focus` has been added.
If true, the datepicker is open when the input field gets the focus (no need to press `Enter`).


jtommy:
@imfaber Thanks! But first i would like check a common way using dropdown (if there is) because I'll be need to add it to timepicker too

imfaber:
hey @jtommy 
No problem, if you think the solution for the datepicker is suitable for the timepicker I can work on it.
Let me know if I may help with anything 😃

jtommy:
@imfaber I would apply a common solution to b-dropodown since datepicker, timepicker, autocomplete use it

jtommy:
@imfaber Thanks

PR #1250, opened 2019-02-28
Adding draggable to table rows

suneab:
Allows user to set :draggable="true" (default false) to enable dragging of rows. 
No swapping of rows has been implemented, but instead emits dragstart, dragover and drop events, with a payload of the event, row and index
draggable0

### DIFF
--- a/docs/pages/components/table/Table.vue
+++ b/docs/pages/components/table/Table.vue
@@ -70,6 +70,13 @@
             <p><small>API from <a href="https://www.themoviedb.org" target="_blank">TMDb</a>.</small></p>
         </Example>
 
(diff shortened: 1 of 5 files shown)


jtommy:
@suneab Great! 
Could you add a simple effect css on drag and drop elements ? For example like your GIF.

suneab:
@jtommy I'm not sure what I'm supposed to change. I just added dropEffect prop. Let me know if i misunderstood. 
draggableevent1

jtommy:
@suneab I meant only in the example and not a new prop.

suneab:
Right, yanked it all out again and added it to the example.

jtommy:
@suneab Thanks!

PR #1266, opened 2019-03-06
add option to use v-if in b-tab-item

tarun1793:
Conditional rendering is a pretty common use case when employing tabs. The current way of achieving this is to keep a track of `$index` and `activeTab`.  Will add documentation and coverage code if you could let me know that this is not an undesirable change.


jtommy:
@tarun1793 Why do you need it ?

jtommy:
Close https://github.com/buefy/buefy/commit/863ab274d76e549db647cda3f3a89ad6edb80f23

PR #1270, opened 2019-03-07
Fix #1269 Switch captures focus twice

### LINKED ISSUE #1269: b-switch is focused twice when navigating with the tab key
### Overview of the problem

**Buefy** version: 0.7.2
**Vuejs** version: 2.6.6
**OS/Browser**:  Vivaldi 2.3 windows 10

### Description

`b-switch` is focused twice, one visibly and one invisibly.

### Steps to reproduce

1. Go to https://buefy.org/documentation/switch/
2. Press tab key until the first switch example is focused
3. Press tab once more
4. Press space to enable/disable the switch

### Expected behavior

The second switch should be focused and changed.

### Actual behavior

The first switch is still selected, to select the second switch you have to press tab twice.

yxngl:
Fix #1269. The hidden input checkbox captures the second focus when tabbing, which is disabled here. The focus on label is needed for visual effects, like glowing.

### DIFF
--- a/src/components/switch/Switch.vue
+++ b/src/components/switch/Switch.vue
@@ -13,6 +13,7 @@
         <input
             v-model="computedValue"
             type="checkbox"
+            tabindex="-1"
             @click.stop
             :disabled="disabled"
             :name="name"


jtommy:
@yxngl Thanks!

PR #1272, opened 2019-03-09
Fix space keydown prevent not working on Firefox at Checkbox

krororo:
Firefox can not prevent space key down event.
And then, after pressing the space key, focus on the checkbox element.

Focus on the checkbox instead of the label, so that native events work.

### DIFF
--- a/src/components/checkbox/Checkbox.vue
+++ b/src/components/checkbox/Checkbox.vue
@@ -4,11 +4,9 @@
         :class="[size, { 'is-disabled': disabled }]"
         ref="label"
         :disabled="disabled"
-        :tabindex="disabled ? false : 0"
-        @keydown.prevent.enter.space="$refs.label.click()">
+        @keydown.prevent.enter="$refs.label.click()">
         <input
             v-model="computedValue"
-            tabindex="-1"
             :indeterminate.prop="indeterminate"
             type="checkbox"
             :disabled="disabled"
--- a/src/components/checkbox/CheckboxButton.vue
+++ b/src/components/checkbox/CheckboxButton.vue
@@ -3,19 +3,22 @@
         <label
             class="b-checkbox checkbox button"
             ref="label"
-            :class="[checked ? type : null, size, { 'is-disabled': disabled }]"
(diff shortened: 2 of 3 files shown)


yxngl:
The problem only appears in Firefox on Windows, MacOS seems fine. The checkbox flashes and state changes back. The same problem exists for Switch.

Not sure it is just a browser quirk that can be easily fixed, or the design of many similar components should be changed to use native input elements. I got a minimal reproducible demo and will ask in Vue issues.

jtommy:
@yxngl Thanks, let me know!

yxngl:
See https://github.com/vuejs/vue/issues/9678
I also think introducing a `checked` data in replacement of clicking label seems to be a simpler approach to fix the problem.

krororo:
@yxngl Thank you for reply and investigation.

This problem also appeared with Firefox on Linux.

I also want to fix that the focus moves label to checkbox. However, I think that it will not be fixed by only introducing the toggle method. What do you think?

yxngl:
What is you motivation for moving focus to checkbox? I just feel it may not be necessary and other components better need to be changed in the same way for consistency. In the end, it is decided by @jtommy and he may already have ideas of rewriting some of the components.

krororo:
Sorry, my explanation was insufficient. Tabindex is set to not focus on the checkbox but if it focus on the checkbox with the space key, it will look like #1269.

yxngl:
@krororo For that purpose, it has already been fixed in https://github.com/buefy/buefy/commit/26d99a45b2ec6dde0e019247f151a4cbb23da494. Similar things have been done for switch, checkbox, radio recently, although they are not live on the documentation site yet.

yxngl:
@krororo  I feel sorry you deleted the branch. Mostly because I was asked for opinions but I had been very busy last week. Could you submit a new PR from your fix-firefox-space-key branch? I think we agree it should be merged. Thanks.

krororo:
@yxngl NP! Thank you for your reply. I opened a new PR #1341. And, I also fixed Radio and Switch.

PR #1273, opened 2019-03-09
Add icon-pack prop to TableMobileSort component

rclement:
This PR adds the missing `icon-pack` prop to `TableMobileSort` component, and make the `Table` component provide the proper icon-pack to it, providing the same sorting arrows icon display in both desktop and mobile views.

### DIFF
--- a/src/components/table/TableMobileSort.vue
+++ b/src/components/table/TableMobileSort.vue
@@ -28,6 +28,7 @@
                         v-show="currentSortColumn === mobileSort"
                         :class="{ 'is-desc': !isAsc }"
                         icon="arrow-up"
+                        :pack="iconPack"
                         size="is-small"
                         both
                     />
@@ -51,7 +52,8 @@
             currentSortColumn: Object,
             isAsc: Boolean,
             columns: Array,
-            placeholder: String
+            placeholder: String,
+            iconPack: String
         },
         data() {
             return {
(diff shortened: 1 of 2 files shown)


src/components/table/TableMobileSort.vue
```diff
@@ -28,6 +28,7 @@
                         v-show="currentSortColumn === mobileSort"
                         :class="{ 'is-desc': !isAsc }"
                         icon="arrow-up"
+                        :pack="iconPack"
```
jtommy: There is `both` prop, you doesn't need to set icon pack.

rclement: Ok fair enough. But using FontAwesome 5 SVG does not seem to work with this specific component (the `Table` component in Desktop view works perfectly with the `arrow-up` and `arrow-down` icons when specifying `:icon-pack="fas"`). Any ideas what's the issue then?

jtommy: Did you set fas in the constructor options?

rclement: Ok now I feel silly. Using `nuxt-buefy`, I had only `defaultIconComponent: 'font-awesome-icon'` in the constructor options, working almost everywhere. Only adding `defaultIconPack: 'fas'` fixed the issue indeed, thanks!

Sorry for the noise, I'll close this PR.

PR #1281, opened 2019-03-14
Check url hash to see if it matches any tab labels

chrisrhymes:
A simple way of jumping to a specific tab when the page is loaded using a url hash value. 

For example, clicking a link to `link/to/page#info` would automatically select the tab with the label `Info` if it exists.

### DIFF
--- a/src/components/tabs/Tabs.vue
+++ b/src/components/tabs/Tabs.vue
@@ -118,6 +118,18 @@
             if (this.activeTab < this.tabItems.length) {
                 this.tabItems[this.activeTab].isActive = true
             }
+
+            /**
+             * Check the url hash and see if any tabs have the same label
+             * Get the index of the tab and change to that tab
+             */
+            let urlHash = window.location.hash.substr(1)
+            let urlTabIndex = this.tabItems.findIndex((tabItem) => {
+                return tabItem.label.toLowerCase().replace(' ', '-') === urlHash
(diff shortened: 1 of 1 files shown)


jtommy:
@chrisrhymes Thanks but I don't understand well why I should add it in the core code.

chrisrhymes:
Maybe it’s just my use case where I am using b-tabs directly in a blade template, rather than in another vue component. This way I won’t have to calculate the hash in another JavaScript function and try and trigger a tab change from outside vue or dynamically set a prop for the active tab.

jtommy:
Your use case is not rare but I think it's better put it on component code however.

scriptburn:
@jtommy this is a legit use case , suppose you a dropdown activates a tab component which has 3 sub tabs and i choosed x item from dropdown and my child tabs got its new data and they are refreshed and now i toggled few tab and ended up on tab num 3 and now if i am changing again an item from drop down then again child tabs need to be reloaded with new data and they all get reset to first tab , now just think  i have 10 tabs and i am trying to compare some information on tab 7 so i have to switch back and forth to same tab but from diffrent set of data from drop down , every time have to select tab again  and this is not good , this is a ui related framework and and it should be in there , some time now when i started to use it i am seeing lots of these kind of small things which i could have avoided just by choosing boostrap but stupid me thought to try something

PR #1287, opened 2019-03-15
DatepickerTable: accept events with time and show them

### LINKED ISSUE #1282: b-datepicker: events fails to display events which have a time set
### Overview of the problem

**Buefy** version: 0.7.3
**Vuejs** version: 2.6.0
**OS/Browser**: Osx/ Chrome (OS / Browser indipendant

### Description

If a date which contains a time different from 0 hours 0 minutes 0 seconds  and 0 milliseconds is passed to te events prop the date is not displayed as an event.

workaround is:

```js
  const date = new Date(t.date)
  date.setHours(0)
  date.setMinutes(0)
  date.setSeconds(0)
  date.setMilliseconds(0)
```

I checked the code and I know what / how to fix this PR incoming in the next few days

DonNicoJs:
Fixes #1282


jtommy:
Good job, thanks!!

PR #1288, opened 2019-03-17
[Snackbar] Add variable to change font color

### LINKED ISSUE #1278: [Snackbar] Inconsistent use of variable
https://github.com/buefy/buefy/blob/9170ec901a0c7643e080e9c5d171e0a088c6468d/src/scss/components/_notices.scss#L43

It should be like `$snackbar-background-color: $dark !default;`.
Maybe: `$snackbar-color: $dark-invert !default;`

This can be useful when we create a dark theme and want to display snackbar with light background.

Heziode:
And use `$snackbar-background-color` for the background of action button

Close #1278

### DIFF
--- a/src/scss/components/_notices.scss
+++ b/src/scss/components/_notices.scss
@@ -56,7 +56,7 @@ $snackbar-color: $dark-invert !default;
                 background: $snackbar-background-color;
             }
             .button:hover {
-                background: darken($snackbar-background-color, 5);
+                background: darken($snackbar-background-color, 5%);
             }
             @each $name, $pair in $colors {
                 $color: nth($pair, 1);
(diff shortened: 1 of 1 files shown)


MattStellify:
👍

src/scss/components/_notices.scss
```diff
@@ -52,6 +53,10 @@ $snackbar-background-color: $dark !default;
             .button {
                 font-weight: $weight-semibold;
                 text-transform: uppercase;
+                background: $snackbar-background-color;
+            }
+            .button:hover {
+                background: darken($snackbar-background-color, 10);
```
jtommy: It's better 5% instead of 10

Heziode: It is done.

jtommy: Please add % too :)

Heziode: Done ;)

jtommy:
Thanks

PR #1290, opened 2019-03-18
Fix double click event for CheckBox(Button) and Radio(Button)

### LINKED ISSUE #940: Radio or Radio-Button click event is triggered twice

yxngl:
Fix #940 for these label wrapped input. Clicking label will also trigger event on input. related to #711

### DIFF
--- a/src/components/checkbox/Checkbox.vue
+++ b/src/components/checkbox/Checkbox.vue
@@ -11,6 +11,7 @@
             tabindex="-1"
             :indeterminate.prop="indeterminate"
             type="checkbox"
+            @click.stop
             :disabled="disabled"
             :required="required"
             :name="name"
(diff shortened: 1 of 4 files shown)


jtommy:
Thanks

yxngl:
This together with the PR on Switch give only one `click` event, but the event target could be form different elements which depends on the interaction and usually not the input element, such as `span.control-label` or `span.check`

PR #1328, opened 2019-04-16
Some table customization

shelru:
percent's also must be allow

### DIFF
--- a/docs/pages/components/table/api/table.js
+++ b/docs/pages/components/table/api/table.js
@@ -420,7 +420,7 @@ export default [
             {
                 name: '<code>width</code>',
                 description: 'Column fixed width in pixels',
-                type: 'String',
+                type: 'Number, String',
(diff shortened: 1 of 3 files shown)


jtommy:
@shelru It's a good idea but did you check to set a width without px o %?

shelru:
o = or?
no.
If you can, just fix it, im newbie in js, xD. Add validator or smthng like dat.

shelru:
So, and another question, where i can get' build files from 'ci/circleci' after tests? For testing.

shelru:
Is that normal way?

shelru:
xD, sorry for names.

shelru:
Gabarage.   https://github.com/shelru/buefy/blob/dev/src/components/table/Table.vue#L51 also there must be v-if check's for `column.sortable`
image
image

@jtommy

PR #1336, opened 2019-04-18
[B-Table] Allow custom detail rows

EvanBarbour3:
### Overview

This pull request allows the usage of a custom detail row. Providing the `customDetailRow` prop to the table will change the `detail` slot from the original one to a custom one, allowing any content to be provided in the detail.

### Reason for pull request

The detail row is a brilliant feature of the table. For normal usage, a custom detail row is unnecessary, but I think other datasets could benefit from an option where the content is not aligned in a single, full colspan `<tr>` tag, and instead in whatever the user wants to pass into the slot.

Previous pull request is https://github.com/buefy/buefy/pull/1127, but I couldn't commit to it and it looks to now be stale.

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -126,7 +126,7 @@
                         <!-- Do not add `key` here (breaks details) -->
                         <!-- eslint-disable-next-line -->
                         <tr
-                            v-if="detailed && !customDetailRow && isVisibleDetailRow(row)"
+                            v-if="isActiveDetailRow(row)"
                             class="detail">
                             <td :colspan="columnCount">
                                 <div class="detail-container">
@@ -138,7 +138,7 @@
                             </td>
                         </tr>
                         <slot
-                            v-else-if="detailed && customDetailRow && isVisibleDetailRow(row)"
+                            v-if="isActiveCustomDetailRow(row)"
                             name="detail"
                             :row="row"
                             :index="index"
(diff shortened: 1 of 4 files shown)


src/components/table/Table.vue
```diff
@@ -126,7 +126,7 @@
                         <!-- Do not add `key` here (breaks details) -->
                         <!-- eslint-disable-next-line -->
                         <tr
-                            v-if="detailed && isVisibleDetailRow(row)"
+                            v-if="detailed && !customDetailRow && isVisibleDetailRow(row)"
```
wanxe: I've to suggest use a method instead, just to avoid more than 2 conditions on an if statement. An name for this method could be something like `isVisibleWhenDetailed(row)`

src/components/table/Table.vue
```diff
@@ -137,6 +137,12 @@
                                 </div>
                             </td>
                         </tr>
+                        <slot
```
wanxe: To improves the readability please, consider use the function with an if instead else-if

EvanBarbour3:
@wanxe Thanks for taking a look at the pull request.

I've made changes in line with what you've said. I went for the method names `isActiveDetailRow(row)` and `isActiveCustomDetailRow(row)` but these can be changed if you have something more suitable. I didn't go for `isVisibleWhenDetailed(row)` as I wasn't sure about the second method name for the custom row.

wanxe:
> @wanxe Thanks for taking a look at the pull request.
> 
> I've made changes in line with what you've said. I went for the method names `isActiveDetailRow(row)` and `isActiveCustomDetailRow(row)` but these can be changed if you have something more suitable. I didn't go for `isVisibleWhenDetailed(row)` as I wasn't sure about the second method name for the custom row.

That's looks so good mate. Thanks :)

jtommy:
@EvanBarbour3 I've just updated changelog because I forgot to mention you in the v0.7.5 release

PR #1341, opened 2019-04-23
Fix space keydown prevent not working on Firefox at Checkbox and Switch

### LINKED ISSUE #1320: Checkbox and Switch can't toggle with space key at Firefox on Linux and Windows
### Overview of the problem

**Buefy** version: [0.7.4]
**Vuejs** version: [2.6.10]
**OS/Browser**: Linux and Windows / Firefox

### Description

Checkbox and Switch flashes when toggling with the space key.

Checkbox:
checkbox

Switch:
switch

related PR: #1272 

### Steps to reproduce

1. Focus on Checkbox or Switch.
2. Press space key.

### Expected behavior

Checkbox and Switch can toggle with the space key.

### Actual behavior

Checkbox and Switch can't toggle with the space key.

krororo:
Fix #1320

### DIFF
--- a/src/components/checkbox/Checkbox.vue
+++ b/src/components/checkbox/Checkbox.vue
@@ -4,11 +4,9 @@
         :class="[size, { 'is-disabled': disabled }]"
         ref="label"
         :disabled="disabled"
-        :tabindex="disabled ? false : 0"
-        @keydown.prevent.enter.space="$refs.label.click()">
+        @keydown.prevent.enter="$refs.label.click()">
         <input
             v-model="computedValue"
-            tabindex="-1"
             :indeterminate.prop="indeterminate"
             type="checkbox"
             @click.stop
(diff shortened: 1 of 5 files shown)


yxngl:
Checkbox and Switch look good to me.
For Radio, there is a small problem that by moving active element back to native input, the inconsistency between browsers on how to navigate between radio buttons in the same group appears.

Before Buefy tabs through all radio buttons. With the changes, natively Chrome and Safari just tab to the group and let arrow keys to go through radio buttons, while Edge and Firefox can tab to individual radio buttons. This would be a breaking change, although it seems tabbing to group and using arrow keys is actually the ARIA recommendation.

Maybe we can skip radio changes for now. It currently still works in Firefox, just not having the active glowing effect after selection. It would be better to support arrow key navigation in the future and consider refactor then.

krororo:
OK. I reverted Radio.

Issue #640, opened 2018-02-23
BUG b-input with barcode scanner


brunoosilva:
### Overview of the problem

**Buefy** version: [0.6.3]
**Vuejs** version: [2.5.13]
**OS/Browser**: Chrome 64.0.3282.167 

### Description

Problem with `v-model` of `b-input` using barcode scanner (put data fast and press key ENTER), has delay for get data of model. I also tested with `v-model.lazy`, but don't work! 

I tested with `<input v-model="" />` pure and it works normally.

### Steps to reproduce

1. Put data fast and press ENTER in input inside of form
2. Get data incomplete in method of submit

### Expected behavior

Expect receive: `PED1519398320465`

### Actual behavior

Return `PED151939832046`

### Screen

![bug_b_input](https://user-images.githubusercontent.com/3465257/36609995-772511de-18ad-11e8-8ab1-34ab8a2b80e8.gif)

jtommy:
@brunoosilva `v-model.lazy` doesn't work with components but it's a sugar sintax of `:value="var" @change="($event) => var = $event.target.value"`. 
`b-input` doesn't emit change so you should bind native event, for example `@change.native="($event) => ..."`

jtommy:
@brunoosilva Is it works ?

afwn90cj93201nixr2e1re:
@jtommy close this too, and maybe we should clear all old issues without feedback's?

jtommy:
There are so many issues in the same status, btw thanks

afwn90cj93201nixr2e1re:
Yeah, i know, this would be better for creating real roadmap issue's list. Without non-feedback.

PR #1343, opened 2019-04-24
new lazy modifier for b-input

FragsterAt:
this is needed for barcode scanners etc.
ugly fix for https://github.com/buefy/buefy/issues/640

### DIFF
--- a/src/components/input/Input.spec.js
+++ b/src/components/input/Input.spec.js
@@ -174,13 +174,12 @@ describe('BInput', () => {
 
         $input.trigger('focus')
         expect(wrapper.emitted()['focus']).toBeTruthy()
+
         $input.trigger('input', {value: 't'});
-
         await wrapper.vm.$nextTick()
-
         expect(wrapper.emitted().input).toBeTruthy()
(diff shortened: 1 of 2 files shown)


jtommy:
@FragsterAt lazy modifier is the sugar syntax of `:value="myvalue @change="onchange"` but about b-input is `:value="myvalue @change.native="($event) => onchange($event.target.value)"`.
Maybe I could emit a change event like input but `lazy` prop is unnecessary

FragsterAt:
@jtommy unfortunately parent component updates value on('input'), even it has v-model.lazy modifier :(
my workaround is to add attribute to avoid emiting 'input' event from child component.

FragsterAt:
@jtommy 
> b-input is :value="myvalue @change.native="($event) => onchange($event.target.value)".

`<b-input :value="value " @change.native="($event) => value = $event.target.value"`
it works, but it is much less clear, than 
`<b-input v-model="value" lazy>`

jtommy:
Yes but i want to follow vue standard

PR #1365, opened 2019-04-29
Add a submit event for Autocomplete and Taginput

kvnkuang:
Hi,

I added a submit event for `Autocomplete` and `Taginput` to signal when a user presses enter while not adding a new tag. It is useful and intuitive for users to submit requests by pressing enter after adding all tags they want.

Thanks for considering the pull request.

### DIFF
--- a/src/components/autocomplete/Autocomplete.vue
+++ b/src/components/autocomplete/Autocomplete.vue
@@ -244,7 +244,12 @@
              * Select the hovered option.
              */
             enterPressed() {
-                if (this.hovered === null) return
+                // If an enter key pressed and there's nothing hovered,
+                // emit an event for custom handling (e.g. submit requests)
+                if (this.hovered === null) {
+                    this.$emit('submit')
+                    return
+                }
                 this.setSelected(this.hovered)
             },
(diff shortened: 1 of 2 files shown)


jtommy:
A new event is not a problem but I I would like understand a real-world use case

kvnkuang:
Hi @jtommy,

Thanks for your prompt response!

Here is the case that we encountered. We implemented a search bar with `Taginput`. Users will search and select multiple items from a list and when they're ready, they need to press a button to submit a search request. However, we notice that many users will just try to press enter multiple times and hope to initiate search this way. It fits their expectation as many websites would allow them to search by pressing enter.

The real problem in our case is that we cannot just register a listener to `keydown.enter` because we cannot differentiate between these two scenarios: 1) user press enter to select from the autocomplete dropdown and 2) user press enter to submit a search request.

Having `Autocomplete` and `Taginput` emit a `submit` event will allow us to handle the 2) scenario easily.

Thanks again for considering,
Kevin

jtommy:
I understand the problem but i think the solution is quite tricky

Issue #1330, opened 2019-04-17
Tooltip on low resolution


shelru:
![image](https://user-images.githubusercontent.com/9306996/56282815-11aee380-6121-11e9-8fd0-5f374feebf85.png)
![image](https://user-images.githubusercontent.com/9306996/56282845-268b7700-6121-11e9-9eb5-6e19e2da4c12.png)

From official example.
https://buefy.org/documentation/table/
Custom header's.

shelru:
So, i think you must refactor some code, because original tooltips(also from buefy work's well).
https://github.com/buefy/buefy/pull/1328 also.
Create new branch, where we can download beta .js for testing.

jtommy:
I close it because I have to refactor tooltip component soon I hope https://github.com/buefy/buefy/issues/1316

shelru:
![image](https://user-images.githubusercontent.com/9306996/56340102-e8925f80-61c1-11e9-95e5-aed7503ffe73.png)
Also if you can, refactor this too.

PR #1369, opened 2019-05-01
Tooltip on table header fix

shelru:
Before:
image
image

After:
image

### DIFF
--- a/src/scss/components/_table.scss
+++ b/src/scss/components/_table.scss
@@ -6,7 +6,10 @@
         margin-bottom: 1.5rem;
     }
     @include touch {
+        // assign auto to y too
         overflow-x: auto;
+        // asign to y - visible
+        overflow: visible;
     }
 }
 


shelru:
Related to https://github.com/buefy/buefy/issues/1330

jtommy:
It's only a workaround. I want to improve tooltip component in order to append it in the body

shelru:
Again? Seriously? This hotfix is not directed at tooltip, but on all elements 
 in the table headers. You can read about overflow-x, but seem's you never get it, again.

shelru:
image
Your corrections, improvments and reviews sometimes take more than a year. Isn't it better to fix now, and when will new tooltip revert commit(If even then you don't understand the problem with overflow-x)?

jtommy:
You could make a PR improving tooltip component if you want.
Could you provide me an example without tooltip that this PR fix?

shelru:
As I said earlier, in a similar way you are only slowing down the development of the project on the background of others. We must approve more PR  asap and check more issue's, nothing is ideal initially, in development process we can review errors to correct and modify components, making them more convenient and customizable.

> You could make a PR improving tooltip component if you want.

I still do not see any reasons for PR, as I wrote above, some PR is not considered for a long time.

> Could you provide me an example without tooltip that this PR fix?

Make some custom slot and on hove show custom tooltip(not component from buefy), and you gonna get same issue on low resolutions.

jtommy:
I already told you, please respect us.
We love contributions, for example datepicker was born from a PR, but they needs a review to check the Buefy philosophy respect and to ensure the proper operation.
Buefy is a free open source project and we can dedicate a part of our free time.

Anyway i might be in wrong and so I asked you to let me know a reason to accept this PR.

PR #1378, opened 2019-05-03
Use aria-hidden=true instead of only aria-hidden

angelabad:
Please use aria-hidden="true" for better accesibility, take a look at:

- https://dequeuniversity.com/rules/axe/3.2/aria-valid-attr-value
- https://developers.google.com/web/fundamentals/accessibility/semantics-aria/hiding-and-updating-content

Thanks in advance!


jtommy:
Thanks!

PR #1394, opened 2019-05-09
Fix #1393 b-input emit change event

### LINKED ISSUE #1393: b-input should emit change event
The [change event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event) fired by the textarea / input should be passed on to the component-level via `$this.emit('change')`. As of fec335898e7b1e5c03847a9fae801f45c5548c3e this is not the case.

bejohu:
See #1393

### DIFF
--- a/src/components/input/Input.vue
+++ b/src/components/input/Input.vue
@@ -11,6 +11,7 @@
             :value="computedValue"
             v-bind="$attrs"
             @input="onInput"
+            @change="onChange"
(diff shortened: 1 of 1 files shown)


jtommy:
Why do you need it ? You might use `@change.native` and also you're emitting change event without date

bejohu:
Thanks, `@change.native` seems to work just fine.

PR #1396, opened 2019-05-10
Fix Mobile referencing dropdown component

nchutchind:
On mobile, the $refs.dropdown component doesn't exist, but the keyPress event handler is still referencing it.
Since the only key event being cared about right now is the Escape key, it may make more sense to just not create or destroy the keyPress event listener, but this change touched the least stuff.


jtommy:
Thanks! Anyway I'll refactor something in datepicker

PR #1399, opened 2019-05-11
feat: Add footer slot on autocomplete component

LeoMouyna:
As header slot, a footer slot as been added at the end of the list.
It can be useful to add a "load more" item at the end.

No issue to see.


jtommy:
Good idea but I think it's better to move it after empty slot, don't you ?

LeoMouyna:
@jtommy You are probably right, I made the change :)

jtommy:
Thanks!

PR #1412, opened 2019-05-17
Added FontAwesome check-circle icon

filippofortino:
FontAwesome check-circle icon is also used, like in the notification component.


jtommy:
@filippofortino thanks!

PR #1428, opened 2019-05-23
Add a way to overwrite the error message

jankeesvw:
For my current project we use the [`pattern`](https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/Form_validation#Validating_against_a_regular_expression) feature of HTML. When using this feature we want be able to customize the error a user sees when the error occurs.

That's why I added a property to the `FormElementMixin.js`.

This is what it looks when you use this code:

```
<b-field>
    <b-input placeholder="User handle (custom validation for only lowercase)"
      type="text"
      required
      validation-error="Only lowercase is allowed"
      pattern="[a-z]*">
    </b-input>
</b-field>
```

### DIFF
--- a/docs/pages/components/input/api/input.js
+++ b/docs/pages/components/input/api/input.js
@@ -79,6 +79,13 @@ export default [
                 default: '—'
             },
             {
+                name: '<code>validation-message</code>',
+                description: 'The message which is shown when a validation error occurs',
+                type: 'String',
+                values: '—',
+                default: '—'
+            },
+            {
                 name: 'Any native attribute',
                 description: '—',
                 type: '—',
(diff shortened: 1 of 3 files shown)


jankeesvw:
Is there anything I have to or can do to make the build green?

jtommy:
Why aren't you using message prop of bField?

jankeesvw:
Thanks for the comment, in my tests it always seemed to render the message. Even when there was no validation error.

jtommy:
Anyway I might accept it but please change the prop name to validation-message (same name of native attribute)

jankeesvw:
I updated this PR to use the attribute name you suggested @jtommy.

> Yes, you should apply a message only when input is not valid

Yes, that's possible but in that case you have to write code to check the error in your app. With this PR you can use the `pattern` attribute to write your validations.

jankeesvw:
Ow, and I added the attribute to the documentation. ✨

jtommy:
Thanks!

PR #1432, opened 2019-05-24
Fix incorrect props for rounded controls

service-paradis:
Replace `rounded` with `controls-rounded` in the documentation samples

### DIFF
--- a/docs/pages/components/numberinput/examples/ExSimple.vue
+++ b/docs/pages/components/numberinput/examples/ExSimple.vue
@@ -13,7 +13,7 @@
         </b-field>
 
         <b-field label="Rounded">
-            <b-numberinput rounded></b-numberinput>
+            <b-numberinput controls-rounded></b-numberinput>
         </b-field>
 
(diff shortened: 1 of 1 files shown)


jtommy:
It's a breaking change, I'll think about it later

jtommy:
It's about input, there is "Customization" for it

service-paradis:
Hum, I'm not sure I understand your comment.

Tjhere was a typo in one of the example in the documentation for `b-numberinput`. If you look the first set of examples in https://buefy.org/documentation/numberinput, the rounded controls example is incorrect.

This Pull Request is only to fix the error in the documentation.

jtommy:
It's ok, rounded prop applies "rounded" style to input while controls-rounded only to buttons

service-paradis:
Ho, I'm sorry. There is 2 different properties. `rounded` and `controls-rounded`.
`rounded` property is missing at the bottom of the page:
image

jtommy:
You're right and I'll add it, thanks

Issue #1410, opened 2019-05-17
Datepicker: years list after pick year


gecheru:
### Overview of the problem

**Buefy** version: 0.7.5
**Vuejs** version: 2.6.6
**OS/Browser**: windows 10, chrome 74.0.3729.131

### Description

List of available years in select is changed after pick any year to "picked year + 3".
If set max-date the years list will increase to max-date value by 3 year after many picks..

### Steps to reproduce

1. Click on years select
2. Pick year
3. Click on years select again

You can try it on docs page. It happens in all datepickers
https://buefy.org/documentation/datepicker/

### Expected behavior

Years list will be not changed

### Actual behavior

Years list is changed to picked year + 3

jtommy:
At the moment the year select contains only a max number of options in order to be light.
Maybe I might add a new prop to change default increment value (+3)

gecheru:
Sorry for my english. I'll try to explain with screenshots:
1. Click on years select. There max year option is 2022.
![image](https://user-images.githubusercontent.com/41947777/57968191-e44b9480-796f-11e9-82f4-2e1aebf4b9ad.png)
2. Pick for example 2014. And click on years select again. There max year option is 2017. It's not previous max value 2022
![image](https://user-images.githubusercontent.com/41947777/57968241-6f2c8f00-7970-11e9-8fb9-f68910821154.png)
3. Pick for example 2017. And max year value is 2020. 
![image](https://user-images.githubusercontent.com/41947777/57968302-f417a880-7970-11e9-8dc5-bc98740472cf.png)

jtommy:
@gecheru because it shows current year + 3 as max value in order to contains about only 100 options in the year select

gecheru:
@jtommy but if I want always to show +- 10 years in select. What I should to do? 
https://codepen.io/gecheru/pen/qGjQda

jtommy:
I might add a new prop or new behaviour (breaking change) in the next releases

adriflorence:
@gecheru not sure if still needed but for that purpose you can just set in your data: 
`data() {
            const today = new Date()
            return {
                date: new Date(),
                minDate: new Date(today.getFullYear() - 10, today.getMonth(), today.getDate()), // ten years ago today
                maxDate: new Date(today.getFullYear() + 10, today.getMonth(), today.getDate()) // ten years from today
            }
}`

and reference minimum and maximum in your datepicker component like so:

`<b-datepicker :min-date="minDate"
[...]

PR #1444, opened 2019-06-03
Datepicker years dropdown updated to show number of years based on prop

msonowal:
Datepicker years dropdown updated to show number of years based on prop.
which is currently `3` now exposed it via prop so that people can customize it for now.

Addresses #1410 

ping @jtommy

### DIFF
--- a/docs/pages/components/datepicker/api/datepicker.js
+++ b/docs/pages/components/datepicker/api/datepicker.js
@@ -177,7 +177,7 @@ export default [
                 default: '-'
             },
             {
-                name: '<code>type</code>',
+                name: '<code>years-to-show</code>',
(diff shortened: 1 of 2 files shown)


jtommy:
@msonowal Thanks! It's a good solution but I don't know if it's the best way.
I thought to add a prop that contains number of years to show, for example prop = 10, the select contains selected year + 5 (max) and selected year -5 (min), what do you think about it ? Or two prop in order to have a different increment

jtommy:
@msonowal Did you check it ? It doesn't work well

jtommy:
Close https://github.com/buefy/buefy/commit/2a5c6bde08a6f8dc92aa55485c0053cb249c0234

PR #1451, opened 2019-06-04
Added border-radius to clockpicker. Fixes #1450

### LINKED ISSUE #1450: Clockpicker no border-radius (style inconsistency)
### Overview of the problem

Clockpicker doesn't have rounded corners (no border-radius) where as other similar components do, such as datepicker.

**Buefy** version: [0.7.7]
**Vuejs** version: [2.6.10]
**OS/Browser**: All

### Description

See above

### Steps to reproduce
See documentation examples.
### Expected behavior

Style should be consistent across components

### Actual behavior

Style is inconsistent.

atinybeardedman:


jtommy:
Thanks! Do you think we need border radius definition on card-header too ?

atinybeardedman:
The card header specifically in the clockpicker that I added styles for, or the more general card component? Sorry just not sure which one you're referring to

jtommy:
Card header in clockpicker

jtommy:
Thanks! I think to remove this lines anyway
``` 
    .card-header {
        border-top-left-radius: $radius;
        border-top-right-radius: $radius;
    }
 ```

atinybeardedman:
I think those were actually needed...because of the way the header was positioned, if you leave them off the component itself has rounded corners but the card doesn't and so the top edges of the card are on top of the rounded corners of the component and it looks weird. I added them to make it consistently rounded.

jtommy:
Ok, I'll restore them

Issue #1463, opened 2019-06-12
Implement "open on focus" on taginput elements


kartsims:
### Overview of the problem

The `taginput` component doesn't have a `openOnFocus` property.

### Description

It would be nice to implement the same feature as for `autocomplete`, ie. when clicking the `taginput` element, the dropdown can be displayed automatically.

I can give it a try if need be, but is it something you would like to see implemented?

jtommy:
You can already use that prop

kartsims:
Shouldn't it be documented then? PR sent for approval

jtommy:
It Is already in the docs, take a look at the bottom

kartsims:
You are right, it is definitely not clear to me, and I bet it is for others too...

Proposals to make it clearer:
- link to the autocomplete props ("Any other native attribute or [Autocomplete prop](#)")
- state it next to the `autocomplete` prop description (ie. "Add autocomplete feature (if `true`, Autocomplete props may be used too)")

jtommy:
Ok, you might make a PR to improve it.
Thanks

kartsims:
Done, thank you for the great library, happy to contribute to this slight improvement 👍

PR #1465, opened 2019-06-12
Added documentation for Autocomplete props on 'tag-input' elements

kartsims:
I opened #1463 because the docs didn't state that `open-on-focus` was available as a property for `tag-input` elements


kartsims:
Commit updated to reflect the discussion in #1463

jtommy:
Thanks

Issue #1462, opened 2019-06-11
Table adds width to other column when a column is toggled invisible and visible again


utkarsh-rg:
### Overview of the problem

**Buefy** version: 0.7.6
**Vuejs** version: 2.6.10
**OS/Browser**: Chromium/Ubuntu 16.10

### Description

When there is a column with some fixed width, on making any other column ```visible: false```, and then ```visible: true``` back again, the width applies to other column as well. 

### Steps to reproduce

1. Create a table where columns can be toggled invisible.
2. Give the last column a width, eg: 1px;
3. Toggle some other column invisble.
4. Toggle that column visible again. Check dev tools, the width will be applied to 2 columns.

### Expected behavior

On toggling a column invisble and back visible, the width should not go to other column.

### Actual behavior

The width applies to other columns.

### Screenshots
A table with the last column with width 1px:
![Screenshot from 2019-06-11 14-59-13](https://user-images.githubusercontent.com/31611229/59261611-6e2f0a80-8c5b-11e9-9401-af2c8e1c8982.png)

After toggling the first column(ID) invisible.
![Screenshot from 2019-06-11 14-59-31](https://user-images.githubusercontent.com/31611229/59261612-6e2f0a80-8c5b-11e9-8899-403afb836f9b.png)

After toggling the column back to visible. There are 2 columns with that width now. This messes up the layout. (Check the Gender column before and after the toggle)
![Screenshot from 2019-06-11 14-59-47](https://user-images.githubusercontent.com/31611229/59261614-6ec7a100-8c5b-11e9-8f0a-94e032bdfbd8.png)

PR #1483, opened 2019-06-19
Fix for #1462

service-paradis:
### Fix for #1462

This reset width styling if width is undefined


service-paradis:
@afwn90cj93201nixr2e1re
I think this should be done in another PR.
I can see if I can do it

jtommy:
@service-paradis Thanks!

PR #1484, opened 2019-06-19
Fix cancelling long press on Chromium

### LINKED ISSUE #1480: NumberInput does not stay set to required value when v-model changes
### Overview of the problem

**Buefy** version: [0.7.7]
**Vuejs** version: [2.6.10]
**OS/Browser**:
MacOS and Windows 10/Chrome

### Description
The number input control relies on a mouse event to stop decrementing/incrementing of values. This allows users to long press the control to change values efficiently, however, it also causes a problem when changing values programmatically as a change event gets called but no stop event seems to be triggered when a button is disabled.

https://codepen.io/anon/pen/YopXRw?&editable=true&editors=101

numInputBug

### Steps to reproduce
1. Set max and min for the number input (e.g. 1 and 7) and bind value using v-model
2. Long press the minus button to decrease the value of the number to the min value (e.g. 1) so the decrement button is disabled
3. Try to set the bound v-model value to a different number (e.g. 3) without using the number input control

**Result:** The number input will change to the new value but then decrement all the way back down to the min value.

~~(May need to repeat steps 1-3 a few times as results aren't super consistent but should be easy enough to reproduce)~~ EDIT - can consistently reproduce results when long pressing to increment/decrement values

Same thing happens when incrementing all the way to max (instead of min) and then changing values. The number will change and increment all the way to the max value.

### Expected behavior

When the v-model value is changed, the number input value will be set to the required value without incrementing/decrementing to the max/min value by itself.

### Actual behavior
When the v-model value is changed, the number input value is set to the required value but then auto increments/decrements all the way to the max/min value

yxngl:
Fix #1480 by moving listeners to stop long press up to the control, since Chromium based on browsers do not produce mouse event on disabled button.

### DIFF
--- a/src/components/numberinput/Numberinput.vue
+++ b/src/components/numberinput/Numberinput.vue
@@ -1,17 +1,18 @@
 <template>
     <div class="b-numberinput field" :class="fieldClasses">
-        <p class="control">
+        <p
+            class="control"
+            @mouseup="onStopLongPress(false)"
+            @mouseleave="onStopLongPress(false)"
+            @touchend="onStopLongPress(false)"
+            @touchcancel="onStopLongPress(false)">
             <button
                 type="button"
                 class="button"
                 :class="buttonClasses"
                 :disabled="disabled || disabledMin"
                 @mousedown="onStartLongPress($event, false)"
-                @mouseup="onStopLongPress(false)"
-                @mouseleave="onStopLongPress(false)"
                 @touchstart.prevent="onStartLongPress($event, false)"
-                @touchend="onStopLongPress(false)"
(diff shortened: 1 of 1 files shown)


yxngl:
Note the current problem can be seen and tested using "Min - Max" example section.

1. Input a number like 3 in Vote
2. Long press minus button to zero
3. Input another number like 4
4. It keeps decreasing to zero

afwn90cj93201nixr2e1re:
@yxngl can you fix minus input too?
Set min prop to -1000.
Input for example 1000.
Double click on it, starty typing minus.
It's gonna change it to min val.

PR #1488, opened 2019-06-20
Added support to `nuxt-link` and it's aliases to `b-button`

gigioSouza:


afwn90cj93201nixr2e1re:
Maybe you should execlude package-lock.json from this commit? Or just hard revert commit and make new.

jtommy:
Thanks!

PR #1496, opened 2019-06-24
Add typings for programmatic use of dist components

marc1404:
This pull request adds simple module declarations for the programmatic use of components outside of Vue.

# Current Situation
```ts
/* example.ts */
import { Toast } from 'buefy/dist/components/toast';
// types for the import will not be found
```

# Target Situation
```ts
/* buefy/types/index.d.ts */
declare module 'buefy/dist/components/toast' {
    export { Toast };
}

/* example.ts */
import { Toast } from 'buefy/dist/components/toast';
// TypeScript is able to resolve the type 😄 
```

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -15,6 +15,26 @@ declare module 'vue/types/vue' {
     }
 }
 
+declare module 'buefy/dist/components/dialog' {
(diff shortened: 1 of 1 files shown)


jtommy:
Thanks

PR #1502, opened 2019-06-25
Allow user to add classes to modal

CobraClutch:
The addition of a fullscreen option for modals is great. Here's an idea to allow repurposing of the modal component. By adding custom classes, a user will be able to style individual modals differently.


jtommy:
Good idea!

PR #1505, opened 2019-06-26
Remove invalid type declarations

marc1404:
@jtommy I'm sorry these type declarations were not valid and didn't show up as errors in my IDE. Apparently, it's not possible to augment the global scope from within a module. Which kind of makes sense since you don't want to pollute the global type definitions from within.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -15,26 +15,6 @@ declare module 'vue/types/vue' {
     }
 }
 
-declare module 'buefy/dist/components/dialog' {
-    export { Dialog };
-}
-
-declare module 'buefy/dist/components/modal' {
-    export { ModalProgrammatic };
-}
-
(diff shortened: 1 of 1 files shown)


jtommy:
Is there a workaround to avoid build error with last release?

marc1404:
It showed up as an error for me but didn't break the build. I'm not aware of a workaround, sorry.

jtommy:
@marc1404 Please, take a look at https://github.com/buefy/buefy/commit/e5248fb72373aed416bc18238db93b49c4649fe0#r34079484

service-paradis:
Just opened an Issues for this (#1507). Did not see this PR before.

service-paradis:
The only workaround I found working temporarily is to add `"skipLibCheck": true` to the `compilerOptions` of `tsconfig.json`

PR #1508, opened 2019-06-27
Add missing properties in TS BuefyConfig type. Fix property spelling in config.js

trebler:


jtommy:
Thanks, I totally forgot to update TS config as well

Issue #594, opened 2018-02-08
Add delay to tooltip


SamUndefined:
Other vue-tooltip plugins and jQuery, for example, allow the developer to add a delay which controls the amount of time the user must hover an element before its tooltip pops up. Perhaps this feature could be part of Buefy, maybe as a prop in the form of milliseconds:

```
<b-tooltip
    label="My delayed tool tip"
    position="is-right"
    delay="1000"
    >
```

By the way, thanks for this excellent component package!

gierloff:
Great idea!

laegeloen:
Any news on adding delay to tooltips? :) please?

emanuelmutschlechner:
@laegeloen 

I've used a simple CSS workaround using the `transition-delay` property. It works fine if you want to change it globally. 

```css
.tooltip:not([data-label=""]):hover:before, 
.tooltip:not([data-label=""]):hover:after {
    transition-delay: 1s;
}
```

To apply a delay to a specific tooltip you could just add a custom class and set that on `<b-tooltip>`.
```css
.tooltip.is-slow:not([data-label=""]):hover:before, 
.tooltip.is-slow:not([data-label=""]):hover:after {
    transition-delay: 3s;
}
```

> Demo
https://codepen.io/anon/pen/XxjxRX

laegeloen:
> @laegeloen
> 
> I've used a simple CSS workaround using the `transition-delay` property. It works fine if you want to change it globally.
> 
> ```css
> .tooltip:not([data-label=""]):hover:before, 
> .tooltip:not([data-label=""]):hover:after {
>     transition-delay: 1s;
> }
> ```
> 
> To apply a delay to a specific tooltip you could just add a custom class and set that on `<b-tooltip>`.
> 
> ```css
> .tooltip.is-slow:not([data-label=""]):hover:before, 
> .tooltip.is-slow:not([data-label=""]):hover:after {
>     transition-delay: 3s;
> }
> ```
> 
> > Demo
> > https://codepen.io/anon/pen/XxjxRX

Thanks! :) Kind of you!

ffxsam:
@jtommy Awesome, I was just looking for this! In which version will this be available?

jtommy:
0.7.9 :)

PR #1512, opened 2019-06-29
Tooltip delay prop to delay it (#594)

service-paradis:
Added a prop for the tooltip component to manage the delay before it appears.
Ref: #594


jtommy:
Thanks! I merge it waiting to refactor the component

PR #1514, opened 2019-06-29
Limit scope of css for fullscreen modals

cybercode:
Only override the `.modal-card` css for the first descendant of the fullscreen
modal (> .animation.content > .modal-card).

The current css will affect any modal opened from within the full screen modal, causing it to be full height and width, with the width actually constraint (by default) to `max-width: 960px` by the `BModal width` prop.


jtommy:
@cybercode I don't undertand that PR, adding it fullmodal doesn't work (for example let's see the example in the docs)

cybercode:
@jtommy My bad. Was testing in a different repos, and copied over the wrong version. I have force pushed the correct version (so there is only one change commit) , and verified via `npm run dev` in this repos.

The purpose and general approach is the same. Limit the "full screen" css to only the first `.animation-content` and `.modal-card` under the full screen modal.

PR #1518, opened 2019-06-30
Fix an error during Tooltip merge

service-paradis:
Looks like there was an error during the latest merge of the tooltip component.
Props were duplicated inside the position prop. It was causing some errors.


jtommy:
Yes, I was fixing but i'll merge it

jtommy:
Yes, I was fixing but I didn't push it :)

PR #1520, opened 2019-06-30
Made Tab Items emit a click event

Nerderer:
Tab Items will now emit an event when being clicked. Tab Items also have an optional value prop, that will passed with the click event.

### DIFF
--- a/src/components/tabs/TabItem.vue
+++ b/src/components/tabs/TabItem.vue
@@ -9,6 +9,10 @@ export default {
         visible: {
             type: Boolean,
             default: true
+        },
+        value: {
+            type: [String, Number, Boolean, Object, Array, Symbol, Function],
+            default: null
         }
     },
     data() {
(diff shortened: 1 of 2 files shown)


jtommy:
Why do you need it ? Isn't enough input event?

Nerderer:
This let's you easier handle the presentation of the tab content yourself, while still having workable tabs.

Let's say I have tabs that works more as filters (real life scenario) than actually showing different views. Instead of wrapping similar content in every tab, I can use the same code under the tabs, that gets different data depending on the tab I clicked. I could, in many cases, just the tab index, but that is probably not always the case.

Nerderer:
`<b-tabs>
    <b-tab-item label="Active" value="active" @click="getData" />
    <b-tab-item label="Inactive" value="inactive" @click="getData" />
</b-tabs>

<div>
{{ filteredData }}
</div>
`

jtommy:
@Nerderer I undestand about it but i don't see an added value compare to `input` event

Issue #1522, opened 2019-07-01
Feature request - responsive tabs -> accordion


wrabit:
I would really like to see tabs collapse to accordion so for mobile they become much more usable. I am working on a branch that adds this just wanted to check the likelihood something like this would be merged before I continue:

Animation:

![Screen Recording 2019-07-01 at 15 30 40 (2)](https://user-images.githubusercontent.com/5918271/60441144-aac8a300-9c16-11e9-93fb-9ca11a5a6eb4.gif)

jtommy:
@wrabit Really nice!
I'll review your PR when you are ready

PR #1523, opened 2019-07-02
Work in progress for tab accordions #1522

wrabit:
First working concept of the mobile responsive tabs. Reading through the requirements, I realise it doesn't fully comply with principles. But I need to start using this in a project ASAP and don't the time right now to revise. In the meantime could you review it and let me know the concerns. Thanks.


jtommy:
I've just take a quick look and I think it's good, thanks!
To be honest i don't like animation using js, do you think to rewrite it using css ?

wrabit:
Javascript is used to determine the height of the next tab-item but it's still css that's performing the animation. I could look at utilising a pure css way animation between height 0 and auto of the tab-item. The tricky bit is to retain the current desktop behaviour.

I have noticed this branch is making the desktop horizontal slide animations buggy so I will look at that.

service-paradis:
@wrabit 
Maybe you can use the `b-collapse` component to help you create the mobile accordion?
I think this component would stay a lot cleaner since the toggling and the animation are already managed by it. What do you think?

wrabit:
@service-paradis agree it would probably make it simpler but won't achieve a true accordion transition  as b-collapse is set to show / hide content. The animation is a personal preference as it helps users keep track of which panel closed and which opened.

service-paradis:
@wrabit
But tab is also used to show / hide content. An accordion is kind of wrapper around `b-collapse`.
In fact, It could even be made as a distinct Buefy component. Something like `b-accordion` who would just be a wrapper around `b-collapse`.
Then, you could use this new component for the mobile version of the tabs.
I'm not sure to understand what you mean by not achiving a true accordion transition since you can customize the `b-collpase`animation.

Maybe I'm missing something! Anyway, great works on this, the behavior is really interesting!

wrabit:
A true accordion action looks like this: https://codepen.io/will-abbott/pen/ewKvXp. We need to know the height of the element we are activating so the browser can carry out the collapsing of the current panel and expanding of the next at the same time and same speed. This makes sure we get a smooth change of accordion panels.

There are hacks out there for pure css, but none of them work ideally for example, for the panel we are switching to we could use max-height '100vh' but this makes both panels animate at different speeds which is not a solution.

I've looked at using the b-collapse but unless the js is allowed to determine height, it won't work as a good accordion IMO. In addition, refactoring has some challenges i.e. dealing with the tab header slots within the collapse component trigger.

As I mentioned - js is to determine the height of the next panel, not the animation itself. @jtommy is this really a dealbreaker?

jtommy:
It's no a dealbraker but i don't like it as first approach and looking your code you aren't manipulating attributes using js only for height/width, for example position, display, etc...

jtommy:
I added multiline prop in a previous release.
It's a simple solution to avoid a scrollbar on mobile.
I like your idea but the implementation isn't "standard" for the project, btw you are free to open a new PR ;)

Issue #1321, opened 2019-04-12
Pagination bad behavior


TheTrigger:
Making 4 pages in a b-table we loose the third button. This is not smart since it occupies the same space. So to show the page n 3 i have to first load the second or fourth page. That's annoying if it's async table.

![image](https://user-images.githubusercontent.com/6072718/56042802-2f083a00-5d3c-11e9-843c-06de26ff819c.png)
![image](https://user-images.githubusercontent.com/6072718/56042827-3deeec80-5d3c-11e9-916e-0a3e2dc983bf.png)

Please make buttons hide only if there are only 2 or more pages to hide.
Thanks mate.

**Buefy** version: [0.7.3]

service-paradis:
@TheTrigger 
Created a PR for this (#1529). Maybe it will be part of 0.8.0 since it is a breaking change.

service-paradis:
@jtommy 
This issue can be closed. Fixed in the next release.

PR #1529, opened 2019-07-03
Pagination ellipsis behavior

service-paradis:
Ref: #1321

This fix the case where the ellipsis were shown with only one item to hide.

Display
image
Instead of
image

-----

Display
image
Instead of
image


jtommy:
Thanks, It's a little breaking change so i'll merge it on 0.8.0

jtommy:
Thanks a lot! The new 0.8.0 contains only your PR 🎉

service-paradis:
Loving Buefy as you can see :)
Happy to help you out improving it!

PR #1535, opened 2019-07-09
Fix #1387

### LINKED ISSUE #1387: TabItem Does Not Obey Ordering with v-for and key
### Overview of the problem

**Buefy** version: v0.7.5
**Vuejs** version: v2.6.10
**OS/Browser**: MacOS 10.14.4/Chrome 73.0.3683.103

### Description

When using `v-for` with `key` on `b-tab-item` the order of the tabs is incorrect if items are removed and added back.

### Steps to reproduce

I've made a demo: https://codepen.io/Skinner927/pen/vwNYLp

1. Observe the proper sorted order of the tabs: "April, Bob, Stan".
1. Click the "Remove Bob" button
1. Click the "Add Bob Back" button
1. Observe the order of the tabs is now "April, Stan, Bob" which is incorrect. The `ul` list (at the top under "Sorted:") uses the same `key` mechanism and is properly sorted.

### Expected behavior

For the order of tabs to be in the same order as the underlying array.

### Actual behavior

Because `v-for` is only adding the missing elements, `b-tab-item`'s `created()` method [always appends new items](https://github.com/buefy/buefy/blob/dev/src/components/tabs/TabItem.vue#L46) to the end of the tab list.

service-paradis:
Add a fix to keep defined slots position for tabs.
When tabs are created/changed/deleted dynamically. This also fix the case if we want to use `v-if` to display or not a tab.

Ref: #1387

### DIFF
--- a/docs/pages/components/tabs/Tabs.vue
+++ b/docs/pages/components/tabs/Tabs.vue
@@ -2,6 +2,11 @@
     <div>
         <Example :component="ExSimple" :code="ExSimpleCode"/>
 
+        <Example :component="ExDynamic" :code="ExDynamicCode" title="Dynamic">
+            <p>Items can be created / modified / deleted and will always keep the defined order.</p>
+            <p>You can also use <code>v-if</code> to show (or not) a Tab Item.</p>
+        </Example>
+
         <Example :component="ExPosition" :code="ExPositionCode" title="Position"/>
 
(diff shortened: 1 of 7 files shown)


service-paradis:
Since Steps component is using the same behavior as Tabs, it should be refactored this way as well.

jtommy:
I need time to check the issue well and then to review it

service-paradis:
Let me know if you have any questions!

PR #1549, opened 2019-07-11
Adds the rounded attribute to the pagination docs

tchristerus:
## Proposed Changes

- This adds the rounded attribute to the pagination docs (API)

### DIFF
--- a/docs/pages/components/pagination/api/pagination.js
+++ b/docs/pages/components/pagination/api/pagination.js
@@ -44,6 +44,13 @@ export default [
                 default: '<code>false</code>'
             },
             {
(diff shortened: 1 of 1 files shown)


docs/pages/components/pagination/api/pagination.js
```diff
@@ -43,6 +43,13 @@ export default [
                 values: '—',
                 default: '<code>false</code>'
             },
+            {
+                name: '<code>Rounded</code>',
```
service-paradis: This should be in lowercase:
`name: '<code>rounded</code>',`

PR #1550, opened 2019-07-11
Fix #1542  for Steps

### LINKED ISSUE #1542: Completed step item is not clickable on default
### Overview of the problem

**Buefy** version: [0.7.10]

### Description

The `clickable` API of step item says 

> Item can be used directly to navigate. If undefined, completed steps are clickable while the others are not.

But even if `clickable` is left unspecified, it will still be `false` not `undefined`. https://github.com/vuejs/vue/issues/8704

### Steps to reproduce

Any example in the doc page.

### Expected behavior

Completed step item is clickable.

### Actual behavior

Completed step item is disabled.

service-paradis:
Fixes #1542

## Proposed Changes

- Improve the default behavior for clickable steps
- This is a potential breaking change.
- Should be considered for 0.8.0

### DIFF
--- a/docs/pages/components/steps/api/steps.js
+++ b/docs/pages/components/steps/api/steps.js
@@ -120,16 +120,9 @@ export default [
                 default: '<code>mdi</code>'
             },
             {
-                name: '<code>completed</code>',
-                description: 'Step is considered completed',
-                type: 'Boolean',
-                values: '—',
-                default: 'false'
-            },
-            {
                 name: '<code>clickable</code>',
                 description: `Item can be used directly to navigate.
-                    If undefined, completed steps are clickable while the others are not.`,
+                    If undefined, previous steps are clickable while the others are not.`,
                 type: 'Boolean',
                 values: '—',
                 default: '—'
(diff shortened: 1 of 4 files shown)


jtommy:
Probably it's a breaking change but it a new component too.
By the way next version will contains some breaking

afwn90cj93201nixr2e1re:
That's not a hood idea for change author paradigmas.

Why do you think this is the good idea to change complete to previos? What about if there some disabled steps and you gonna check with condition if a > b, nice, very good, keep going.

service-paradis:
@afwn90cj93201nixr2e1re
I dont really understand what th problem is in your example.

It is only changing a little bit the default behavior for something we often see in Steps component. It is what most people expect from a Steps component. This behavior can be entirely customized anyway using the `clickable` prop on a step item.

This will also e merged in 0.8.0 which is a breaking release.

afwn90cj93201nixr2e1re:
No that's not default way :/
I wanna set completed when i want, not when this is last item at all.

service-paradis:
The prop `completed` was doing nothig before thir PR.
The only thing this PR is doing is to make a previous step clickable to be able to navigate back to it. This behavior can be customized using the `clickable` prop. So you can use it the way you want.

afwn90cj93201nixr2e1re:
изображение

Oh, yes, nothing.

service-paradis:
It was doing nothing because the default value was `false`. There was no way to set this value to `undefined`.
This condition was never met.

PR #954, opened 2018-08-21
<b-progress> component

MeikelLP:
For features request #953

**TODO**
* Coding
  * [x] Write basic component
  * [x] Value updates automatically
  * [x] Colors
  * [x] Sizes
  * [x] Shown values (text)
  * [x] Status icon
* Documentation
  * [x] Add to documentation page  
  * [x] Simple example
  * [x] Complex example
  * [x] API reference

Issue #953, opened 2018-08-21
[Feature request] b-progress


zhaojh329:
### Description

![image](https://user-images.githubusercontent.com/28673295/44409032-250aae80-a594-11e8-8806-f3cc8c63bc76.png)

zhaojh329:
![image](https://user-images.githubusercontent.com/28673295/44439406-04734080-a5f6-11e8-9f74-3284200a7a67.png)

PR #1552, opened 2019-07-12
Component Proposal: Progress Bar

service-paradis:
## Proposed Changes

- Add a new compenent for progress
- Wrapper around Bulma progress (which is Bulma Style over native progress element)

captured

### Works with types
captured

### Works with sizes
captured (1)

### Value customization
captured (2)

### Label can be anything using slot
captured (1)

Ref: #953, #954

### DIFF
--- a/src/components/progress/Progress.vue
+++ b/src/components/progress/Progress.vue
@@ -1,0 +1,108 @@
+<template>
+    <div class="progress-wrapper">
+        <progress
+            ref="progress"
+            class="progress"
+            :class="newType"
+            :max="max">{{ newValue }}</progress>
+        <p
+            v-if="showValue"
+            class="progress-value has-text-white"><slot>{{ newValue }}</slot></p>
+    </div>
+</template>
+
+<script>
+
+export default {
+    name: 'BProgress',
+    props: {
+        type: {
+            type: String | Object,
+            default: 'is-darkgrey'
+        },
+        size: String,
+        value: {
+            type: undefined | Number,
+            default: undefined
+        },
+        max: {
+            type: Number,
+            default: 100
+        },
+        showValue: {
+            type: Boolean,
+            default: false
+        },
+        format: {
+            type: String,
+            default: 'raw',
+            validator: (value) => {
+                return [
+                    'raw',
+                    'percent'
+                ].indexOf(value) >= 0
+            }
+        },
+        precision: {
+            type: Number,
+            default: 2
+        },
+        keepTrailingZeroes: {
+            type: Boolean,
+            default: false
+        }
+    },
+    computed: {
+        isIndeterminate() {
+            return this.value === undefined || this.value === null
+        },
+        newType() {
+            return [
+                this.size,
+                this.type
+            ]
+        },
+        newValue() {
+            if (this.value === undefined || this.value === null || isNaN(this.value)) {
+                return undefined
+            }
+
+            if (this.format === 'percent') {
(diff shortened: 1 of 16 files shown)


jtommy:
Looks nice! I'll review it soon

src/components/progress/Progress.vue
```diff
@@ -0,0 +1,108 @@
...
+                const val = this.toFixed(this.value * this.max / 100)
+                return `${val}%`
+            }
+            const val = this.toFixed(this.value)
+            return val
+        }
+    },
+    watch: {
+        value(value) {
+            this.setValue(value)
```
jtommy: In order to remove the `setValue` call in mounted hook, you might add `immediate` option in the watcher, for example:
```
value: {
    immediate: true,
    handler(value) {
      ...
    }
  }
```

src/components/progress/Progress.vue
```diff
@@ -0,0 +1,108 @@
...
+        }
+    },
+    methods: {
+        /**
+        * When value is changed back to undefined, value of native progress get reset to 0.
+        * Need to add and remove the value attribute to have the indeterminate or not.
+        */
+        setValue(value) {
+            if (this.isIndeterminate) {
+                this.$refs.progress.removeAttribute('value')
```
jtommy: Why do you need to work using refs ?

src/components/progress/Progress.vue
```diff
@@ -0,0 +1,108 @@
...
+export default {
+    name: 'BProgress',
+    props: {
+        type: {
+            type: String | Object,
+            default: 'is-darkgrey'
+        },
+        size: String,
+        value: {
+            type: undefined | Number,
```
jtommy: There is no need to specify undefined type

src/components/progress/Progress.vue
```diff
@@ -0,0 +1,108 @@
...
+    </div>
+</template>
+
+<script>
+
+export default {
+    name: 'BProgress',
+    props: {
+        type: {
+            type: String | Object,
```
jtommy: Is a really correct syntax ? I usually use [String, Object] array syntax.

src/components/progress/Progress.vue
service-paradis: In order to have an indeterminate progress bar, the `value` attribute need to be removed completely fro the component. If we set the value to `null` or something, the value is considered to be 0 and wont be displayed as indeterminate.

Is there another way to remove an attribute without $refs?

src/components/progress/Progress.vue
service-paradis: My mistake! i'll fix it!

src/components/progress/Progress.vue
service-paradis: Your syntax is more common for Vue props. I'll fix it!

src/components/progress/Progress.vue
service-paradis: The problem by using the `immediate` property is the function will be called before vue's mount hook. That means that $refs won't be visible from the watch function.
You can see my next comment about the usage of $refs.

src/components/progress/Progress.vue
jtommy: No, it isn't. I think it's ok at the moment

src/components/progress/Progress.vue
jtommy: You're right!

PR #1554, opened 2019-07-13
List of files is cleared by clicking cancel

it-fm:
In chrome if you click cancel in a window to select files the previous list is clean, is it expected behavior?

## Proposed Changes
- Removing the line that clears this.newValue behavior does not occur


jtommy:
@it-fm I remember that main idea was to respect the browser behaviour

PR #1555, opened 2019-07-13
Problem selecting a deleted file again

it-fm:
In chrome, if you select two files individually, then delete the second file and again try to send it nothing happens

## Proposed Changes

- check if the file stored in the input exists in the variable `value`, if there is no clean input

### DIFF
--- a/src/components/upload/Upload.vue
+++ b/src/components/upload/Upload.vue
@@ -72,17 +72,8 @@ export default {
          */
         value(value) {
             this.newValue = value
-            // IE +11 findIndex
-            function findFileInArray(array, fileName) {
-                for (let i = 0; i < array.length; ++i) {
-                    if (array[i].name === fileName) {
-                        return i
-                    }
-                }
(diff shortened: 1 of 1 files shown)


src/components/upload/Upload.vue
```diff
@@ -73,7 +73,8 @@ export default {
         value(value) {
             this.newValue = value
             if (!this.newValue ||
-            (Array.isArray(this.newValue) && this.newValue.length === 0)) {
+            (Array.isArray(this.newValue) && this.newValue.length === 0) ||
+            this.newValue.findIndex((x) => x.name === this.$refs.input.files[0].name) < 0) {
```
jtommy: Please a IE11+ compatible function

it-fm: Ok, I ended up using it because I found the `findIndex` in use in the `Steps.vue` file, in case it is an error I can send a pull in Steps too

PR #1567, opened 2019-07-16
Fix slanted x when attached props is used

service-paradis:
## Proposed Changes

- Adjust width for the delete button using `attached` prop

The lines for the `x` are related to width and height of the direct parent. `$taginput-height` is not necessarily equals to the default width for the `x` parent.

### Current behavior without this fix
image
If you look closely, the `x` is slightly slanted (default parent size is 2em x 1.7em)

### Behavior after the fix
image


jtommy:
Thanks!

PR #1568, opened 2019-07-16
Add new style options for switch

service-paradis:
Proposed changes:
- Give a more squarish default style
- Add a rounded prop
- Add a outlined prop

Breaking changes:
- The default style is not rounded anymore

I think it fits more with the default Bulma/Buefy styling.

Probably need to be discussed before merging (even in a breaking release)

captured

### DIFF
--- a/docs/pages/components/switch/Switch.vue
+++ b/docs/pages/components/switch/Switch.vue
@@ -6,6 +6,8 @@
 
         <Example :component="ExSizes" :code="ExSizesCode" title="Sizes" vertical/>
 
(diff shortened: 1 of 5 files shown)


jtommy:
It's ok for me, but at the moment i prefer rounded as default

service-paradis:
Sure, we can change the default back to rounded!
I made it this way since it is the only rounded component in Buefy (and in Bulma)

jtommy:
You're right, we might change it later

Issue #520, opened 2018-01-09
[Feature Request] Add functionality to listen to mouseover for a table row


modstore:
### Overview of the problem

**Buefy** version: [0.6.2]
**Vuejs** version: [2.5.13]

### Description

I want to be able to show/hide based on mouseover of a table row, is this possible at all?

Thanks in advance.

rafaelpimpa:
Not currently

modstore:
Updated to "feature request". Thanks for the great work so far.

rafaelpimpa:
Have to test if there will be any performance downsides.

service-paradis:
@jtommy 
I think we can close this issue

PR #1569, opened 2019-07-17
feat: add mouse hover events to b-table

afm-sayem:
Why we need this:
We are trying to add an action bar on hover of table row (similar to Gmail's inbox shows a bar on hovering over an email). Currently it's not possible because mouse enter/leave events are not emitted from b-table.  

Implements #520
## Proposed Changes

- Emit a mouseenter event when mouse enters a table row
- Emit a mouseleave event when mouse leaves a table row

### DIFF
--- a/docs/pages/components/table/api/table.js
+++ b/docs/pages/components/table/api/table.js
@@ -417,7 +417,7 @@ export default [
 
             },
             {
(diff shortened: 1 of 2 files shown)


jtommy:
Thanks

PR #1572, opened 2019-07-17
Add a  focusable prop to dropdown-item

### LINKED ISSUE #1437: Custom dropdown items are always focusable.
### Overview of the problem

**Buefy** version: 0.7.6
**Vuejs** version: 2.6.10
**OS/Browser**: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/74.0.3729.169 Safari/537.36

### Description

Custom dropdown items are always focusable. But it should not be focusable sometimes. For example, when the dropdown is used to show a login form, the first focusable item should be the input inside the item, not the item itself.

### Steps to reproduce

1. Open the document page and see the 'Content and position' part.
2. Press 'Login' dropdown button in the example.
3. Press 'Tab' button on the keyboard.
4. The item is now focused.

image

### Expected behavior

Custom dropdown item can be focusable or not, depending on a property passed in.

### Actual behavior

Custom dropdown items are always focusable.

service-paradis:
Fix #1437

### Proposed changes:
The default behavior for dropdown item right now is to be focusable.
This prop will keep the default behavior, but can be disabled.

The login example in the doc ("Content and position" from https://buefy.org/documentation/dropdown) is a good example. In this case, we do not really expect the item to get focus. Instead, we want the first input to get focus when using tab.

### Current behavior when clicking on the dropdown and using tab key:
captured

### New behavior when clicking on the dropdown and using tab key (when `focusable` is false):
captured (1)

### DIFF
--- a/docs/pages/components/dropdown/Dropdown.vue
+++ b/docs/pages/components/dropdown/Dropdown.vue
@@ -9,6 +9,7 @@
 
         <Example :component="ExContentPosition" :code="ExContentPositionCode" title="Content and position" paddingless>
             <p>Add the <code>custom</code> prop to the item to add <strong>any type of content</strong>.</p>
+            <p>Add the <code>:focusable="false"</code> prop to the <code>dropdown-item</code> if you dont want it to be focusable.</p>
         </Example>
 
         <Example :component="ExHasLinkDisabled" :code="ExHasLinkDisabledCode" title="Links within" paddingless>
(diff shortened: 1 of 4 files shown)


jtommy:
Great! 
I'm thinking of adding it as breaking change too (default `false`)

Issue #1573, opened 2019-07-17
Improve the pagination component flexibility


modularcoder:
### Description

The current pagination component doesn't have much customization options. I think we can improve the customization by adding slots support. So the component consumer can leverage of the dynamic pages range generation (which again might be customized e.g. user might want to display different number of pages nearby the active page), but be able to pass into the pagination the desired class buttons/components.

@jtommy  what do you think about this?

### Why Buefy need this feature

Will improve the pagination component reusability

jtommy:
It's so simple component. What do you need to customize?

modularcoder:
@jtommy here are the use cases that come to my mind

- someone might not want having pagination-previous / pagination-previous buttons. 
- use custom icons for the pagination items
- instead of showing only 2 page numbers around the active page, show more pages
- use custom classes for button components or other tags

I think slots can solve this use cases with keeping the backwards compatibility

jtommy:
Ok, I think the major part of these use cases might be fulfilled using new props. 
By the way I'm waiting your PR 😃

jtommy:
Close https://github.com/buefy/buefy/issues/1575

PR #1575, opened 2019-07-18
[Pagination] Added props for adding buttons before and after current page

lsv:
Fixes some of #1573 

I wasnt sure what I should call the props, so it became before and after

I also updated the documentation for pagination

I have made it on dev branch so #1529 (with breaking changes) also will be included. (can also make it to another branch if needed)

But otherwise its not a breaking change, as the props are default as 1

Screenshot from 2019-07-18 07-35-03

Screenshot from 2019-07-18 07-38-30

### DIFF
--- a/docs/pages/components/pagination/api/pagination.js
+++ b/docs/pages/components/pagination/api/pagination.js
@@ -16,15 +16,15 @@ export default [
                 default: '<code>20</code>'
             },
             {
-                name: '<code>before</code>',
-                description: 'Items to show before current page',
+                name: '<code>range-before</code>',
+                description: 'Number of pagination items to show before current page',
                 type: 'Number',
                 values: '—',
                 default: '<code>1</code>'
             },
             {
-                name: '<code>after</code>',
-                description: 'Items to show after current page',
+                name: '<code>range-after</code>',
+                description: 'Items to paginatation items to show after current page',
                 type: 'Number',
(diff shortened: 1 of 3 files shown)


service-paradis:
@lsv 
Good job! Not sure avout the prop name, but @jtommy will see if it needs to be renamed.

1 thing to note in your second screenshot: There are 2 buttons for page 20

lsv:
@service-paradis yep - I noticed the two 20 pages, and I fixed it in the latest commit :)

modularcoder:
I would suggest adding some sort of prefix for the prop names, because IMO just `before` and `after` are not fully expressing the definitive meaning also might be confused with "previous" "next" buttons.

What about something from this?

`num-pages-before` / `num-pages-after`
`num-pages-before-active` / `num-pages-after-active`
`range-before` / `range-after`
`range-before-active` / `range-after-active`
`pages-range-before` / `pages-range-after`
`num-pages-range-before` / `num-pages-range-after`

Sorry for spamming :)

lsv:
@modularcoder sure any of them are great, but... they are long :)
I would properly prefer range-before / range-after

@jtommy what would you call these attributes?

jtommy:
@lsv ok!

lsv:
Documentation also updated with the new attribute names

modularcoder:
👍

PR #1586, opened 2019-07-23
Focus for Radio, Checkbox and Switch

yxngl:
## Proposed Changes

This is a follow-up of #1341 

1. The focusable element is moved to input for Radio like what was done for Checkbox and Switch. I said browser behavior slightly differs, but it is only for ungrouped radio buttons (without name). In practice, it should not be a breaking change. Also name attribute is added for the first example in the doc.

2. There is a [browser bug](https://allyjs.io/tutorials/mutating-active-element.html#running-tests) in MacOS FireFox and Safari that checkbox and radio input do not receive focus when clicked. So manually calling focus method is added. I think it is necessary in radio group so that arrow keys can be used to navigate after clicking one radio button. For Checkbox and Switch, it is mostly for the visual box shadow.
Although button is also affected, it is untouched because the desired behavior could be more complicated, e.g. focus is recommended to move into modal if one is activated upon clicking.

### DIFF
--- a/docs/pages/components/radio/examples/ExSimple.vue
+++ b/docs/pages/components/radio/examples/ExSimple.vue
@@ -2,18 +2,22 @@
     <section>
         <div class="block">
             <b-radio v-model="radio"
+                name="name"
                 native-value="Flint">
                 Flint
             </b-radio>
             <b-radio v-model="radio"
+                name="name"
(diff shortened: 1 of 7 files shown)


jtommy:
At the moment I'm not able to test on Mac but I think it's ok, thanks!

PR #1590, opened 2019-07-24
Add aria label to close button in b-tag

yxngl:
## Proposed Changes
Add `aria-label` to the close button in b-tag for better accessibility.

### DIFF
--- a/src/components/tag/Tag.vue
+++ b/src/components/tag/Tag.vue
@@ -10,6 +10,7 @@
         <a
             class="tag is-delete"
             role="button"
+            aria-label="close"
             :tabindex="tabstop ? 0 : false"
             :disabled="disabled"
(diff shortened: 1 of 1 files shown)


service-paradis:
I think this should be a prop (`ariaCloseLabel`) like in other components. Would be better for localization.

yxngl:
Good point

jtommy:
You are free to open a new PR ;)

yxngl:
Just realized message has the same prop, but it does not have a default. Should I leave it empty as well?

jtommy:
In message component there isn't a default value

gntlman-bstrd:
is there some progress regarding that issue? 
Accessibility is still a problem for tags with a close button. :/

jtommy:
@gntlman-bstrd There is already a prop to set it in the last version

PR #1595, opened 2019-07-25
 Add ariaCloseLabel prop for Tag

yxngl:
## Proposed Changes

- Add `aria-close-label` prop for `tag` component to let users customize the label
- To be consistent with `message`, no default is set
- Examples updated


jtommy:
Thanks

PR #1598, opened 2019-07-25
clone value prop

### LINKED ISSUE #1596: v-model for b-tag-input is modified inside component
### Overview of the problem
For TagInput, v-model is passed by reference and has sides effets.

**Buefy** version: [0.7.7]
**Vuejs** version: [2.6.10]
**OS/Browser**: Chrome 75/Ubuntu 18.04

### Description
For the TagInput component, v-model is always an Array. When the v-model is passed to the TagInput component, the `value` prop is copied into the `tags` data in the component. Though the prop is not [cloned by referenced](https://github.com/buefy/buefy/blob/dev/src/components/taginput/Taginput.vue#L148) since an Array is an object in javascript. This lead to child component modifying the parent data.

This is more of a problem when using v-model from vuex. The passed vuex property is modified inside the TagInput component even if we use :value="myTags". We get the warning from vuex that we should not modify the state outside of the mutation handlers.

### Steps to reproduce

1. Create a vue.js Buefy project
2. In a component, use the TagInput component and use the :value @input notation instead of the v-model and omit the @input (<b-tag-input :value="myTags" />)
3. Go in your browser and use the TagInput to add a few tags
4. In Vue DevTools, see that myTags have been modified, but should not have been modified

### Expected behavior

Using only :value, I would expect my data variable not to have changed. The variable should only change when using @input.

`myTags` should not be modified inside TagInput

### Actual behavior

`myTags` were modified inside the child component TagInput.

lunfel:
Fixes #1596

## Proposed Changes
Clones the array passed in value prop for TagInput so it does not have sides effets anymore

-
-
-

### DIFF
--- a/src/components/taginput/Taginput.vue
+++ b/src/components/taginput/Taginput.vue
@@ -145,7 +145,7 @@ export default {
     },
     data() {
         return {
-            tags: this.value || [],
+            tags: this.value.slice(0) || [],
             newTag: '',
             _elementRef: 'input',
             _isTaginput: true
--- a/src/components/taginput/Taginput.vue
+++ b/src/components/taginput/Taginput.vue
@@ -145,7 +145,7 @@ export default {
     },
     data() {
         return {
-            tags: Array.isArray(this.value) ? Array.from(this.value) : (this.value || []),
+            tags: Array.isArray(this.value) ? this.value.slice(0) : (this.value || []),
             newTag: '',
             _elementRef: 'input',
             _isTaginput: true
--- a/src/components/taginput/Taginput.vue
+++ b/src/components/taginput/Taginput.vue
@@ -145,7 +145,7 @@ export default {
     },
     data() {
         return {
-            tags: this.value.slice(0) || [],
+            tags: Array.isArray(this.value) ? Array.from(this.value) : (this.value || []),
             newTag: '',
             _elementRef: 'input',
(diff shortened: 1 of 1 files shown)


jtommy:
Slice works well if it's a string array, btw you in case null or undefined it doesn't work

lunfel:
Ok, I'll check out how to make it IE11 compatible

lunfel:
Array.isArray seems to be compatible with IE9, and idk why I changed to Array.from. I put back .slice(0)

jtommy:
Ok!

ManUtopiK:
This issue is not resolved! With vuex, it's throw this error:
`[vuex] do not mutate vuex store state outside mutation handlers.`

`.slice(0)` doesn't drop Observer. You should use `JSON.parse(JSON.stringify(this.value))` instead...

lunfel:
@ManUtopiK I'm not sure which Observer you are talking about. Do you mean Vue.js watchers?

Because AFAIK, your solution (parse/stringify) would be a valid solution for a deep clone. But as the tags are strings and not object, only a shallow copy is needed here. I have found another place where it is passed by reference, I have a good feeling that this is causing the problem. I will open a PR for this.

https://github.com/buefy/buefy/blob/dev/src/components/taginput/Taginput.vue#L252

PR #1603, opened 2019-07-26
add hasCounter to Taginput

dejangeci:
Added hasCounter prop to Taginput.
Counter wrapper was changed from &lt;p&gt; to &lt;small&gt; to match Input component.


jtommy:
@dejangeci right! Thanks :)

PR #1604, opened 2019-07-26
Fixes #1576

### LINKED ISSUE #1576: label-position prop is incorrect using Number Input Component.
### Overview of the problem

**Buefy** version: 0.7.10
**Vuejs** version: 2.6.10

### Description

`label-position` prop is not displayed correctly using Number Input Component.

### Steps to reproduce

1. Create a `b-field` component having `label-position="on-border"` or `label-position="inside"`
2. Create a `b-numberinput` component inside
3. Check the label position

Here is a codepen: https://codepen.io/anon/pen/OKJqrG?&editable=true&editors=101
captured (1)
captured

### Expected behavior

The label should be displayed above the input only and not over the controls.

### Actual behavior

The label is starting above the controls

service-paradis:
Fixes #1576

## Proposed Changes

- Manage label position for number input component
- Also works with sizes
captured

### DIFF
--- a/docs/pages/components/numberinput/examples/ExCustomize.vue
+++ b/docs/pages/components/numberinput/examples/ExCustomize.vue
@@ -16,5 +16,18 @@
                 controls-rounded>
             </b-numberinput>
         </b-field>
+
+        <b-field label="Label inside" label-position="inside">
+            <b-numberinput></b-numberinput>
+        </b-field>
+        <b-field label="Label on border" label-position="on-border">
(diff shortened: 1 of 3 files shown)


jtommy:
To be honest I don't like add refs of other componnets into field but probably it's the only way to fix it.
I should think about it but before please remove the new examples, thanks ;)

service-paradis:
I understand, I also dont really like to add other components refs. I just could not find another way right now. Maybe I'm missing something....
Yes, I can remove the examples

PR #1605, opened 2019-07-29
Enable code splitting and tree shaking with multiple entry points

TrySound:
The only problem here is component modules naming (index.js). Actual
names would solve the problem.


jtommy:
Thanks a lot !!! But I need to fix entry names

PR #1606, opened 2019-07-29
Fixes #1601

### LINKED ISSUE #1601: Removed needless <span> from BTableColumn
Is there any reason why the [unnamed slot](https://github.com/buefy/buefy/blob/201692497c290445bb993d6630eb0ea5de81e135/src/components/table/TableColumn.vue#L6) in BTableColumn is wrapped inside a span? It seems to serve no specific purpose from what I can tell.

I have a pretty big table (~16 columns, ~50 rows) that I need to display and removing the span would shave ~800 DOM elements off of the page (Google/Lighthouse [suggests](https://developers.google.com/web/tools/lighthouse/audits/dom-size) having DOM tree with 1500 nodes or fewer for optimal performance).

service-paradis:
Fixes #1601 

## Proposed Changes

- Remove unneeded span around default column slot (potential minor breaking change)

### DIFF
--- a/src/components/table/TableColumn.vue
+++ b/src/components/table/TableColumn.vue
@@ -3,7 +3,7 @@
         v-if="visible"
         :class="rootClasses"
         :data-label="label">
-        <span><slot/></span>
+        <slot/>
     </td>
 </template>
 


jtommy:
Thanks! I have to fix some examples too

PR #1615, opened 2019-08-01
Icon props (next / previous)

service-paradis:
## Proposed Changes#

- Add default config for next/previous icon (used in Steps, Datepicker and Pagination)
- [Steps] Use default config for next/previous button icons. Also add a prop into the component
- [Datepicker] Use default config for next/previous button icons. Also add a prop into the component
- [Pagination] Use default config for next/previous button icons. Also add a prop into the component

### DIFF
--- a/src/components/pagination/Pagination.vue
+++ b/src/components/pagination/Pagination.vue
@@ -8,7 +8,7 @@
             @click.prevent="prev"
             :aria-label="ariaPreviousLabel">
             <b-icon
-                :icon="newIconPrev"
+                :icon="iconPrev"
                 :pack="iconPack"
                 both
                 aria-hidden="true"/>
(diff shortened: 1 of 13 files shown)


src/components/pagination/Pagination.vue
```diff
@@ -134,6 +137,13 @@ export default {
             return parseInt(this.rangeAfter)
         },
 
+        newIconPrev() {
+            return this.iconPrev || config.defaultIconPrev
```
jtommy: Why do we need this ? I think you might add  config as default value of the prop (so in other components)

service-paradis: Sure!
It was done this way in some other components.

Fixed!

jtommy:
It's ok but I don't want to add an example for each components, for example you might remove datepicker example

service-paradis:
Removed!
https://github.com/buefy/buefy/pull/1615/commits/a7cf76cde0a21423df10ba81fde1140b04c43d08

jtommy:
Please check conflicts

service-paradis:
Sorry for the conflict. CircleCI take a long time to run before I can see if there are any conflicts. Should be resolved now, but waiting to see CircleCI results.

service-paradis:
@jtommy Fixed!

PR #1621, opened 2019-08-01
Fix single file upload error

### LINKED ISSUE #1616: b-upload single file upload returns error
### Overview of the problem

**Buefy** version: 0.8.0
**Vuejs** version: 2.5.22
**Browser/OS**: Windows 10, Tested on Firefox and Chrome

### Description

Using <b-upload> without the "multiple" property returns an error "this.newValue.some is not a function"
Example at [Codesandbox](https://codesandbox.io/embed/vue-template-08171)
In console, you can see the error.

With the multiple property, this does not occur.

### Steps to reproduce

1. Select a file or drag and drop (with drag-drop property) into the Upload component.

### Expected behavior

Select a (single) file

### Actual behavior

Returns an error

yxngl:
Fixes #1616

### DIFF
--- a/src/components/upload/Upload.vue
+++ b/src/components/upload/Upload.vue
@@ -76,7 +76,8 @@ export default {
 
             if (!this.newValue || (Array.isArray(this.newValue) && this.newValue.length === 0) ||
             !inputFiles[0] ||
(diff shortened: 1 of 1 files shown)


jtommy:
Thanks

PR #1628, opened 2019-08-02
Fixes #1627

### LINKED ISSUE #1627: b-dropdown Keep open when clicking outside
### Description

I am trying to find some way to keep the dropdown menu open after a modal launches, but as soon as a user clicks on the modal or closes it, the dropdown will close as well. I would like some way to prevent the closing of the dropdown when clicking outside, and only close if the user clicks back on the trigger.

I have tried to prevent the `this.$refs.dropdown.toggle()` method with no success.

It would be a nice little feature to expand the flexibility of dropdowns.

Thanks!

service-paradis:
Fixes #1627

## Proposed changes

- New prop to determine what action can close the dropdown
- Default to escape key and clicking outside

### DIFF
--- a/src/components/dropdown/Dropdown.vue
+++ b/src/components/dropdown/Dropdown.vue
@@ -37,6 +37,8 @@
 <script>
 import config from '../../utils/config'
 
+const DEFAULT_CLOSE_OPTIONS = ['escape', 'outside']
+
 export default {
     name: 'BDropdown',
     props: {
@@ -75,6 +77,10 @@ export default {
         closeOnClick: {
             type: Boolean,
             default: true
+        },
+        canClose: {
+            type: [Array, Boolean],
+            default: true
         }
     },
     data() {
@@ -97,6 +103,13 @@ export default {
         },
         isMobileModal() {
             return this.mobileModal && !this.inline && !this.hoverable
+        },
+        cancelOptions() {
+            return typeof this.canClose === 'boolean'
+                ? this.canClose
+                    ? DEFAULT_CLOSE_OPTIONS
+                    : []
+                : this.canClose
         },
         ariaRoleMenu() {
             return this.ariaRole === 'menu' || this.ariaRole === 'list' ? this.ariaRole : null
@@ -188,9 +201,21 @@ export default {
         * Close dropdown if clicked outside.
         */
         clickedOutside(event) {
+            if (this.cancelOptions.indexOf('outside') < 0) return
             if (this.inline) return
 
             if (!this.isInWhiteList(event.target)) this.isActive = false
+        },
+
+        /**
+         * Keypress event that is bound to the document
+         */
+        keyPress(event) {
+            // Esc key
+            if (this.isActive && event.keyCode === 27) {
+                if (this.cancelOptions.indexOf('escape') < 0) return
+                this.isActive = false
+            }
         },
 
         /**
@@ -216,11 +241,13 @@ export default {
     created() {
         if (typeof window !== 'undefined') {
             document.addEventListener('click', this.clickedOutside)
+            document.addEventListener('keyup', this.keyPress)
         }
     },
     beforeDestroy() {
         if (typeof window !== 'undefined') {
             document.removeEventListener('click', this.clickedOutside)
+            document.removeEventListener('keyup', this.keyPress)
         }
     }
 }
--- a/docs/pages/components/dropdown/api/dropdown.js
+++ b/docs/pages/components/dropdown/api/dropdown.js
@@ -64,6 +64,13 @@ export default [
                 type: 'Boolean',
                 values: '—',
                 default: '<code>false</code>'
+            },
+            {
+                name: '<code>can-close</code>',
+                description: 'Can close dropdown by pressing escape or by clicking outside',
+                type: 'Boolean, Array',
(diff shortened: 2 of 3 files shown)


src/components/dropdown/Dropdown.vue
```diff
@@ -188,11 +201,23 @@ export default {
...
             if (!this.isInWhiteList(event.target)) this.isActive = false
         },
 
+        /**
+         * Keypress event that is bound to the document
+         */
+        keyPress(event) {
+            // Esc key
+            if (this.isActive && event.keyCode === 27) {
+                if (this.cancelOptions.indexOf('escape') < 0) return
```
afwn90cj93201nixr2e1re: Maybe we should move it to function start?

afwn90cj93201nixr2e1re:
Also would be nice if @jtommy gonna add close/escape/clickoutside to autocomplete too, like in modal's.
https://github.com/afwn90cj93201nixr2e1re/buefy/commit/967aeb9dde502e026c83907f83260f83ced17843

https://github.com/afwn90cj93201nixr2e1re/buefy/commit/e97df02f5ac2a4f071357829c9f5f7fe9f641794#diff-92a4fea4a43a1a0c5cd3adf2ae44f845

src/components/dropdown/Dropdown.vue
service-paradis: Thanks for your comment. Not sure about the order.

Since `keyup` is bound to the document, every key pressed in the page will trigger this function. If we fill a form or something, I want to exit the function as soon as possible if it is not needed. In this case, if the dropdown is not active, it wont even need to check whick key were pressed.

What are your thoughts on this?

afwn90cj93201nixr2e1re: How do you think, regular user, push esc on page too often?) xD

service-paradis: I'm not saying users press esc too often. I'm just saying that I'm not sure about the benefit of switching the conditions. Surely wont be that huge of an optimization.

afwn90cj93201nixr2e1re: But it's optimization.

2 ways
closed dropdown
opened dropdown

1: pushing a - call func check active -> exit
    write some another text on page - check active -> exit

2: pushing a - call func -> exit
    pushing another text in some field for example ->exit

Idk, that's your choice, ok.

jtommy: For me it's ok now.
@afwn90cj93201nixr2e1re Maybe you're right but I don't think about huge  decrease of performance.
If you have time you might make a benchmark ;)

jtommy:
@afwn90cj93201nixr2e1re you're right but I think the autocomplete will be refactored in order to use dropdown component internally

afwn90cj93201nixr2e1re:
Also add prop to make dropdown to not close on item click.

```js
 <b-dropdown v-model="$i18n.locale" aria-role="list" hoverable>
                            <button class="button is-primary" type="button" slot="trigger">
                                <b-icon :custom-class="`flag-icon flag-icon-${language[$i18n.locale].v}`"></b-icon>
                                <span>{{language[$i18n.locale].title}}</span>
                                <b-icon icon="caret-down"></b-icon>
                            </button>
                            <div  v-for="(lang,index,i) in language" :key="index">
                                <b-dropdown-item :value="lang.v" aria-role="listitem">
                                    <div class="media">
                                        <b-icon class="media-left" size="is-medium" :custom-class="`flag-icon flag-icon-${lang.v}`"></b-icon>
                                        <span>{{lang.title}}</span>
                                    </div>
                                </b-dropdown-item>
                                <hr v-if="i+1 != Object.keys(language).length" class="dropdown-divider" aria-role="listitem">
                            </div>
                        </b-dropdown>
```

I don't know why it's work's, but, you can check it here, https://2.shelru.ru

But with same configuration it's doesn't work on https://codepen.io/anon/pen/mNBoMJ?editors=1010

Maybe i added something to my version, idk, rly.
https://github.com/afwn90cj93201nixr2e1re/buefy/tree/dev/src/components/dropdown2
or https://github.com/afwn90cj93201nixr2e1re/buefy/tree/dev/src/components/dropdown

@jtommy 
https://github.com/buefy/buefy/blob/dev/src/components/dropdown/Dropdown.vue#L162-L173
that's not right way to check is item in dropdown as you know.

Coz we have custom slot, so, there can be something like:
```HTML
<div class=a>
abcd
<div class=b>
asdasdad
<div class=button>
asdasdasdasda123
</div>
</div>
</div>
```
if we gonna click on `asd123` dropdown gonna close, so, maybe just delete this part of code? Or make a prop for not close on click's? Coz it's realy anoying.

service-paradis:
@afwn90cj93201nixr2e1re 
There is already a prop `closeOnClick` for Dropdown. I think the other issues with autocomplete should be in their own issues. This PR is only to add a prop to help user controle what action can close the Dropdown.

@jtommy 
You want me to do something before merging this one?

afwn90cj93201nixr2e1re:
https://buefy.org/documentation/dropdown
can you find closeOnClick in doc's? xD

service-paradis:
The docs probably need to be updated, but it exists:
https://github.com/buefy/buefy/blob/9a686589500f12d775b683bdcd834ba215abf983/src/components/dropdown/Dropdown.vue#L77

PR #1017, opened 2018-10-04
[RANGE-DATE] Add functionality to select range of dates to DatePicker.

afifsohaili:
Addresses #932.

* To maintain backward compatibility, the component still accepts Date,
  but checks are performed to differentiate if users passed an Array of
  Date.
* Turn off border-radius for is-within selected cells.
* Turn off right border-radiuses for is-first selected and left
  border-radiuses for is-last selected cells.
* Add specs for DatePicker's dateformatter.
* Add specs for classObject in DatepickerTableRow.

p.s. I'm not sure what's up with the `yarn.lock`. It could be due to the different `yarn` version. I'm using `v1.10.1`

datepicker

**UPDATED**

buefy

PR #1637, opened 2019-08-07
Fix bug with DatePicker range not initializing correctly with empty v…

afifsohaili:
Fix bug in this [comment](https://github.com/buefy/buefy/pull/1017/#issuecomment-518170959)


jtommy:
Thanks

PR #1643, opened 2019-08-08
Component Proposal: Rate

apolokaklab:
### Proposed Changes
- Add a new compenent for Rate
- Half star is supported in Read only.
Buefy Rate

**Note**: this default use FA for better example.

@jtommy i waiting you feedback :)


src/components/rate/Rate.vue
```diff
@@ -0,0 +1,128 @@
...
+            type: Number,
+            default: 0
+        },
+        max: {
+            type: Number,
+            default: 5
+        },
+        icon: {
+            type: String,
+            default: 'fas fa-star'
```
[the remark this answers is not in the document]
apolokaklab: OK, I'll fix it :)

src/scss/components/_rate.scss
```diff
@@ -0,0 +1,90 @@
...
+    color: transparent;
+    overflow: hidden;
+  }
+  
+  .rate-text {
+    display: inline-block;
+    margin-left: 8px;
+    font-size: $size-small;
+    vertical-align: middle;
+  }
```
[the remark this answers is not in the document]
apolokaklab: I think it's better to separate, but if you don't think I will fix it

src/components/rate/Rate.vue
```diff
@@ -0,0 +1,128 @@
...
+    <div class="rate" :class="[size, { 'is-disabled': disabled }]">
+        <ul
+            class="rate-list"
+            @mouseleave="resetNewValue">
+            <li
+                v-for="(rate, index) in max"
+                class="rate-item"
+                :class="rateClass(rate)"
+                :key="index">
+                <i
```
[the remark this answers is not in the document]
apolokaklab: i have problem in style with other font icon, for now i working for fixing.

src/components/rate/Rate.vue
```diff
@@ -0,0 +1,128 @@
...
+            default: 5
+        },
+        icon: {
+            type: String,
+            default: 'fas fa-star'
+        },
+        size: String,
+        disabled: Boolean,
+        showScore: Boolean,
+        showText: Boolean,
```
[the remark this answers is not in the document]
apolokaklab: About ``showText`` some case ecommerce: for user input rating replace value with ``showText`` from template ``texts``. and ``customText`` for added maybe like text **Points** or text **total user ratings**.

apolokaklab:
Update!
With icon component
Rate-Simple
Rate-Full

src/scss/components/_rate.scss
```diff
@@ -0,0 +1,73 @@
...
+    }
+  }
+  .icon {
+    color: $rate-color;
+    pointer-events: none;
+    width: inherit;
+    // fix MDI when half star in all size
+    line-height: 1;
+  }
+  .is-half > i {
```
jtommy: This part is quite tricky. What happen when the icon pack isn't FA o MDI ? Because in the next release (in the current dev code) there is a feature that allows to add new icon packs

apolokaklab: Actually not only for MDI, but other icons can. Because I can't try all font icons so this style is just to reset the line height.

If there is a problem about this later, then I will fix it again soon.

jtommy: Yes, but I mean `is-half` style and `icon` class override

apolokaklab: Yes about that, I think it is the best solution at this time to support decimal values, please add this PR, so that in the latest version I can make it better with the latest features on the component icons. and support for all font icons listed in bulma.

sorry if my english is not good.

src/components/rate/Rate.vue
```diff
@@ -0,0 +1,135 @@
...
+        size: String,
+        spaced: Boolean,
+        disabled: Boolean,
+        showScore: Boolean,
+        showText: Boolean,
+        customText: String,
+        texts: {
+            type: Array,
+            default() {
+                return ['Very Bad', 'Bad', 'Good', 'Very Good', 'Awesome']
```
service-paradis: I'm not sure we should include default texts. It would be too much opiniated I think.

src/scss/components/_rate.scss
```diff
@@ -0,0 +1,73 @@
+$rate-color: $grey-lighter !default;
+$rate-color-active: darken($warning, 5%) !default;
+$rate-space: 5px !default;
```
service-paradis: I think we should try to use `em` or `rem` instead of px (there are other places too)

src/scss/components/_rate.scss
service-paradis: Buefy should accept other icon packs, but I think it should officially support only FA and MDI.
People who uses other packs should be responsible for style customization if needed.
So I think, it is probably good this way.

src/scss/components/_rate.scss
```diff
@@ -0,0 +1,73 @@
...
+    // This follow width in rate.vue
+    // Because b-icon not support custom style
+    position: absolute;
+    left: 0;
+    top: 0;
+    width: inherit;
+    overflow: hidden;
+    // fix FA not same height + position from size default (fa-lg)
+    &.fa-lg {
+      top: 1.25px;
```
service-paradis: Not sure why customizing top is necessary... how is it possible that the same icon have different height?

service-paradis: This is my proposal that removed the need for this fix. You will get the same behavior without having a font specific fix.

Add this scss:
```
.rate .rate-item.set-half .is-half {
    position: absolute;
    left: 0;
    top: 0;
    overflow: hidden;
}
```

Remove `.rate .is-half > i.fa-lg` entirely

Change `.rate .is-half > i` to :
```
.rate .is-half > i {
    position: absolute;
    left: 0;
}
```

src/components/rate/Rate.vue
service-paradis: What do you think about having default texts @jtommy ?

src/scss/components/_rate.scss
service-paradis: Maybe something like `0.25rem` for the `rate-space`? It would be the same item spacing as Bulma pagination (`$pagination-item-margin`).

apolokaklab:
@service-paradis Thanks for solution :tada:

src/components/rate/Rate.vue
jtommy: @apolokaklab  I agree with @service-paradis , please remove it

service-paradis: Could add an example in the doc on how to use it :)

jtommy:
@apolokaklab please check @service-paradis review

src/scss/components/_rate.scss
```diff
@@ -0,0 +1,64 @@
...
+    position: absolute;
+    left: 0;
+  }
+  .rate-text {
+    align-items: center;
+    font-size: calc(#{$size-normal} - 2px);
+    margin-left: 6px;
+    // Alternative for size text from b-icon
+    &.is-small {
+      font-size: calc(#{$size-small} - 2px);
```
service-paradis: I would try to use something along `font-size: calc(#{$size-small} / 1.5);` instead of using px for `.rate-text`

service-paradis:
@apolokaklab 
Great job on this component! 

I added some more comments, but it looks great! You will probably need to do a rebase to prevent conflict with single component fix. It would be really great if you could add some more unit tests :)

apolokaklab:
@jtommy done, @service-paradis of course.

jtommy:
@apolokaklab Great!

PR #1664, opened 2019-08-14
Update Openbook to Okuna

lifenautjoe:
We're one of the featured sites.

We've recently changed our name from Openbook to Okuna.

# Proposed Changes

Change feature page from Openbook to Okuna


jtommy:
Thanks!!

Issue #1355, opened 2019-04-27
Password reveal icon state incorrect


Henry-W:
### Overview of the problem
Looking at other websites, it appears the icon (eye) is the wrong way round for the state of the input.

For example:

- Google shows a crossed out eye icon when the password is masked with the standard dots.
- Buefy shows a normal eye icon when the password is masked with the standard dots.

I have identified that the following line will need changing:
https://github.com/buefy/buefy/blob/fec335898e7b1e5c03847a9fae801f45c5548c3e/src/components/input/Input.vue#L164

to

`return this.isPasswordVisible ? 'eye' : 'eye-off'`

### Expected behavior

Icon to display as crossed out eye (eye-off) when password is masked.

### Actual behavior

Icon displays as normal eye (eye) when password is masked.

This is important to be consistent with other websites and to provide a good user experience.

jtommy:
@Henry-W I think there isn't a correct or incorrect state.
For example you take a look at Adobe login, it follows same pattern of buefy

shelru:
reveal ico z-index value must be lower than LoadingComponent z-index, i described it on https://github.com/buefy/buefy/issues/1323#issuecomment-483121923

Henry-W:
> @Henry-W I think there isn't a correct or incorrect state.
> For example you take a look at Adobe login, it follows same pattern of buefy

Fair enough, but I think Google's version makes more sense, considering a strike out shows it's not normal and hence masked.

service-paradis:
We could maybe add a prop `password-reveal-icons` with `['eye' : 'eye-off']` as default?
This way, someone who want to reverse the icon could change this prop without adding a breaking change to Buefy. What do you think?

PR #1681, opened 2019-08-20
Proposal: Add icons customization possibility

### LINKED ISSUE #179: Feature Request: Custom icon fonts
**Buefy** version: 0.4.5
**Vuejs** version: 2.4.1
### Description

Now, there are only two icon fonts support, FontAwesome and MDI, ¿there are way to user use their custom fonts? My pull request was rejected, I can make other pull request, but I need know how it should be.

### Steps to reproduce
Components that uses b-icon inside, only get icon name, and it is concatenated with fa or showed with mdi format, there are a lot of icon fonts and now there are no way to use a custom icon with this.

This: `<b-input type="text" icon="edit"></b-input>`

Include a icon with `<i class="fa fa-edit"></i>` or `<i class="mdi">edit</i>`

### Expected behavior

This: `<b-input type="text" icon="edit"></b-input>`

Include a icon with `<i class="any_custom_prefix-edit"></i>`

### Actual behavior

Now there are two types of icons, user icons(for example de input icons) and internal icons(icons that are needed by buefy). @rafaelpimpa suggest use a svg to internal fonts and `b-icon` to other icons, we have to consider that if we only use svg to internal icons and cant' be customized it could break design(Inconsistency between user icons and internal icons), besides increment css file with a svg when I dont want use buefy default icons. 

### Solution 1:
Follow using packages:
- Create configuration to add new packages [{ packname, prefix, style(class_style[fa] vs content_style[mdi]) }].
- Convert internal icons to use svg.
- Add optional parameter to customize internal icons.

Example:

```Vue.use(Buefy, { moreFontPackages: {package1: {} ...} });```

### Solution 2:
Don't define icons packages.
- Create a configuration to use a package that not add a prefix, using `<b-icon icon="custom_prefix-edit"></b-icon>` or `<b-icon icon="fa fa-edit"></b-icon>`.
- To back compatibility, we have to leave defaultIconPack, but deprecate it.
- Convert internal icons to use svg, and leave user override icons overrides css classes.
* This way doesn't support mdi format (although there are a mdi style based in class)

Accept suggests..

### LINKED ISSUE #536: Decouple icons fonts from buefy
### Overview of the problem

**Buefy** version: 0.6.2
**Vuejs** version: 2.5.13
**OS/Browser**: Ubuntu/Firefox

### Description

It would be easier to the users to switch to any new font icons from different font families or use svg fonts. 
For it to be realized, Icons in other components can be slots.

### PR DESCRIPTION
**Pull request by service-paradis**
Fixes #179
Fixes #536

## Proposed Changes

- Add a new configuration property named `customIconPacks`
- Create icon pack configuration for the 2 default icon pack
- Changes Icon component to use the icon pack configuration to display icons
- Add an example on how to use the configuration
- This is not even a breaking change!

I'm open to discussion on this PR!

### DIFF
--- a/docs/pages/components/icon/Icon.vue
+++ b/docs/pages/components/icon/Icon.vue
@@ -32,7 +32,7 @@
             <p>You can also use object syntax for <code>type</code> props just like Vuejs <code>class</code>.</p>
         </Example>
 
-        <Example :component="ExCustom" :code="ExCustomCode | pre" title="Custom font" vertical>
+        <Example :component="ExCustom" :code="ExCustomCode" title="Custom font" vertical>
             <div class="tags has-addons">
                 <span class="tag is-success">New!</span>
                 <span class="tag is-info">0.8.3</span>
@@ -58,6 +58,7 @@
     import ExObjectSyntaxCode from '!!raw-loader!./examples/ExObjectSyntax'
 
     import ExCustom from './examples/ExCustom'
+    import ExCustomCode from '!!raw-loader!./examples/ExCustom'
 
     export default {
         data() {
@@ -70,6 +71,7 @@
                 ExFaCode,
                 ExObjectSyntaxCode,
                 ExCustom,
+                ExCustomCode,
                 usage: `
                 import { library } from '@fortawesome/fontawesome-svg-core';
                 // internal icons
@@ -89,38 +91,7 @@
                 Vue.use(Buefy, {
                   defaultIconComponent: 'vue-fontawesome',
                   defaultIconPack: 'fas',
-                });`,
-                ExCustomCode: `
-                Vue.use(Buefy, {
-                    defaultIconPack: 'cap-icon',
-                    customIconPacks: {
-                        'cap-icon': {
-                            sizes: {
-                                default: 'is-size-5',
-                                'is-small': '',
-                                'is-medium': 'is-size-3',
-                                'is-large': 'is-size-1'
-                            },
-                            iconPrefix: 'ci-',
-                            internalIcons: {
-                                check: 'check',
-                                information: 'info',
-                                'check-circle': 'info-circle',
-                                alert: 'info',
-                                'alert-circle': 'info-circle',
-                                'arrow-up': 'arrow-up',
-                                'chevron-right': 'arrow-right',
-                                'chevron-left': 'arrow-left',
-                                'chevron-down': 'arrow-down',
-                                eye: 'eye',
-                                'eye-off': 'unlock',
-                                'menu-down': 'arrow-down-circled',
-                                'menu-up': 'arrow-up-circled'
-                            }
-                        }
-                    }
-                });
-                `
+                });`
             }
         }
     }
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by service-paradis** [comment]
I dont understand why unit tests take so long on CircleCI, but I can assure you every tests are passing on this PR.

**Comment by jtommy** [comment]
@service-paradis The idea is great but I need time to review it, thanks!

File: docs/pages/components/icon/examples/ExCustom.vue [hunk 1]
```diff
@@ -0,0 +1,135 @@
...
+                    'eye-off': 'unlock',
+                    'menu-down': 'arrow-down-circled',
+                    'menu-up': 'arrow-up-circled'
+                }
+            }
+        }
+    }
+    export default {
+        created() {
+            setOptions(Object.assign(config, customIconConfig))
```
jtommy: You might replace it using `this.$buefy.config.setOptions(Object.assign(config, customIconConfig))`, so it's a good opportunity to show programmatic config 😄

docs/pages/components/icon/Icon.vue
```diff
@@ -77,7 +89,38 @@
                 Vue.use(Buefy, {
                   defaultIconComponent: 'vue-fontawesome',
                   defaultIconPack: 'fas',
-                });`
+                });`,
+                ExCustomCode: `
```
jtommy: You can remove it and show the example code adding `import ExCustomCode from '!!raw-loader!./examples/ExCustom'` (without | pre)

docs/pages/components/icon/examples/ExCustom.vue
```diff
@@ -0,0 +1,135 @@
...
+            'cap-icon': {
+                sizes: {
+                    default: 'is-size-5',
+                    'is-small': '',
+                    'is-medium': 'is-size-3',
+                    'is-large': 'is-size-1'
+                },
+                iconPrefix: 'ci-',
+                internalIcons: {
+                    check: 'check',
```
jtommy: Only for consistency you might use always string as key, for example `'check': 'check'`

service-paradis:
Thanks for the comments! I just updated the code with your suggestions :)

Should we add a note somewhere that this could be used to customize the internal icons of the default packs?

And since we can do it, `defaultIconPrev` and `defaultIconNext` would be unnecessary and could be removed on the next breaking change.

jtommy:
Good job!! Thanks 🎉

service-paradis:
@jtommy 
To be even more flexible, we could use deep merging. This way, we could be able to customize the default icon packs by doing something like this:
```
customIconPacks: {
            fas: {
                sizes: {
                    'default': faIconPrefix + 'sm',
                    'is-small': faIconPrefix + 'xs',
                    'is-medium': faIconPrefix + '2x',
                    'is-large': faIconPrefix + '3x'
                }
            },
     }
```

We could be able to close some more issues like #1434 and #1355 for example.

jtommy:
It's a good idea, is not already possible after this PR ?

service-paradis:
I tough so, but unfortunately not :(. Object.assign only do a shallow merge instead of a deep merge:
https://stackoverflow.com/questions/27936772/how-to-deep-merge-instead-of-shallow-merge

service-paradis:
I kind of like the option from this comment:
https://stackoverflow.com/a/53092443/2774496

It could be a great replacement for Object.assign. It would be compatible with IE and would do a deep merge if needed.

PR #1683, opened 2019-08-21
Adding way to programatically set a field invalid for remote field validations.

GerryWilko:
# Proposed Changes

I have added a way to set the validity on an input manually. This is usefull for when further validation occurs on an API and validation messages need to be applied back to fields.

Added to FormElementMixin so should be available on all Buefy form inputs for people to use.

Cheers!


jtommy:
Thanks!

PR #1687, opened 2019-08-22
Disable timepicker hour if no enabled minutes available

### LINKED ISSUE #1686: Timepicker with min-time and increment-minutes allows hours with no available minutes to be selected
### Overview of the problem

Buefy version: 0.8.2
Vuejs version: 2.6.10
OS/Browser Chrome version 76.0.3809.100

### Description

If a timepicker with increment-minutes and a min-time is set, can lead to being able to select an hour even through there are no available minutes to be selected for that hour

### Steps to reproduce

CodePen [https://codepen.io/philnee/pen/qBWqGGJ](https://codepen.io/philnee/pen/qBWqGGJ)
1. Create a timepicker with increment-minutes set to 10 and min-time set to 10:55
2. Select the hour 10 from the timepicker
3. All minutes values are disabled

### Expected behavior

I would have expected in this scenario that the hour 10 is disabled and the first hour available was 11 as the earliest valid time would be 11:00

### Actual behavior

The user can still select 10, however then cannot complete the time as no valid minutes options are available.

I will try and submit a PR later on to fix the issue

philnee:
Fixes #

Fixes #1686 

## Proposed Changes

When checking against minTime in isHourDisabled() in TimepickerMixin.js, add a further call to check if all of the minutes values are disabled and set disabled to true if so. 

The check for the disabled minutes was extracted from the isMinuteDisabled function and extracted to a seperate function isMinuteDisabledForHour, now used by both functions.

Test included in Timepicker.spec.js

### DIFF
--- a/src/components/timepicker/Timepicker.spec.js
+++ b/src/components/timepicker/Timepicker.spec.js
@@ -22,4 +22,15 @@ describe('BTimepicker', () => {
 
         expect(wrapper.vm.nativeStep).toBe('1')
     })
+
+    it('disables hour if no selectable minutes available', () => {
+        const wrapper = shallowMount(BTimepicker, {
+            propsData: {
+                'min-time': new Date(2019, 8, 22, 11, 55),
+                'increment-minutes': 10
+            }
+        })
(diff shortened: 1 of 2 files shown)


jtommy:
Thanks

PR #1690, opened 2019-08-22
Add a polyfill for the findIndex function

service-paradis:
## Proposed Changes

- Add a new helper function named findIndex
- Use the native function if available but revert to polyfill if not
- Steps pagination is now compatible with IE

### DIFF
--- a/src/components/steps/Steps.vue
+++ b/src/components/steps/Steps.vue
@@ -67,6 +67,7 @@
 import Icon from '../icon/Icon'
 import SlotComponent from '../../utils/SlotComponent'
 import config from '../../utils/config'
+import { findIndex } from '../../utils/helpers'
 
 export default {
     name: 'BSteps',
@@ -127,7 +128,7 @@ export default {
          * Check the first visible step index.
          */
         firstVisibleStepIndex() {
-            return this.stepItems.findIndex((step, idx) => {
+            return findIndex(this.stepItems, (step, idx) => {
                 return step.visible
             })
         },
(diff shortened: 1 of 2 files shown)


jtommy:
I didn't check findIndex but I prefer fix it using map + indexOf approach

service-paradis:
No problem!
I made it this way since most of the browsers support it natively ([findIndex Compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex#Browser_compatibility)). I can change it using map/indexof instead.

jtommy:
Yes I know but I don't want add other code at the moment

service-paradis:
Yeah, I understand. Will fix the PR without the jelper function and having map/indexof to replace the actual findIndex.

service-paradis:
@jtommy Fixed!

afwn90cj93201nixr2e1re:
@jtommy do you know anything 'bout js less memory using and performance or no?

service-paradis:
@afwn90cj93201nixr2e1re 
What alternatives would you use to keep browsers compatibility?
The usage in steps would probably dont be affected that much by map performance. I dont think people use this cmponent with a lot of steps. Probably always below 10.

jtommy:
@afwn90cj93201nixr2e1re I know it but I agree with @service-paradis

afwn90cj93201nixr2e1re:
Just simple itteration i=0;i<array.length etc. Its faster 10 times and works on every browser.

Also polyfill that's suggested by service-paradise is better in this case.

Also we don't know how fast gc in ie10,11 etc.

PR #1692, opened 2019-08-23
WIP: First try as pagination button customization

### LINKED ISSUE #1430: Pagination href
### Overview of the problem

**Buefy** version: [0.7.4]
**Vuejs** version: [2.6.10]
**OS/Browser**: Chrome 74

### Description

Currently it is only possible to navigate with the pagination component via javascript. It would be awesome if you could add an option to custom render the href attribute of the a tags. This could get handy in server side rendered env like nuxt.

### Expected behavior

A fallback method for non javascript users.

### Actual behavior

JavaScript is fully required.

### LINKED ISSUE #1654: Pagination with vue-router
Hi, I dont know how to program this. I need pagination widget from buefy, but I want to have pagination in url for generating permalinks.

Its possible to do this? And without refreshing in page every url (pagination) change?

Thanks!

service-paradis:
Fixes #1654
Fixes #1430

This is a WIP. I just made a minimal PR to see if it makes sense. Hoping to have some feedback about it!

## Proposed Changes

- Add a new component to customize pagination buttons
- The tag can be customized (`button`, `a`, `router-link`, ...)
- Attributes can be inherited (`to` for `route-link` for example)
- Default slot to customize the content

pagination

### DIFF
--- a/docs/pages/components/pagination/Pagination.vue
+++ b/docs/pages/components/pagination/Pagination.vue
@@ -2,6 +2,11 @@
     <div>
         <Example :component="ExSimple" :code="ExSimpleCode"/>
 
+        <Example :component="ExSlot" :code="ExSlotCode" title="Custom pagination buttons">
+            <p>You can customize the buttons using <b><code>b-pagination-button</code></b> as scoped slot.</p>
+            <p>For example, you can customize the inner text or use <code>router-link</code> as tag.</p>
+        </Example>
+
         <ApiView :data="api"/>
     </div>
 </template>
@@ -12,12 +17,17 @@
     import ExSimple from './examples/ExSimple'
     import ExSimpleCode from '!!raw-loader!./examples/ExSimple'
 
+    import ExSlot from './examples/ExSlot'
+    import ExSlotCode from '!!raw-loader!./examples/ExSlot'
+
     export default {
         data() {
             return {
                 api,
                 ExSimple,
-                ExSimpleCode
+                ExSimpleCode,
+                ExSlot,
+                ExSlotCode
             }
         }
     }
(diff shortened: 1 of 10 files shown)


modularcoder:
@service-paradis I think the previous/next links should be customizable as well, not just the pagination buttons.

Also I would suggest improving the prev/next costomizability as well

- User might want to have a custom next/prev buttons
- User might not want to have prev/next buttos at all (now there's only two options simple and default)

What do you think?

service-paradis:
@modularcoder 
Thanks for the feedback!

Yes, I would use the same behavior and subcomponent for next/previous button with a proper slot. Just did not want to go too far before having some feedback.

To hide the next/previous button, maybe we could add a new prop for this?

modularcoder:
@service-paradis 

maybe for the showin/hiding the prev/next there could be props. But I guess there can be the following use cases

< Custom Prev Button Component | ...Custom Pagination Components... | Custom Next Button Component >

So I guess the prev/next might have own slots and not necessarily always use `b-pagination-button`

service-paradis:
@modularcoder 

I tough it was easier to keep the same style by using the same subcomponent. Something like this:
localhost_9000_documentation_pagination

It is easier to only change the content or the link behavior without having to recreate the entire button in a slot. I dont know which behavior is more natural or which way will fit the most with Buefy.

anna-sorts:
Thank you for you work, @service-paradis. It will solve the problem I'm having when trying to edit the child components inside b-pagination. 
Hope it gets approved soon.

jtommy:
Awesome!
Please resolve conflict (now there is no need to add named exports, only default)

PR #1694, opened 2019-08-23
Sorting on multiple columns

suneab:
Added the option to enable sort-multiple columns. This feature will indicate with the regular icon and an incrementing number of how your custom sorting method sorted.

### DIFF
--- a/docs/pages/components/table/examples/ExSortMultiple.vue
+++ b/docs/pages/components/table/examples/ExSortMultiple.vue
@@ -5,13 +5,13 @@
           <b-switch v-model="multipleColumnsSwitch">Sort multiple rows</b-switch>
       </div>
       <div class="control is-flex">
-          <span class="button" @click="resetPriority">reset</span>
+          <span class="button" @click="resetPriority" :disabled="sortingPrioirty.length === 0">Reset sorting</span>
       </div>
     </b-field>
     <b-table 
     :data="data"
-    :backend-sorting="true"
     @sort="sortPressed"
+    :backend-sorting="true"
     :sort-multiple="multipleColumnsSwitch"
     :sort-multiple-data="sortingPrioirty"
     >
--- a/docs/pages/components/table/examples/ExSortMultiple.vue
+++ b/docs/pages/components/table/examples/ExSortMultiple.vue
@@ -1,0 +1,73 @@
+<template>
+  <div>
+    <b-field grouped group-multiline>
+      <div class="control is-flex">
+          <b-switch v-model="multipleColumnsSwitch">Sort multiple rows</b-switch>
+      </div>
+      <div class="control is-flex">
+          <span class="button" @click="resetPriority">reset</span>
+      </div>
+    </b-field>
+    <b-table 
+    :data="data"
+    :backend-sorting="true"
+    @sort="sortPressed"
+    :sort-multiple="multipleColumnsSwitch"
+    :sort-multiple-data="sortingPrioirty"
+    >
+      <template slot-scope="props">
+        <b-table-column field="first_name" label="First name" sortable>
+            {{ props.row.first_name }}
+        </b-table-column>
+        <b-table-column field="last_name" label="Last name" sortable>
+            {{ props.row.last_name }}
+        </b-table-column>
+        <b-table-column field="team" label="Team" sortable>
+            {{ props.row.team }}
+        </b-table-column>
+      </template>
+    </b-table>
+  </div>
+</template>
+
+<script>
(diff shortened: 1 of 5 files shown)


jtommy:
It looks nice but it think it's better enable the additional column sort by a key + mouse click combination (for example `alt + click` or `ctrl + click` or other key customizable using a prop)

src/components/table/Table.vue
```diff
@@ -86,8 +86,23 @@
...
+                                        sortMultipleData.find(i =>
+                                    i.field === column.field)">
+                                    <b-icon
+                                        :icon="sortIcon"
+                                        :pack="iconPack"
+                                        both
+                                        :size="sortIconSize"
+                                        :class="{ 'is-desc': isDesc(column)}"
+                                    />
+                                    {{ sortMultipleData.findIndex(i =>
```
jtommy: Please avoid to add logic in the template, you might use a computed prop

src/components/table/Table.vue
```diff
@@ -569,6 +592,11 @@ export default {
         }
     },
     methods: {
+        isDesc(column) {
+            let foundSortData = this.sortMultipleData.length > 0
+                ? this.sortMultipleData.find((i) => i.field === column.field) : false
```
jtommy: Please don't use `find` method because it isn't IE 11 compatibile

src/components/table/Table.vue
suneab: The logic isDesc to find the current row in the data i pass. Not sure how i would fix that one, if thats what you mean. 
I changed line 101 to pass whatever you like, ex a,b,c instead of 1,2,3

suneab:
I hope if fixed the issues you mentioned about the code review.
Your comment about additional key to add additional sort columns made sense, but instead of editing the table to handle it internally, i emitted the event to the user as they would need it for backend sorting anyways. I updated the example as well.

src/components/table/Table.vue
```diff
@@ -86,8 +86,23 @@
...
+                                        sortMultipleData.find(i =>
+                                    i.field === column.field)">
+                                    <b-icon
+                                        :icon="sortIcon"
+                                        :pack="iconPack"
+                                        both
+                                        :size="sortIconSize"
+                                        :class="{ 'is-desc': isDesc(column)}"
+                                    />
+                                    {{ sortMultipleData.find(i =>
```
jtommy: You should remove `find` method and refactor it into a method

docs/pages/components/table/examples/ExSortMultiple.vue
```diff
@@ -0,0 +1,82 @@
...
+            sortingPrioirty: [],
+            onlyWhenKeyIsDown: false
+          }
+      },
+      methods: {
+        resetPriority(){
+          this.sortingPrioirty = []
+          this.loadAsyncData()
+        },
+        sortPressed(field, order, event) {
```
jtommy: I don't agree with you, the sort logic should to be in table. For example take a look at https://handsontable.com/docs/7.1.1/demo-multicolumn-sorting.html

suneab:
@jtommy i don't know if I'm misunderstanding something. Backend sorting is required for this to work. Is it because you want it to work without backend sorting?

jtommy:
Add client side sort could be nice to have, but the major issue is when we have to action the multiple sort and using only simple ok is not right

suneab:
Hi @jtommy.
I've been looking at our conversation again and I'm sorry, but i don't know what I'm supposed to change. 
As far as i understand, a default sort method is not required. 
The way i implemented the multi column sorting, backend sorting is required. 
Could you clarify, what I'm supposed to change?

jtommy:
We should provide a full feature if we want add it in the core. So it has to work not only using backend sort and add a way to enable it (key + click)

suneab:
Hi @jtommy 
Just writing to let you know I'll get the request changes soon, like over the weekend or something. 
Sorry about the delay.

ms-saltation:
This feature would be nice to have. Any chance that this feature is implemented in the near future ?
Should i search for another library, if i rely on this.

suneab:
> This feature would be nice to have. Any chance that this feature is implemented in the near future ?
> Should i search for another library, if i rely on this.

Hi
I hoped to get a chance to implement this in the near future as i stated previously. I'll give it another try soon

suneab:
will create a new pull request for this feature

suneab:
@ms-saltation 
i just opened a new pull request for this feature on #2155. Feel free to provide feedback over there

PR #1697, opened 2019-08-25
Fixed navbar adding incorrect class to body when navbar is fixed at top/bottom

Silvea12:
Fixes incorrect body class being added to the body when navbar is fixed to the top/bottom, which prevented the correct spacing to be added.

See https://bulma.io/documentation/components/navbar/#fixed-navbar


jtommy:
Thanks

PR #1702, opened 2019-08-25
Fix #1701 - Taginput cannot add Footer slot

### LINKED ISSUE #1701: Taginput cannot add a Footer slot (unless Header is present)
### Overview of the problem

**Buefy** version: [0.8.2]
**Vuejs** version: [2.6.10]
**OS/Browser**: osx/Chrome

### Description
On the TagInput, the Footer slot does not appear unless the Header slot is present. This is, I believe, due to a typo (details below).

### Steps to reproduce

See [Codepen](https://codepen.io/tsummerer/pen/JjPWyZd?editors=1010)

1. Create a TagInput and add a Footer slot like this:
```js
<b-taginput
    ...various attributes...
>
    <template slot="footer">
            This is the footer!
     </template>
</b-taginput>
```
### Expected behavior

There should be a Footer item at the end of the list of options

### Actual behavior

There is no footer. 

### Workaround:
If you add a **Header** slot then both items work as expected:

```js
<b-taginput
    ...various attributes...
>
    <template slot="header">
            This is the header!
     </template>
    <template slot="footer">
            This is the footer!
     </template>
</b-taginput>
```
Now both slots show up!

### Cause:
I believe that this code is wrong:

```js
// Taginput.vue
// line 185
headerSlotName() {
    return this.hasHeaderSlot ? 'header' : 'dontrender'
},

footerSlotName() {
    return this.hasHeaderSlot ? 'footer' : 'dontrender'   // <-- oops
},
```

I'll submit a PR shortly.

tsummerer:
Fixes #1701 
## Proposed Changes

Fixed typo in Taginput.vue. This will allow Taginput fields to have Footer slot in the list of options. Currently it won't work unless there is _also_ a header slot.

### DIFF
--- a/src/components/taginput/Taginput.vue
+++ b/src/components/taginput/Taginput.vue
@@ -187,7 +187,7 @@ export default {
         },
 
         footerSlotName() {
-            return this.hasHeaderSlot ? 'footer' : 'dontrender'
+            return this.hasFooterSlot ? 'footer' : 'dontrender'
         },
 
         hasDefaultSlot() {


jtommy:
Thanks

PR #1710, opened 2019-08-29
Fix bug with vee-validate 3.0

tiamo:
**ValidationProvider** doesn't work with **b-select**, this pull request fixes this bug.


DevGoslinga:
The author of vee-validate has fixed this issue recently. 
https://github.com/logaretm/vee-validate/issues/2293

PR #1721, opened 2019-08-31
Fix background color for Outlined Switch with type

service-paradis:
Fix background color for Outlined Switch with type

From a Discord comment


jtommy:
Thanks!

PR #1725, opened 2019-09-01
docs: steps has navigation

### LINKED ISSUE #1724: Steps prop "has-navigation" is type Boolean not a String
### Overview of the problem

**Buefy** version: [0.8.2]
**Vuejs** version: [2.5.13]

### Description
Docs do not reflect the actual type of the prop that needs to be passed.

### Expected behavior
The docs say `has-navigation` is a String

### Actual behavior
The prop type for `has-navigation` is Boolean

harps116:
Fixes #1724
## Proposed Changes

- change from String to Boolean on has navigation prop


jtommy:
Thanks!

PR #1736, opened 2019-09-04
Add 'is-right' option for NavbarDropdown

VladNeacsu:
## Proposed Changes
- Add 'is-right' class for `NavbarDropdown` menus in order for menus that are put in the `end` slot of the Navbar component to not be cut

### DIFF
--- a/src/components/navbar/NavbarDropdown.vue
+++ b/src/components/navbar/NavbarDropdown.vue
@@ -11,7 +11,7 @@
             <template v-if="label">{{ label }}</template>
             <slot v-else name="label" />
         </a>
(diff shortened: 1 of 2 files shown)


src/components/navbar/NavbarDropdown.vue
```diff
@@ -28,7 +28,11 @@ export default {
     props: {
         label: String,
         hoverable: Boolean,
-        active: Boolean
+        active: Boolean,
+        isRight: {
```
jtommy: I prefer only "right" as name

jtommy:
Thanks

PR #1754, opened 2019-09-09
Tabbed components mixin

service-paradis:
## Proposed Changes

- Add mixins for Steps and Tabs
- Add mixins for Steps and Tabs Items

Steps and Tabs had a lot in common. Those are now lightweight.

### Possible Breaking Change
Unfortunately, if people were calling methods directly, thiscould be a breaking change. We should probably discuss about it before merging it!

### DIFF
--- a/src/components/steps/StepItem.spec.js
+++ b/src/components/steps/StepItem.spec.js
@@ -6,7 +6,7 @@ const BSteps = {
     template: '<b-step-stub></b-step-stub>',
     data() {
         return {
-            stepItems: [],
+            childItems: [],
             _isSteps: true
         }
     },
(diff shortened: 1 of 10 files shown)


afwn90cj93201nixr2e1re:
> were calling methods directly

You can create global proxy inside component(for example with extend) for back. comp.

afwn90cj93201nixr2e1re:
@jtommy call me when u gonna make break release, im gonna try to remind what' we should add.

Tofandel:
@service-paradis If methods are not documented, they should never be supported for BC, it means they are only internal and people use them at their own risks, they do have the advantage of triggering an error if the method is removed and was used instead of failing silently

service-paradis:
Sure! I just though it was preferable to wait a breaking release (ex. 0.9.0) since it was not something urgent.

PR #1766, opened 2019-09-16
Added scrollTo for all form elements

karimvally1:
## Proposed Changes

Adding support for scrollTo element using the HTML scrollIntoView function.

### DIFF
--- a/src/utils/FormElementMixin.js
+++ b/src/utils/FormElementMixin.js
@@ -106,7 +106,7 @@ export default {
         },
 
         scrollTo(options) {
-            this.getElement().scrollIntoView(options);
+            this.getElement().scrollIntoView(options)
         },
 
         setInvalid() {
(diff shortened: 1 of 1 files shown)


jtommy:
Why do you need it ? I don't understand why do you think it's a good feature to add in the core code

karimvally1:
> Why do you need it ? I don't understand why do you think it's a good feature to add in the core code

It's a good feature because it exposes the underlying HTML5 function to scroll to element which is useful for invalid states from a user experience point of view. In my case, a long form which has an invalid input using the scroll to function would bring the invalid state to the users attention.

jtommy:
@karimvally1 can you provide me a code example about it ?

jtommy:
I think it isn't an added value adding it in the core code at the moment.  
Maybe in a future form validation feature.
Btw thanks!

PR #1769, opened 2019-09-16
add full-screen option to ./api/modal

airandopal:
It doesnt appear that I can propose a change to the ./api/modal file so I'm sending it in here. And it doesn't really fit the contributing guidelines so I hope this is ok.

I believe this is the file [here](https://github.com/buefy/buefy/blob/8c4c464448af869a1e6ae6551fe54740cecd2e4f/docs/pages/components/modal/api/modal.js)

Would be something like this:
```
            {
                name: '<code>full-screen</code>',
                description: 'Option to cover the whole page.',
                type: 'String',
                values: '—',
                default: '—'
            },
```

Fixes #

## Proposed Changes
add full-screen option to ./api/modal
-
-
-

### DIFF
--- a/docs/pages/components/modal/Modal.vue
+++ b/docs/pages/components/modal/Modal.vue
@@ -27,6 +27,7 @@
 </template>
 
 <script>
+    // Update data here to include 'full-screen' option
     import api from './api/modal'
 
     import ExSimple from './examples/ExSimple'


jtommy:
Why did you add a comment? If you want to add a prop in the docs you have to update modal.js

airandopal:
Hi,
my apologies, I left the comment because I thought I wasnt able to edit the file. I found the file through search and today, reading your comment trying to  figure out where I went wrong, I realized the file I found was not on the dev branch, and that I needed to be on the dev branch in order to fork and edit. Thanks for the lesson!

PR #1772, opened 2019-09-18
Improve icon for support fab

apolokaklab:
Example: 
`<b-icon pack="fab" icon="github"/></b-icon>`
This not working because `fab` not added in `utils/icons.js`


jtommy:
Thanks

PR #1778, opened 2019-09-19
Use Bulma Variables for progress value size

service-paradis:
## Proposed Changes

- Use Bulma Variables for progress value size instead


jtommy:
Good idea

PR #1779, opened 2019-09-19
Adds aria-role options to menu list and menu item.

mateuswetah:
## Proposed Changes

- Adds `aria-role` prop to menu list component;
- Adds `aria-role` prop to menu item component;

The idea is to follow the same semantic of the a11y proposals made for dropdown. Not every developer might use the Menu component strictly for navigation, so in this case, it should be passed a role parameter of `menu` and `menu-item`. There is no need for passing `list` and `list-item` as internally list tags are already rendered.


jtommy:
Thanks!

PR #1783, opened 2019-09-20
Trap focus (Fixes #1706)

### LINKED ISSUE #1706: Add focus trap for dialog and modal.
### Description

Add focus trap for dialog and modal.

Related: #576 

Without focus trap, dialog and modal may have bugs.

Steps to reproduce:

1. Open a dialog in the document site.
2. Press `Tab` button on keyboard multiple times.
3. You can see the url of current focused `<a>` in the bottom-left of the browser (I use Firefox).
4. Press `Enter` and you can be navigated to other sites, GitHub etc.

Introduce https://github.com/davidtheclark/focus-trap may help.

service-paradis:
Fixes #1706
Related to: #576

## Proposed Changes

- Add a trap focus directive
- Add a prop `trap-focus` to dialog, dropdown and modal
- Trap focus inside the component when the prop is true

### DIFF
--- a/docs/pages/components/dialog/api/dialog.js
+++ b/docs/pages/components/dialog/api/dialog.js
@@ -116,6 +116,13 @@ export default [
                 type: 'String',
                 values: '<code>confirm</code>, <code>cancel</code>',
                 default: '<code>confirm</code>'
+            },
+            {
+                name: '<code>trap-focus</code>',
+                description: `Trap focus inside the dialog.`,
+                type: 'Boolean',
+                values: '—',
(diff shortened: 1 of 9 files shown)


jtommy:
Thanks a lot 🎉

service-paradis:
No problem!
I think it could be good to add this to the default config in constructor options. This could even be defaulted to true in a future breaking change. I think this is the expected behavior most of the time. What do you think?

jtommy:
Sure, it will be set to true in the next breaking version

PR #1796, opened 2019-09-26
Bugfix/1286 taginput autocomplete dropdown

### LINKED ISSUE #1286: Taginput should open clicking anywhere within container
Latest Buefy, Chrome, Osx
Taginput with autocomplete and open-on-focus

When clicking inside the visual boundaries you would expect the dropdown to appear, however the @focus listener is on the input contained within and not on the `.taginput-container.is-focusable` container. The input has no visual boundary so it simulates buggy behaviour and not useful if the user needs to know there is data they can choose from:

Example 1:

Example 2:

image

image

Related to this is: when someone deletes a tag, the input gains focus, but the dropdown does not appear.

Would be really great if this could be fixed.

wrabit:
Fixes #1286

### DIFF
--- a/src/components/autocomplete/Autocomplete.vue
+++ b/src/components/autocomplete/Autocomplete.vue
@@ -123,6 +123,15 @@ export default {
                     whiteList.push(child)
                 }
             }
+            if (this.$parent.$options.name === 'BTaginput') {
+                // Add taginput container
+                whiteList.push(this.$parent.$el)
+                // Add .tag and .delete
+                const tagInputChildren = this.$parent.$el.querySelectorAll('*')
+                for (const tagInputChild of tagInputChildren) {
+                    whiteList.push(tagInputChild)
+                }
+            }
 
             return whiteList
         },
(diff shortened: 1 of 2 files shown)


src/components/autocomplete/Autocomplete.vue
```diff
@@ -123,6 +123,15 @@ export default {
                     whiteList.push(child)
                 }
             }
+            if (this.$parent.$options.name === 'BTaginput') {
```
jtommy: Please use `_isTaginput` to check the component

wrabit: _isTaginput is not available via `this.$parent._isTaginput`

wrabit: `this.$parent.$data._isTaginput` is ok?

jtommy: Sure

jtommy:
Thanks

PR #1808, opened 2019-09-28
Add support for FontAwesome Light Style icons

vduricu:
## Proposed Changes
Add support for FontAwesome Light Style icons.

Example:
<b-icon pack="fal" icon="clock"/></b-icon>
This is not working because fal was not added in utils/icons.js


jtommy:
Thanks!

PR #1809, opened 2019-09-28
B-navbar documentation update for close on click

LaurenceJJones:
Close on click default is true rather than false update docs to reflect this. If this is not intended behaviour suggest to change default within Navbar.vue

Fixes #

## Proposed Changes

-
-
-


jtommy:
You're right!

PR #1827, opened 2019-10-04
make b-table with take number as px, or else user string directly

druppy:
Fixes #

## Proposed Changes

It would be really nice if we could use table column width as either a number, that is pixels as it is now, but if given as a string it it could be use it directly ?

This PR change the column render to handle this situation.

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -64,7 +64,7 @@
                             }"
                             :style="{
                                 width: column.width === undefined ? null :
-                                (isNaN(parseInt(column.width)) ? column.width : column.width + 'px')
+                                (isNaN(column.width) ? column.width : column.width + 'px')
                             }"
                             @click.stop="sort(column)">
                             <div
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -64,7 +64,8 @@
                             }"
                             :style="{
                                 width: column.width === undefined ? null :
-                                (column.width == 'number' ? column.width + 'px' : column.width )
+                                (typeof column.width == 'number' ? column.width + 'px' :
+                                column.width)
                             }"
                             @click.stop="sort(column)">
                             <div
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -63,7 +63,8 @@
                                 'is-sortable': column.sortable
                             }"
                             :style="{
-                                width: column.width === undefined ? null : column.width + 'px'
+                                width: column.width === undefined ? null :
+                                (column.width == 'number' ? column.width + 'px' : column.width )
                             }"
                             @click.stop="sort(column)">
                             <div
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -64,8 +64,7 @@
                             }"
                             :style="{
                                 width: column.width === undefined ? null :
-                                (typeof column.width == 'number' ? column.width + 'px' :
-                                column.width)
+                                (isNaN(parseInt(column.width)) ? column.width : column.width + 'px')
                             }"
                             @click.stop="sort(column)">
                             <div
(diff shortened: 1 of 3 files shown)


jtommy:
I should think about it, btw it's a breaking change

src/components/table/Table.vue
```diff
@@ -63,7 +63,8 @@
                                 'is-sortable': column.sortable
                             }"
                             :style="{
-                                width: column.width === undefined ? null : column.width + 'px'
+                                width: column.width === undefined ? null :
+                                (column.width == 'number' ? column.width + 'px' : column.width )
```
service-paradis: This is not really how you verify that something is a number

druppy: You are right, sorry it is really `typeof column.width == 'number'`

druppy:
It is really useful to be able to use percentage in table width, especially in a responsive environment. I realize this may brake existing code, but the only alternative is to add something like a `width_percent` property, and that could be a bit confusing too.

service-paradis:
@druppy
There could be a way to add the expected behavior without adding a breaking change.

By keeping the `column.width` as-is and by adding a new prop `column.widthUnit` defaulting to `px`. This prop could accept `String` and have a validator to accepts `px`, `pt`, `em`, `rem`, `vw`, ...

service-paradis:
`b-table` already use the CSS style `width`, which is the recommended way.
Do you have some source about `px` being the default unit? I really dont find anything saying this.

druppy:
Sorry i did not notice that we used css width, that is really nice.

I looked for the default type of css width, and I think I found it in the standart, as the [canonical unit](https://en.wikipedia.org/wiki/Canonical_units), ses this link [absolute-length](https://www.w3.org/TR/css-values-4/#absolute-lengths). 

Here it states "All of the absolute length units are compatible, and px is their canonical unit."

Did that help ?

druppy:
Ok, I see ... the only "default unit" is in quirk mode, and that is not part of any standard :-)

So we are back to detecting if the `width` property is type string, and it does not contain other than numbers, add a `px` unit ? That could be done in a computed value to keep the template less confusing.

druppy:
Where you thinking about using `parseInt()` on the `width` property (if type string), and if the string do not parse as an int, `b-table` could try to apply it directly ?

service-paradis:
Not really, because `parseInt("123%")` will return `123`.
Was thinking about [isNaN](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/isNaN)

Something like `isNaN(column.width)`.

There are the results of many conbination:
```
parseInt(123)
-> 123
parseInt("123")
-> 123
parseInt("123%")
-> 123
isNaN(parseInt("123"))
-> false
isNaN(parseInt(123))
-> false
isNaN(parseInt("123%"))
-> false
isNaN(123)
-> false
isNaN("123")
-> false
isNaN("123%")
-> true
```

druppy:
I see you point :-) `parseInt?  will parse until first non numeric char, not what we needed.

service-paradis:
You will also need to remove `package-lock.json` from the PR.
If you can add some unit tests to check some use cases, it would be great :)

druppy:
Removed lock file and added width test to unit test

jtommy:
@druppy thanks and @service-paradis thanks for the support

wrabit:
It's a bit unclear what this pull request added, couldn't see anything about column width in the docs. Can we now use percentage for column width?

service-paradis:
@wrabit 
You are right, the documentation is not up-to-date. But, yeah, you can use the prop `width` with a string value now. If so, you need to define the number and the unit, ex. `width="30%"`.

wrabit:
Ok, still not working for me. I am using the `<b-table-column>` component and see that nothing is done with `width` prop in that component.

service-paradis:
Which version of Buefy are you using?
Working fine on my side. The width is applied as a style attribute in the table header.

wrabit:
Ok, I got it, width was correctly being applied, but because my `<table>` width is 100%, width on `<th>` was not taking effect. Setting `table-layout: fixed;` on `<table>` fixes this.

PR #1828, opened 2019-10-04
Validate tabItems index

isjcoder:
## Issues
- Sends error when using dynamic tabs.
- If you remove a tab got this: TypeError: Cannot read property 'activate' of undefined

Fixes #

## Proposed Changes

- Include additional validation in changeTab method to validate tabItems index to work with dynamic tabs


jtommy:
Thanks

PR #1839, opened 2019-10-07
Closes #1830

### LINKED ISSUE #1830: Improvements to modal accessibility
### Description

Besides `trap-focus` we could add more features to improve modal accessibility for keyboard-only navigation. 

- [ ] Auto focus modal parent when opened. This could be done by invoking the element on `mounted()`. When using `<b-modal>` for example, the tag can be on a region of the html that is far from the calling element, making it difficult to reach the modal.
- [ ] Add aria-modal and role="dialog". It can be discussed if the developer might prefer to use role="alertdialog" (something I would expect to see in the Dialog component actually). When testing with screen readers, this properties help identify that the user entered a modal-like region.

There might be other ways to do it, or other improvements, this is what I can think so far. I've been simulating the use of screen readers with ChromeVox extension for GC. 

It would significantly improve overall accessibility of this component.

service-paradis:
Fixes #1830

## Proposed Changes

- Gives focus to the first focusable element on open (when using `trap-focus`)
- Add some a11y props to modal
- Add some a11y props to dialog
- Add typing for a11y props

In the next breaking release, wecould discuss about giving some defaults to a11y props.

### DIFF
--- a/docs/pages/components/dialog/api/dialog.js
+++ b/docs/pages/components/dialog/api/dialog.js
@@ -123,6 +123,20 @@ export default [
                 type: 'Boolean',
                 values: '—',
                 default: '<code>false</code>'
+            },
+            {
+                name: '<code>aria-role</code>',
+                description: `Role attribute to be passed to modal container for better accessibility.`,
+                type: 'String',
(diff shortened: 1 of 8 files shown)


jtommy:
Good, thanks!

PR #1840, opened 2019-10-07
Add an Accordion effect example for collapse

### LINKED ISSUE #592: Accordion? [req]
How come buefy doesn't have **accordion** option?

### LINKED ISSUE #1835: Adding an Accordion/Dropdown
### Description
I created an Accordion component a few years ago, and often use it alongside Buefy in my projects. I thought that it might make sense to include this component, or something similar, inside the Buefy bundle.

Here is a link to the docs [vue-bulma-accordion](https://lunrtick.github.io/vue-bulma-accordion/).

It doesn't **need** this feature, but I thought it may be nice to include.

service-paradis:
Closes #1835
Closes #592

## Proposed Changes

- Add an example in the collapse component documentation page to show how it could be used as a accordion (kind of)

### DIFF
--- a/docs/pages/components/collapse/Collapse.vue
+++ b/docs/pages/components/collapse/Collapse.vue
@@ -8,6 +8,8 @@
 
         <Example :component="ExPosition" :code="ExPositionCode" title="Position" vertical/>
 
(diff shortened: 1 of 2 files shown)


jtommy:
Ok, at the moment could be enough only an example

PR #1848, opened 2019-10-09
Taginputlabelposition

luisarruda:
## Proposed Changes

Fix the label position on the taginput when using maxtags with label-position setted to inside or on-border

### DIFF
--- a/src/scss/components/_form.scss
+++ b/src/scss/components/_form.scss
@@ -242,7 +242,6 @@ $floating-in-height: 3.25em !default;
 
 .control {
     .help.counter {
-        float: none;
         margin-left: 0.5em;
         text-align: right;
     }
--- a/src/scss/components/_form.scss
+++ b/src/scss/components/_form.scss
@@ -242,8 +242,9 @@ $floating-in-height: 3.25em !default;
 
 .control {
     .help.counter {
-        float: right;
+        float: none;
         margin-left: 0.5em;
+        text-align: right;
     }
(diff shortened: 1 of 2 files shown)


luisarruda:
This is what happens on the current version. 
IMG_20191009_005617_500
IMG_20191009_005615_219

jtommy:
@luisarruda Thanks but i'm checking it and there is an issue using two or more taginput below 
image

luisarruda:
@jtommy Sorry, I didn't notice that, now I change the ".help.counter" the input use this class too, but the change does not affect it. (on my tests)

src/scss/components/_form.scss
```diff
@@ -242,8 +242,9 @@ $floating-in-height: 3.25em !default;
 
 .control {
     .help.counter {
-        float: right;
+        float: none;
```
jtommy: Please remove it!

luisarruda: Ok, done. Because the default value of the property is none?

jtommy: Sure

jtommy: Yes, but i'm checking and float: right attribute removes the top margin, so you should restore it

jtommy:
@luisarruda i fixed it by https://github.com/buefy/buefy/commit/eac754819b29114556990844a5132025bfcf9b86, btw thanks for the tip

PR #1852, opened 2019-10-10
Add vertical prop to Tabs

service-paradis:
- Adjust various style to set the tabs vertically
- Position `is-right` still works
- Type `boxed` and `toggle` still works
- Can still be expanded
- Sizes still works

captured


jtommy:
Thanks! Very nice :)

PR #1854, opened 2019-10-10
added defaultButtonRounded to constructor options

### LINKED ISSUE #1851: Constructor option to change default value of `rounded` prop on <b-button>
### Description

Could there be a new Constructor Option called something like `defaultButtonRounded` which can be set to `true` or `false`?

The project that I'm working on requires that all buttons be rounded. Currently this would require that I manually set the rounded prop to true on every button, or create some sort of wrapper component around b-button to set rounded to true.

I could also do a global override on the `.button` bulma class to add a large border-radius but that would have some unintended side effects, like the buttons on the b-numberinput component would become fully rounded when I would want those to stay with the normal slightly rounded corners

bohdan-prylutskyi:
Fixes #1851

## Proposed Changes

-
-
-

### DIFF
--- a/docs/pages/installation/api/constructor-options.js
+++ b/docs/pages/installation/api/constructor-options.js
@@ -317,6 +317,13 @@ export default [
                 values: '—',
                 default: '<code>false</code>'
             },
+            {
+                name: '<code>defaultButtonRounded</code>',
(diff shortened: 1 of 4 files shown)


jtommy:
Thanks

Issue #1591, opened 2019-07-25
Native Datepicker issue on mobile Chrome for Android


francoislevesque:
### Overview of the problem

**Buefy** version: 0.7.10
**Vuejs** version: 2.6.10
**OS/Browser**: Chrome mobile Android
**Timezone**: America/Toronto

### Description

The native datepicker doesn't selected the correct date on Chrome Mobile for Android. I haven't tested the other mobile browsers.

### Steps to reproduce

1. Create a simple b-datepicker component.
2. Open the test app on Android Chrome
3. Selected a date
4. The registered date is one day off

### Expected behavior

The registered date should be the selected one.

### Actual behavior

The registered date is one day off.

### Video link

https://youtu.be/lklamdGzGbw

jtommy:
Did you check the value (v-model) ?
Is it incorrect?

afwn90cj93201nixr2e1re:
Try to change timezone on your phone, xD.

francoislevesque:
It is currently 18:52 in Montreal, Canada (Eastern Time).

When I select the 27 of july and console.log the output of change event, I receive the following string:
`Fri Jul 26 2019 20:00:00 GMT-0400 (Eastern Daylight Time)`

The problem does not occur when using the buefy datepicker, only with the native one.

francoislevesque:
The issue is here :

**Datepicker.vue**
```
/*
 * Parse date from string
 */
onChangeNativePicker(event) {
  const date = event.target.value
  console.log(date, new Date(date)) // I added this to see the values
  this.computedValue = date ? new Date(date) : null
},
```

`date` is the expected "2019-07-27". However, once it is converted to a Date object, it becomes *4 hours earlier* (Eastern time is -4 hours), hence the `Fri Jul 26 2019 20:00:00 GMT-0400 (Eastern Daylight Time)`.

Because of that, when the value is parsed in the input via the `formatYYYYMMDD()` methods, the getDay method returns the day in my timezone, 26, instead of the selected 27.

jtommy:
@francoislevesque indeed you can define a parser function

PR #1856, opened 2019-10-10
fixed datepicker v-model on ios devices

### LINKED ISSUE #1834: DatePicker sets invalid date on iOS Safari
### Overview of the problem

**Buefy** version: 0.8.5
**Vuejs** version: 2.6.10
**OS/Browser**: iOS 13 / Safari

### Description

Using a date picker on iOS with Safari always produces an "invalid date".
This happens on iOS 12 & iOS 13 but not on macOS Safari 12 & 13.
This makes date pickers unusable on iPhone/iPad.
This is similar to #1516 I reported earlier this year, that was fixed quickly.
Now both normal date pickers and month pickers don't work anymore.
I don't know in which Buefy Version this broke but 0.8.3 already has this problem.
Note that when initializing the v-model value, the date picker shows the correct date.
As soon as the date is then changed via the date picker, the date becomes invalid.

### Steps to reproduce

1. Open https://codepen.io/melvinfrohike/pen/LYYYdVV in iOS Safari.
2. Select any date

### Expected behavior
The 'selected date' label should show the selected date's string representation

### Actual behavior
The 'selected date' label shows "invalid date".

bohdan-prylutskyi:
Fixes #1834

## Proposed Changes

-
-
-

### DIFF
--- a/src/components/datepicker/Datepicker.vue
+++ b/src/components/datepicker/Datepicker.vue
@@ -593,7 +593,7 @@ export default {
         */
         onChangeNativePicker(event) {
             const date = event.target.value
-            this.computedValue = date ? new Date(date + ' 00:00:00') : null
+            this.computedValue = date ? new Date(date) : null
         },
 
         updateInternalState(value) {


jtommy:
@bohdan-prylutskyi did you test it on Android device too ?

jtommy:
Checking the history I found out that I added time part because there was an issue about timezone https://github.com/bohdan-prylutskyi/buefy/commit/3f7bb8fe401e6a6722ef0ec7d28c1d418b6d80ad

service-paradis:
I think, the best way to make this compatible with any browsers would be to use the `Date` constructor this way:
`new Date(year, month, day, hours, minutes, seconds)`

The function could look like this:
```
onChangeNativePicker(event) {
    const date = event.target.value
    let dateParams = []
    if (date) {
        dateParams = `${date} 00:00:00`.split(/[\s-:]/)
    }
    this.computedValue = date ? new Date(...dateParams) : null
}
```

Would need to be tested!

bohdan-prylutskyi:
@jtommy  yeah, i see. The problem is that Safari doesn't support such format 
`YYYY-MM-DD 00:00:00` for new Date(). 
According to this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse we can pass: 
`YYYY-MM-DDT00:00:00` that will fix the issue but it will add timezone offset to time in Safari.

jtommy:
Ok, please add it

service-paradis:
Just removing the hours bring back the problem from #1591.
You can see it in https://github.com/buefy/buefy/issues/1957.

PR #1858, opened 2019-10-10
improve sorting with nested properties

willhatchVista:
Fixes #

## Proposed Changes

-modify helper function 'getValueByPath' to always return a value, never null. Null values interfere with the sort function when this helper throws an error as it attempts to access a property of null
-
-


jtommy:
I think it's no correct to add it in a generic helper function because I could have the need to manage a null value in other components.

willhatchVista:
Excellent point, it should still return null when the value doesn't exist. I'll update and submit another request.

PR #1863, opened 2019-10-11
insure getValueByPath successfully returns

### LINKED ISSUE #1862: unable to reliably sort on nested property
### Overview of the problem

Sorting is inconsistent when using nested properties. In the case we are accessing a property on an undefined nested object there is an error when trying to access an index on null

**Buefy** version: [0.8.5]
**Vuejs** version: [2.5.2]
**OS/Browser**: Windows 10 Version 1803, Build 17134.1006 / Chrome Version 77.0.3865.120

### Description

Sorting fails when a nested object's property is not defined for all rows. 

For example this table will fail to sort the Nested Prop column:

```vue
<template>
  <div id="main">
    <b-table :data="data">
      <template slot-scope="data">
        <b-table-column 
          field="name" 
          label="Name" 
          sortable>
          {{ data.row.name }}
        </b-table-column>
        <b-table-column
          field="nested.prop"
          label="Nested Prop"
          sortable>
          {{ safeValueAccess(data.row, 'nested.prop') }}
        </b-table-column>
      </template>
    </b-table>
  </div>
</template>

<script>
export default {
  name: "App",
  components: {},
  data() {
    return {
      data: [
        { name: "item 1", nested: { prop: "Hello, World!" } },
        { name: "item 2" },
        { name: "item 3" }
      ]
    };
  },
  methods: {
    safeValueAccess(object, path) {
      return path.split(".").reduce((o, i) => (o || {})[i], object);
    }
  }
};
</script>

<style>
#main {
  padding: 50px;
}
</style>
```

### Steps to reproduce

See sandbox code: https://codesandbox.io/s/buefy-playground-x20uz

### Expected behavior

Column sorting should work the same as it does for any other undefined values

### Actual behavior

Sorting fails, echoing the error to the console:
```
TypeError: Cannot read property 'prop' of undefined
```

willhatchVista:
insure that getValueByPath always returns a value without throwing an error caused by missing nested objects

Fixes #

Fixes #1862 

## Proposed Changes

-improve reliability of getValueByPath helper function, insuring it does not throw error on null nested object
-
-

### DIFF
--- a/src/utils/helpers.js
+++ b/src/utils/helpers.js
@@ -2,8 +2,10 @@
  * Get value of an object property/path even if it's nested
  */
 export function getValueByPath(obj, path) {
-    const value = path.split('.').reduce((o, i) => o[i], obj)
-    return value
+    if (typeof obj === 'object'){
+        const value = path.split('.').reduce((o, i) => o ? o[i] : null, obj)
+        return value
+    }
 }
 
 /**
--- a/src/utils/helpers.js
+++ b/src/utils/helpers.js
@@ -2,7 +2,7 @@
  * Get value of an object property/path even if it's nested
  */
 export function getValueByPath(obj, path) {
-    if (typeof obj === 'object'){
+    if (typeof obj === 'object') {
         const value = path.split('.').reduce((o, i) => o ? o[i] : null, obj)
         return value
     }
--- a/src/utils/helpers.js
+++ b/src/utils/helpers.js
@@ -2,10 +2,8 @@
  * Get value of an object property/path even if it's nested
  */
 export function getValueByPath(obj, path) {
-    if (typeof obj === 'object') {
-        const value = path.split('.').reduce((o, i) => o ? o[i] : null, obj)
-        return value
-    }
+    const value = path.split('.').reduce((o, i) => o ? o[i] : null, obj)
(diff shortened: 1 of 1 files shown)


willhatchVista:
I'm confused by the linter error, as I didn't add any braces: 
/home/circleci/repo/src/utils/helpers.js
  5:33  error  Missing space before opening brace  space-before-blocks

src/utils/helpers.js
```diff
@@ -2,8 +2,10 @@
  * Get value of an object property/path even if it's nested
  */
 export function getValueByPath(obj, path) {
-    const value = path.split('.').reduce((o, i) => o[i], obj)
-    return value
+    if (typeof obj === 'object') {
```
jtommy: I think it's no correct add this check

willhatchVista: I wasn't sure if this helper might ever have a type passed in that would throw an error when accessing the index. I think this would happen in case of int being passed in

jtommy: In case of string or other it'll return null value

willhatchVista: thank you! updating to reflect

jtommy:
Thanks

Issue #1175, opened 2019-01-16
Feature Request: Multiple Dates on Date Picker


kerbo:
Provide a multiple date option for date-picker. Return selected dates to v-model in an array.

jtommy:
There is already a PR https://github.com/buefy/buefy/pull/1017.
I'm waiting changes that I requested some time ago

jtommy:
I'm sorry, you're right. 
@kerbo Contributions are always welcome 😃

afifsohaili:
I think the DatePicker component is worth refactoring. It's already quite complicated as of now. To support this behavior on top of single date and a range of date, it'll be quite a mess.

jtommy:
Close https://github.com/buefy/buefy/pull/1865

PR #1865, opened 2019-10-13
Added functionality, docs, and tests for multiple dates feature

jamenamcinteer:
Update for #1175 

## Proposed Changes
- allow multiple dates to be picked that do not need to be contiguous (not a range of dates)
- keep the datepicker open to allow selection of multiple dates
- if a selected date is re-selected, remove it from the list of selected dates
- multiple selected dates are separated by commas


jtommy:
@jamenamcinteer well done!! Thanks!

PR #1873, opened 2019-10-15
Add "expanded" prop to RadioButton 

saaso:
- Added "expanded" prop to RadioButton 
- Also css .b-radio.radio changed to "display: flex" to expanded works properly

P.S. If CSS property display: inline-flex is critical for some behavior, we can just add new:
```
.control.is-expanded .b-radio.radio {
    display: flex;
}
```
We have tested most cases of using RadioButton, and can confirm that .b-radio.radio with property display: flex; is works properly

### DIFF
--- a/src/components/radio/RadioButton.vue
+++ b/src/components/radio/RadioButton.vue
@@ -1,5 +1,5 @@
 <template>
-    <div class="control">
+    <div class="control" :class="{ 'is-expanded': expanded }">
         <label
             class="b-radio radio button"
             ref="label"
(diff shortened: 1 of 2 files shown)


saaso:
Some examples 
examples

jtommy:
@saaso did you check all examples ?
image

jtommy:
Thanks

saaso:
Thanks you too! I can also make same changes to Checkbox Button and create new Pull request

PR #1876, opened 2019-10-15
Add "expanded" prop to CheckboxButton

saaso:
- Added "expanded" prop to Checkbox Button
- Also added property "display: flex" to css selector .b-checkbox.checkbox.button to expanded works properly


jtommy:
Thanks

saaso:
@jtommy there is an error with quotes in my commit. Line 2 should be:
`	    <div class="control" :class="{ 'is-expanded': expanded }">`

jtommy:
Fixed, next time please update docs too

PR #1877, opened 2019-10-15
Close 950

### LINKED ISSUE #950: Smaller breakpoint for mobile version of dropdown.
### Overview of the problem

**Buefy** version: [0.6.6]
**Vuejs** version: [2.9.3]
**OS/Browser**: Firefox/Linux Ubuntu

### Description
The `mobile-modal` feature for drowpdowns is awesome, but it's breakpoint seems a bit "early" to me. I tested here and it goes popup from around 1088px. This resolution is surely enough for displaying small content dropdowns, such as menus. So I wonder if you could make it smaller (1024, 768?) or even better, set an option for the constructor so we can decide that by ourselves?

### Steps to reproduce

1. Open any dropdown (that hasn't had the `mobile-modal` feature disabled) on a screen smaller than 1088px; You can check that on Buefy's official codepen: https://codepen.io/pen?editors=1010

### Expected behavior
Feature available from a smaller breakpoint, such as 768px or 1024px;

### Actual behavior
Unnecessary popup version of small dropdowns on a large screen.

service-paradis:
Close #950 

## Proposed Changes

- Add a SCSS variable to control the breakpoint for mobile
- Add the variable in the documentation

### DIFF
--- a/docs/components/VariablesView.vue
+++ b/docs/components/VariablesView.vue
@@ -21,12 +21,19 @@ export default {
     props: {
         data: Array
     },
-    data() {
-        return {
-            variablesColumns: [
-                { label: 'Name', field: 'name', renderHtml: true },
-                { label: 'Default', field: 'default', renderHtml: true }
-            ]
+    computed: {
+        variablesColumns() {
+            let columns = []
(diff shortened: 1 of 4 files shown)


jtommy:
Good catch!

PR #1878, opened 2019-10-16
Search filter

### LINKED ISSUE #555: Searchable Table
Hello, How to add search box in table? Is it possible?

Thank you.

### LINKED ISSUE #888: Search Table: Feature Request
### Overview of the problem

**Buefy** version: Latest
**Vuejs** version: 2.9.6
**OS/Browser**: GNU/Linux - Mozilla Firefox

### Description

Hi. when I was working in my company, I added a new feature to Buefy. Because I needed it. Our data was very large. We needed to search in the table. So, I did small modifications for Buefy tables.

All those modifications are very basic to use. There are three props for this feature. These are;

**search-table**: This prop is giving boolean value.

**search-location**: This prop is giving string value. The values must be `is-pulled-right` or `is-pulled-left`. The default value is `is-pulled-right`

**search-place-holder**: This prop is giving string value. If the developer wants to show placeholder. The default value is `Search...`

I added a gif to understand what I'm saying. If this feature seems useful, I can open pull request to dev branch.

### Animated Gif

search-buefy

robertmarney:
Fixes #888
Fixes #555

Proposed Changes
Adds a searchable modifier to table initialization which will create a column specific search filter within the table header.

Contributors include @tonystonee, @quadra-diffusion @Vandenbriele


src/components/table/Table.vue
```diff
@@ -108,6 +108,23 @@
                             </template>
                         </th>
                     </tr>
+                    <tr v-if="searchable && hasSearchablenewColumns">
+                        <th
+                            v-for="(column, index) in newColumns"
+                            v-if="(column.visible || column.visible === undefined)"
```
jtommy: Please use `visibleColumns` computed instead so you can avoid to use `v-if`

src/components/table/Table.vue
```diff
@@ -108,6 +108,23 @@
...
+                            :key="index"
+                            :style="{ width: column.width + 'px' }">
```
jtommy: I've merged a PR about column width, please replace it with ```:style="{
                                width: column.width === undefined ? null :
                                (isNaN(column.width) ? column.width : column.width + 'px')
                            }"```

src/components/table/Table.vue
```diff
@@ -108,6 +108,23 @@
...
+                            <div class="th-wrap">
+                                <div v-if="column.searchable">
+                                    <b-field>
+                                        <b-input
+                                            v-model="filters[column.field]"
+                                            :type="column.numeric ? 'number' : 'text'" />
+                                    </b-field>
+                                </div>
```
jtommy: We don't need all, you might change it to
```
<template v-if="column.searchable">
     <b-input
         v-model="filters[column.field]"
         :type="column.numeric ? 'number' : 'text'" />
</template>
```
And remember to add bInput into components attribute of bTable because it's important for a correct bundle

viralpatel380: I wants to full-text search in this [Table](https://imgur.com/a/5gasuPL)

jtommy: @viralpatel380 please wait, it will be merged after this review

robertmarney: @jtommy - Just to confirm I understand your comment about adding bInput, that is just add 

import Input from '../input/Input'
...
...
components: {
       ...
       ...

        [Input.name]: Input,

Thanks for the thorough review!

jtommy: Correct!

robertmarney:
@jtommy - apologies for all the minor commits, I am on my work computer and atom was not playing nice with es-lint. Should be good to go now.

jtommy:
Thanks

TiredFalcon:
This feature breaks with both the `checkable` and the `detailed` features, since they both add an extra column to the table.
The input fields are added to the wrong columns (i.e. if the first column should have a search field, it is shown on the "check" or on the "details" switch column).
Example with searchable on the first column (ID):
https://codepen.io/tiredfalcon/pen/poowqNQ?&editable=true&editors=101

dauriata:
It seems to break while using custom template. Looks like the whole component disappear when typing in the search area if data is more than one character long.
https://codepen.io/zouloufr/pen/JjjvPxq?editors=1011
 ( try typing '0' or '2' in the Id search field, '1' works)

Issue #1764, opened 2019-09-15
SVG Icon Sizing does not work properly


service-paradis:
### Overview of the problem

**Buefy** version: 0.7.3
**Vuejs** version: latest
**OS/Browser**: any

### Description

When using SVG icon packs (vue-fontawesome for example), it is not always possible to use the sizing properties.
This is the line that is preventing it:
https://github.com/buefy/buefy/blob/dev/src/scss/components/_icon.scss#L11

It can be seen, for example, when using the message element with icons. The icon is supposed to be large (fa-3x).

### Steps to reproduce

1. Use Vue-Fontawesome
2. Add a message componentn with icon
3. Check the results

### Expected behavior

![image](https://user-images.githubusercontent.com/12817388/64926277-b9603880-d7c9-11e9-8c90-61d76c785f95.png)
![image](https://user-images.githubusercontent.com/12817388/64926287-cf6df900-d7c9-11e9-8f3f-987372ea8c9b.png)

### Actual behavior

![image](https://user-images.githubusercontent.com/12817388/64926270-a8172c00-d7c9-11e9-8935-d8e4c15304ff.png)

mateullull:
I have exactly the same issue. with SVG, the `size` attr doesn't work (icons always have the same size).

service-paradis:
You can now customize the default icon size using `$icon-svg-width` and `$icon-svg-height`.
You can, for example, redefine them to `auto`.

Sogl:
@service-paradis So how to use your solution with variables? With default width and height I can't grow my icons more than 2x.

Before (`font-size` doesn't affect):
![image](https://user-images.githubusercontent.com/7556290/69269097-5fbd2580-0be1-11ea-9282-7167405a720d.png)

After (font-size works):
![image](https://user-images.githubusercontent.com/7556290/69269250-aad73880-0be1-11ea-84cc-973b1e8dde35.png)

service-paradis:
@Sogl 
I'm not sure  understand what you want exactly. On my side I did not want he Buefy default size for icon. So I customize SCSS variables as below:
```
$icon-svg-width: auto;
$icon-svg-height: auto;
```

Sogl:
Ok, clear, these lines works fine.

liukuo362573:
I meet the same issue,width can be overrided by browser,but height doesn't override

liukuo362573:
by the way,"auto" works for me

Sogl:
[...]

PR #1879, opened 2019-10-17
Add variables to control width/height of icon svg

service-paradis:
Not really a fix for #1764, but can be a workaround


jtommy:
At the moment it's enough, thanks

Issue #535, opened 2018-01-16
Checking multiple table rows when "shift" key is held


nikazooz:
First of all, I'd like to use this opportunity to thank you for sharing these amazing UI components. :)

I wanted to ask if checking multiple table rows when holding the shift key is something that would be possible to implement, either in parent component or as part of the table component.

I kinda need such feature and was wondering if you'd accept a PR for something like that if I manage to get it working?

Kind regards and thanks again!

jbbr:
Still interested :)

PR #1894, opened 2019-10-21
Implement checking table rows with shift key

nikazooz:
This PR adds the ability to check multiple table rows if clicking on checkbox while holding shift key.

Closes https://github.com/buefy/buefy/issues/535

## Proposed Changes

- New state variable is added to keep track of index for the last checked row;
- When row is checked with shift key, handling is delegated to a new method that goes through all rows between currently checked and previously checked rows (including the two), uncheck all, and depending whether rows should be checked - checks them;
- This decision is made based on the state of last clicked checkbox: if it was checked before all rows are unchecked and if it wasn't all are checked;
- In order to detect whether the shift key was down, we need to listen for clicked event instead of changed event on checkbox component. It is possible to do it this way because here checkbox component is controlled - it's state is determined from the state of the parent (`checkedRows`)

I'd really like some opinion on this approach, so feel free to point out any possible issue that I might have overlooked, or maybe better solutions :slightly_smiling_face: 

Cheers!

P.S. Thanks for the great set of UI components! :slightly_smiling_face:

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -168,8 +168,7 @@
                                 <b-checkbox
                                     :disabled="!isRowCheckable(row)"
                                     :value="isRowChecked(row)"
-                                    @change.native="checkRow(row)"
-                                    @click.native.stop
+                                    @click.native.prevent.stop="checkRow(row, index, $event)"
                                 />
                             </td>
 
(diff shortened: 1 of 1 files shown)


jtommy:
@nikazooz it looks good!

PR #1897, opened 2019-10-22
Refactor Table/TableColumn to use provide/inject (fixes #1881)

### LINKED ISSUE #1881: Use provide-inject in b-table-column instead of $parent
### Description

The `b-table-column` component uses the `$parent` option internally to update the tables `newColumns` variable.

The same can be done using Vue's provide-inject options: https://vuejs.org/v2/api/#provide-inject

> Note: the provide and inject bindings are NOT reactive. This is intentional. However, if you pass down an observed object, properties on that object do remain reactive.

Because passed down properties are reactive the `newColumns` array should stay reactive when being mutated by the `splice()` and `push()` methods inside the `b-table-column` component.

```
Error: You should wrap bTableColumn on a bTable
```

Because the `b-table-column` component requires the `$parent` option it restricts the use of nested components inside the table.

My project requires me to have table cell values to be directly editable by clicking on them. Below the LineItem component is required to update a single item (or row) of an array in the Vuex state.

bTable
image

LineItem
image

anthonkendel:
Update Table/TableColumn to use provide/inject
instead of calling methods on the $parent property.

To do this a new utility module provide.js is added
beside the Table and TableColumn components.

provide.js exports Symbols for each method and
a utility function that creates the provide Vue
object.

Fixes #1881

## Proposed Changes

-
-
-

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -284,6 +284,7 @@ import Pagination from '../pagination/Pagination'
 import SlotComponent from '../../utils/SlotComponent'
 import TableMobileSort from './TableMobileSort'
 import TableColumn from './TableColumn'
+import { createProvideOptions } from './provide'
 
 export default {
     name: 'BTable',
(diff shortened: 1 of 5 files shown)


anthonkendel:
@jtommy Any progress on this PR?

jtommy:
@anthonkendel no at the moment but probably table component needs a refactor

anthonkendel:
> @anthonkendel no at the moment but probably table component needs a refactoring

Then I don't see any need for this issue/PR to be closed/merged. What do you think?

PR #1900, opened 2019-10-23
Added 404 page

apolokaklab:
## Proposed Changes

- Added 404 page


docs/data/routes.json
```diff
@@ -5,6 +5,15 @@
         "breadTitle": "Home",
         "path": "/"
     },
+    "404": {
```
jtommy: I prefer to use "NotFound" as route name

apolokaklab: Done

PR #1910, opened 2019-10-25
Update Datepicker docs for API

tsukakei:
Dear maintainers,

Though there is a description about `readonly` flag in the docs, the flag doesn't work.

As far as I've read source code of datepicker, this component accepts `editable` flag, not `readable`, as one of props.
https://github.com/buefy/buefy/blob/dev/src/components/datepicker/Datepicker.vue#L292
Actually, `editable` flag works.

## Proposed Changes

- correct Datepicker API docs.


jtommy:
You're right, thanks

PR #1911, opened 2019-10-25
Fix syntax error in IE11 and Edge

### LINKED ISSUE #1908: 0.8.3 introduced unsupported token
### Overview of the problem

**Buefy** version: [0.8.3] and higher
**Vuejs** version: [2.5.7]
**Webpack** version: [3.6.0]
**OS/Browser**: MacOS - Chrome

### Description

When using 0.8.3 or higher I get an error from webpack that it does not understand the spread operator token in:
./node_modules/buefy/src/utils/helpers.js

```
Module parse failed: Unexpected token (36:29)
You may need an appropriate loader to handle this file type.
|         .map((prop) =>
|             ({ [prop]: isDeep(prop) ? mergeFn(target[prop], source[prop]) : source[prop] }))
|         .reduce((a, b) => ({ ...a, ...b }), {})
| 
|     return {
```

### Steps to reproduce

try and build or hotreload with 0.8.3

### Expected behavior

Page to load normally

### Actual behavior

Page becomes unavailable with error (above) overlaying the entire page.

rhwilr:
Fixes #1908 

IE 11 and Edge do not support the object spread operator. They should get transpiled, but for some reason, when using buetify with nuxt, this line does not get transpiled. 
All other occurrences of the object spread operator do however.  

This caused out application to break in IE 11 and Edge.

## Proposed Changes

- Use the Object.assign syntax in this helper

### DIFF
--- a/src/utils/helpers.js
+++ b/src/utils/helpers.js
@@ -33,12 +33,9 @@ const mergeFn = (target, source) => {
     const replaced = Object.getOwnPropertyNames(source)
         .map((prop) =>
             ({ [prop]: isDeep(prop) ? mergeFn(target[prop], source[prop]) : source[prop] }))
-        .reduce((a, b) => ({ ...a, ...b }), {})
+        .reduce((a, b) => Object.assign({}, a, b))
 
-    return {
-        ...target,
-        ...replaced
-    }
+    return Object.assign({}, target, replaced)
 }
 export const merge = mergeFn
 


jtommy:
@rhwilr Object.assign is not supported by IE11 anyway

rhwilr:
True. However, nuxt includes core-js, which polyfills `Object.assign`.
The cause for this bug may be something completely different, since it should get transpiled in the first place.

service-paradis:
It get transpiled on my side (using Vue).
It also get transpiled in the Buefy build:
https://unpkg.com/buefy@0.8.6/dist/buefy.js (search for mergeFn).

However, `Object.assign` is not transpiled in the above version.

rhwilr:
I've found our problem. We created a custom variation of the message component and imported `buefy/src/utils/MessageMixin.js`. And of course this file does not get transpiled since it is in the node_modules folder.

PR #1928, opened 2019-10-30
Additional fix for #1918.

jomp16:
This is an additional fix for issue #1918.

## Proposed Changes

Use getValueByPath in `Number.isInteger` to check if it's a number.


jtommy:
Thanks

PR #1931, opened 2019-10-31
Replace Object.assign with custom merge function

service-paradis:
As discussed in Discord with @jtommy , here is a PR that fix IE11 problems with Object.assign.

## Proposed Changes

- Replace Object.assign with custom merge function

### DIFF
--- a/src/components/dialog/index.js
+++ b/src/components/dialog/index.js
@@ -1,6 +1,7 @@
 import Vue from 'vue'
 import Dialog from './Dialog'
 
+import { merge } from '../../utils/helpers'
 import { use, registerComponent, registerComponentProgrammatic } from '../../utils/plugins'
 
 function open(propsData) {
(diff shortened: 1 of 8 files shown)


afwn90cj93201nixr2e1re:
Why we should replace standart function, with custom stuff with spread operator inside?

afwn90cj93201nixr2e1re:
Why we cant just use spread operator?

afwn90cj93201nixr2e1re:
Vue using object spread/assign, why we should use another one for simple object without any nesting stuff?

@service-paradis

afwn90cj93201nixr2e1re:
> As discussed in Discord with @jtommy
In pm? Nothing was found in chats.
Again

afwn90cj93201nixr2e1re:
Maybe now you should refactor custom merge? (If it's have deep merging support, add new param and check is there Object.assign support or not, sometimes object assign faster 100x times and spread faster 2000x on chrome and guaranteeing no memory leaks on object's).

But i still dont get why we cant just use spread operator here if its simple object's.

service-paradis:
#1934

PR #1933, opened 2019-10-31
Render label text as html

isjcoder:
Fixes #

## Proposed Changes

- Render label text with v-html directive
- Being able to use any ASCII character in the label name


jtommy:
Why? There is already "label" slot if you need include html

PR #1934, opened 2019-10-31
Refactor merge to use Object.assign if it exists

service-paradis:
Refactor merge to use Object.assign if it exists


jtommy:
Good catch!

afwn90cj93201nixr2e1re:
Thanks.

Issue #1672, opened 2019-08-16
Feature request: CheckboxTag


dejangeci:
## CheckboxTag

![image](https://user-images.githubusercontent.com/1375514/63200378-515eeb00-c081-11e9-8668-5bada9329ebc.png)

### Description

Similar to CheckboxButton, but using tags instead.

### Why Buefy need this feature

When regular checkboxes take too much space and you need a lot of options.

### PR

I've done some initial work and can create a PR, but there are more things to consider: handling attached tags, icons, focus highlight etc. CSS tweaks are also needed, so if someone can continue the work, that would be great.

jtommy:
What are the differences between checkbox button and checkbox tag ?

dejangeci:
Functionally, no difference. Other than that, same as between button and tag. Since it would use .tag, it would be easier to match the design with other tags used throughout an app by customizing the '$tag-xxx' variables.

PR #1936, opened 2019-10-31
Add a mixin for Checkbox/Radio Components

service-paradis:


jtommy:
You're right, it's time to refactor!! :)

service-paradis:
With this mixin, it could be relatively easy to add #1672.

PR #1940, opened 2019-11-01
BREAKING: Use KeyboardEvent.key instead of keyCode

### LINKED ISSUE #1485: Taginput and cyrillic symbol "б"
### Overview of the problem

**Buefy** version: 0.7.7
**Vuejs** version: 2.6.10
**OS/Browser**: Mac OS

### Description

In the Cyrillic keyboard layout (for example in the Russian language) to type letter `б` you should press the same button that types `,` in the Latin layout (keycode: 188). After pressing this button in the taginput component, I expect to see the typed letter in the default behavior (without overriding `confirmKeyCodes`) but actually got confirmed tag. 

In the Cyrillic layout, pressing Shift+6 causes typing a comma, but I would expect to see the confirmed tag. 

As a solution, we can handle the `keypress` event in combination with `keydown`. `keypress` event has information about char code, where `б` and `,` have different codes.  

### Steps to reproduce

1. Add simple component (without overriding `confirmKeyCodes`: 
```
<b-taginput
  v-model="tags"
  ellipsis
  icon="label"
  placeholder="Add a tag">
</b-taginput>
```
2. Type word which contains symbol "б", like "Абсолютно" and press enter.
3. This word should create only one tag, but got two:

image

P.S. It's not duplication of #1148

### LINKED ISSUE #1915: Taginput: Error when pressing russian letter "Б" on entering tags
When enter the Russian letter "Б" in the tag input mode, the current tag closes and the entry of the new tag begins. The Russian letter "Б" is on the same button with <, but in the Russian layout

service-paradis:
**BREAKING CHANGE**

- Fixes #1485
- Fixes #1915

### DIFF
--- a/docs/pages/components/taginput/api/taginput.js
+++ b/docs/pages/components/taginput/api/taginput.js
@@ -100,14 +100,14 @@ export default [
                 description: 'Allow removing last tag when pressing given keys, if input is empty',
                 type: 'Array',
                 values: '—',
-                default: '<code>[8]</code>'
+                default: '<code>["Backspace"]</code>'
             },
             {
-                name: '<code>confirm-key-codes</code>',
-                description: 'Array of key codes which will add a tag when typing (default comma, enter and tab)',
+                name: '<code>confirm-keys</code>',
+                description: 'Array of key (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values) which will add a tag when typing (default comma and enter)',
                 type: 'Array',
                 values: '—',
-                default: '<code>[13, 188, 9]</code>'
+                default: '<code>[",", "Enter"]</code>'
             },
             {
                 name: '<code>on-paste-separators</code>',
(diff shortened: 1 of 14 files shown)


afwn90cj93201nixr2e1re:
I think we should discuss using key.code too. What do you think about it?

service-paradis:
Maybe. Right now, I dont know in which case it would be useful. https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code

Since the physical key can be really different between different keyboard (AZERTY vs QWERTY for example)

> This property is useful when you want to handle keys based on their physical positions on the input device rather than the characters associated with those keys; this is especially common when writing code to handle input for games which simulate a gamepad-like environment using keys on the keyboard. Be aware, however, that you can't use the value reported by KeyboardEvent.code to determine the character generated by the keystroke, because the keycode's name may not match the actual character that's printed on the key or that's generated by the computer when the key is pressed.

PR #1948, opened 2019-11-03
Fix close event not firing Fixes #1947

### LINKED ISSUE #1947: Close event on modal not firing on close nor on cancel
### Overview of the problem

**Buefy** version: [0.8.5]
**Vuejs** version: [2.6.10]
**OS/Browser**: Windows 10/Chrome

### Description

Close event not firing when using  this.$buefy.modal.open. Per documentation it should fire on cancel or close.

example here:  https://codepen.io/mgambill/pen/wvvpYKe 

### Steps to reproduce

a) click button to open modal, then click close button.

b) click button to open modal, click on the black to cancel

### Expected behavior

a) I expected close alert.
b) I expected both alerts to fire.

### Actual behavior

cancel alert fires when cancelling but never fires close event.

### Note

When stepping through the code in browser it appears the vue runtime looks at vm._events this property is empty but I can see vm.events has the expected property. Maybe vue expects events to be registered a certain way? A possible work around would be to add an onClose property similar to onCancel.

mgambill:
Fixes #1947 

## Proposed Changes

- Use the event prop directly when present.

### DIFF
--- a/src/components/modal/Modal.vue
+++ b/src/components/modal/Modal.vue
@@ -189,7 +189,7 @@ export default {
             this.$emit('close')
             this.$emit('update:active', false)
             // if the events prop has a close event call it.
-            this.events.close && this.events.close()
+            this.events && this.events.close && this.events.close()
 
             // Timeout for the animation complete before destroying
             if (this.programmatic) {
--- a/src/components/modal/Modal.vue
+++ b/src/components/modal/Modal.vue
@@ -185,9 +185,11 @@ export default {
         * Call the onCancel prop (function).
         * Emit events, and destroy modal if it's programmatic.
         */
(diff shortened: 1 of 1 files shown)


jtommy:
You don't need it, you have to listen 'close' event from modal instance getter by open method

mgambill:
@jtommy - can you explain or respond on the issue?
```
this.$buefy.modal.open({
                parent: this,
                component: ModalForm,
                hasModalCard: true,
                customClass: 'custom-class custom-class-2',
                props: { ...this.formProps },
                onCancel: () => alert("Cancelled Event fired"),
                events: {
                  close: () => alert("Close Event fired")
                }
            })
```
is this not correct?

PR #1958, opened 2019-11-05
Fix #1951: Modal a11y improvements

### LINKED ISSUE #1951: Trouble with trap-focus in modal
### Overview of the problem

I saw the trap-focus prop was added to the Modal component in 0.8.6, which is great! This is an accessibility win and I might even suggest enabling it by default at some point in the future. I'm having a problem, though - even when I set it to true, I don't see the correct behavior. Here's my example pen: https://codepen.io/dalton/pen/BaaYNGJ

As you can see, after you hit the button and then hit tab a couple of times, you're still focused on the background page, not in the modal. 

Am I doing something wrong, and would there be a way to focus on the close button upon opening the modal, similar to [Micromodal](https://micromodal.now.sh/)?

Cheers.

**Buefy** version: 0.8.6
**Vuejs** version: latest

service-paradis:
Fixes #1951

## Proposed Changes

- Fix Modal close button to be focusable (using `v-show` instead of `v-if` so it exists when the directive is bind)
- Give focus to the modal element on open so focus is inside modal when tabbing
- Remove focus within bind function (element is not yet displayed when bind is called, so it is not useful)

### DIFF
--- a/src/components/modal/Modal.vue
+++ b/src/components/modal/Modal.vue
@@ -27,7 +27,8 @@
                 <slot v-else/>
                 <button
                     type="button"
-                    v-if="showX && !animating"
+                    v-if="showX"
+                    v-show="!animating"
                     class="modal-close is-large"
                     @click="cancel('x')"/>
             </div>
--- a/src/components/modal/Modal.vue
+++ b/src/components/modal/Modal.vue
@@ -8,6 +8,7 @@
             class="modal is-active"
             :class="[{'is-full-screen': fullScreen}, customClass]"
             v-trap-focus="trapFocus"
+            tabindex="-1"
             :role="ariaRole"
             :aria-modal="ariaModal">
             <div class="modal-background" @click="cancel('outside')"/>
@@ -134,6 +135,11 @@ export default {
     watch: {
         active(value) {
             this.isActive = value
+            this.$nextTick(() => {
+                if (value && this.$el && this.$el.focus) {
+                    this.$el.focus()
+                }
+            })
         },
         isActive() {
             this.handleScroll()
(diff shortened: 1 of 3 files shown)


src/components/modal/Modal.vue
```diff
@@ -8,6 +8,7 @@
             class="modal is-active"
             :class="[{'is-full-screen': fullScreen}, customClass]"
             v-trap-focus="trapFocus"
+            tabindex="-1"
```
service-paradis: I added this so we can trigeer `focus()` on the element.

src/components/modal/Modal.vue
```diff
@@ -27,7 +28,8 @@
                 <slot v-else/>
                 <button
                     type="button"
-                    v-if="showX && !animating"
+                    v-if="showX"
+                    v-show="!animating"
```
service-paradis: I use `v-show` instead of `v-if` because I want to be able to focus on the close button when using `trap-focus`. When we use `v-if`, the element does not exist within the `bind` function of the `trapFocus` directive.

src/components/modal/Modal.vue
```diff
@@ -133,6 +135,11 @@ export default {
     watch: {
         active(value) {
             this.isActive = value
+            this.$nextTick(() => {
+                if (value && this.$el && this.$el.focus) {
+                    this.$el.focus()
+                }
+            })
```
service-paradis: This is to give the focus to the element. If we dont do this, once we open the modal and we use `TAB`, the focus is not given within the modal. Take this eample:
https://codepen.io/dalton/pen/BaaYNGJ

jtommy:
Thanks!

PR #1959, opened 2019-11-05
BREAKING: Set defaultTrapFocus to true

service-paradis:
For the next breaking release

### DIFF
--- a/src/components/clockpicker/__snapshots__/Clockpicker.spec.js.snap
+++ b/src/components/clockpicker/__snapshots__/Clockpicker.spec.js.snap
@@ -2,7 +2,7 @@
 
 exports[`BClockpicker render correctly 1`] = `
 <div class="b-clockpicker control is-primary">
-    <b-dropdown-stub mobilemodal="true" animation="fade" closeonclick="true" canclose="true" appendtobodycopyparent="true">
+    <b-dropdown-stub mobilemodal="true" animation="fade" trapfocus="true" closeonclick="true" canclose="true" appendtobodycopyparent="true">
         <b-input-stub autocomplete="off" usehtml5validation="true" type="text" hascounter="true" customclass="" readonly="true"></b-input-stub>
         <div custom="" class="card">
             <!---->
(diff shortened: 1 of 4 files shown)


afwn90cj93201nixr2e1re:
Can you explain little bit about that?
Coz really dont get why it needed here, for what is that trap Focus?

Only for broken browser's which are can't hook unfocus or what?

service-paradis:
It helps keep the focus into an open modal.
Without this directive, if you open a modal and keep using the `TAB` key, the browser will give focus to external element.
It is mostly for better accessibility

afwn90cj93201nixr2e1re:
Get you, one more, on dropdown in modal, there's no div's, which prevent parent scrolling.


Also, maybe we should just write some wrapper for buefy?
https://github.com/theKashey/react-focus-lock

service-paradis:
The directive already exists to trap focus on element. The props have been added to various components.
This PR is just to set the default behavior to true.

afwn90cj93201nixr2e1re:
https://codepen.io/arina11/pen/QWWrjGx?editors=1010

Fix that too. Coz i don't wanna.
GIF

service-paradis:
It is not related to this PR...
If there is an issue for it, I will try to take a look at it when I will have time for this.

afwn90cj93201nixr2e1re:
https://github.com/buefy/buefy/blob/dev/src/components/modal/Modal.vue#L14 there is the fix.

One more question, why it's breaking change? I don't see anything broken with `true`.

service-paradis:
Because, right now, the focus is not trapped inside those component when they are opened. By setting this to `true`, it will change this default behavior.

afwn90cj93201nixr2e1re:
I know, but what's is breaking in that?

service-paradis:
Nothing will really break, but it is a breaking change since a default behavior is changed.

Actually, someone that do not want the focus to be trapped is not using `trap-focus`. If this gets merged, they will need to add a prop to keep their current behavior.

francoism90:
Just a heads up, someone can already define this by default using [Contructor Options](https://buefy.org/documentation/constructor-options): `defaultTrapFocus: true`

PR #1961, opened 2019-11-06
Fix: Datetimepicker ignores expanded prop

cypressious:
## Proposed Changes

- Pass `expanded` prop to Datepicker

### DIFF
--- a/src/components/datetimepicker/Datetimepicker.vue
+++ b/src/components/datetimepicker/Datetimepicker.vue
@@ -9,6 +9,7 @@
         :loading="loading"
         :inline="inline"
         :editable="editable"
+        :expanded="expanded"
         :close-on-click="false"
         :date-formatter="defaultDatetimeFormatter"
         :date-parser="defaultDatetimeParser"


jtommy:
Thanks

frederic117:
Hello
When i use this and want to pass 
timepicker:{
                incrementMinutes:15,
                minTime:new Date(min),
                maxTime:new Date(max)
            },
it works for incrementMinutes but no for the minTime and maxTime (there are date object :))

Thanks for your help

jtommy:
@frederic117 you have to use max-datetime and max-datetime to datetimepicker

frederic117:
Hello 
thanks for your reply
But for exemple i want to use this to select a date between two dates (and min-datetime and max are  working fine) but the selectable time should be between 9 and 20:00 for exemple

jtommy:
For each days?

frederic117:
Yes

Le 12 nov. 2019 à 11:51, à 11:51, Walter Tommasi <notifications@github.com> a écrit:

jtommy:
You'right, but please open an issue

PR #1962, opened 2019-11-06
Emit event on Input icon click

service-paradis:
From a Discord comment

## Proposed Changes

- Add a new prop  `icon-clickable` that makes the icon clickable
- Emit `icon-click` event on click

### DIFF
--- a/docs/pages/components/input/api/input.js
+++ b/docs/pages/components/input/api/input.js
@@ -58,6 +58,13 @@ export default [
                 default: '<code>mdi</code>'
             },
             {
+                name: '<code>icon-clickable</code>',
+                description: 'Make the icon clickable',
+                type: 'Boolean',
+                values: '—',
+                default: '<code>false</code>'
+            },
+            {
                 name: '<code>maxlength</code>',
                 description: 'Same as native <code>maxlength</code>, plus character counter',
                 type: 'String, Number',
@@ -107,6 +114,11 @@ export default [
             {
                 name: '<code>blur</code>',
                 description: 'Triggers when input has lost focus',
+                parameters: '<code>event: $event</code>'
+            },
+            {
+                name: '<code>icon-click</code>',
+                description: 'Triggers when the icon is clickable and have been clicked',
                 parameters: '<code>event: $event</code>'
             },
             {
--- a/docs/pages/components/input/examples/ExIcons.vue
+++ b/docs/pages/components/input/examples/ExIcons.vue
@@ -4,7 +4,9 @@
         <b-field>
             <b-input placeholder="Search..."
                 type="search"
-                icon="magnify">
+                icon="magnify"
+                icon-clickable
+                @icon-click="searchIconClick">
             </b-input>
         </b-field>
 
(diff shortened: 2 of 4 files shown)


afwn90cj93201nixr2e1re:
@service-paradis are you sure? Coz i can't find is-disabled

afwn90cj93201nixr2e1re:
https://github.com/buefy/buefy/search?q=pointer-events&type=Code

afwn90cj93201nixr2e1re:
@click.native call also if condition is not complete.
And called on nextTick, so, there is a bug with promises or other stuff like that. I think we should just delete it, it's can't be implemented in rightWay.

service-paradis:
image
https://github.com/jgthms/bulma/blob/9a28ea17876715d00d0a8a59b9fdabfee967e56b/sass/form/tools.sass#L176

afwn90cj93201nixr2e1re:
Get you, but it's still not possible, it's not trully reactive.
You can test it, create button and ico.

Button should be async, for example with timeout 5 seconds.
Icon should have just handler for logging.

Click on button, and you also can click on ico, but event gonna be called after async function end up.
So, create variable loading: false, set it true on async function start and false on end, then try to repeat, it's also gonna be called.

afwn90cj93201nixr2e1re:
Seem's like native click ignore any pointer rules.

afwn90cj93201nixr2e1re:
https://css-tricks.com/almanac/properties/p/pointer-events/
yeah, seem's like they aren't designed for that at all.

afwn90cj93201nixr2e1re:
https://bitsofco.de/theres-no-reason-to-use-pointer-events-for-html-elements/

service-paradis:
Buttons are ont related to this PR... This PR add a click event on `b-input`...
Did you try this PR? It works... No emit when the icon is not clickable.

https://css-tricks.com/almanac/properties/p/pointer-events/ is explaining exactly the use made in this PR.

https://bitsofco.de/theres-no-reason-to-use-pointer-events-for-html-elements/
The exampe where he think it is not useful are not related to this PR. Also, it is a really opiniated article. No real arguments about having problem using it.

afwn90cj93201nixr2e1re:
> Did you try this PR? It works... No emit when the icon is not clickable.

Did you tried what i wrote?
Bind some variable and check, it's gonna be clickable after condition is not true.

afwn90cj93201nixr2e1re:
pointers are not solution for that, totally.

service-paradis:
> Did you tried what i wrote?
You want me to try to put an icon inside a button.... This PR is about an icon within an input field. Totally not related....

It would not be the same solution if it was for a button... Pointers is already the solution for the mask password thing.

If you want to use something else. Create a new PR.

afwn90cj93201nixr2e1re:
Spans arent designed for that, and there no reasons for that pr, that's not working as expected.

So, any pointers rules which arent related to **controls** should be removed from scss files.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes span is not control.

service-paradis:
Open a new issue in Bulma since pointers came from there.

Like I said, if you're not happy, create a new PR. Change the way toggling password is working right now too.

afwn90cj93201nixr2e1re:
So, pr's like that then causing issues. That's not nice practice, ok, as @jtommy merged it, it's his headache.

PR #1965, opened 2019-11-06
Useless input emits fix

afwn90cj93201nixr2e1re:
Fixes #

## Proposed Changes

-
-
-

### DIFF
--- a/src/components/tabs/Tabs.vue
+++ b/src/components/tabs/Tabs.vue
@@ -7,7 +7,7 @@
                     :key="index"
                     v-show="tabItem.visible"
                     :class="{ 'is-active': activeTab === index, 'is-disabled': tabItem.disabled }">
-                    <a @click="activeTab !== index && tabClick(index)">
+                    <a @click="tabClick(index)">
                         <template v-if="tabItem.$slots.header">
                             <b-slot-component
                                 :component="tabItem"
@@ -123,6 +123,7 @@ export default {
         * Tab click listener, emit input event and change active tab.
         */
         tabClick(value) {
+            if (this.activeTab === value) return
             this.$emit('input', value)
             this.changeTab(value)
         }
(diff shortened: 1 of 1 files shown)


jtommy:
Why do you need it? It should be avoided by css

afwn90cj93201nixr2e1re:
How it can be avoided by css?
Coz if user click on active tab, click also emited, i think it shouldn't.

jtommy:
https://github.com/buefy/buefy/blob/ef279334a259bfd3877f4cbf2cea6d4b5567840f/src/scss/components/_tabs.scss#L6

afwn90cj93201nixr2e1re:
So, we should add is-disabled to active tab or what?

jtommy:
Why do you want exclude click to not active tab ?

afwn90cj93201nixr2e1re:
I want exclude click on active tab, coz it's spaming now on clicks with input event, check one more time condition which was added. If it's not active tab then emit input.

jtommy:
I don't know if it makes sense but maybe a user wants to listen a click before changing the current tab so it could be a bug or a breaking change :smile:

afwn90cj93201nixr2e1re:
Really dont get it. How he can do it?
Click before changing gonna call not same item.

It's blocking emits only on clicking on current (active) tab.

jtommy:
I'm sorry, you are right. Please put the check in the methods and not in the template

afwn90cj93201nixr2e1re:
Ok.

afwn90cj93201nixr2e1re:
Ready.

jtommy:
Thanks

afwn90cj93201nixr2e1re:
What do you think about useless emit's inside table?

jtommy:
For example?

afwn90cj93201nixr2e1re:
https://github.com/buefy/buefy/blob/dev/src/components/table/Table.vue#L142

It's also spaming. but im not using any drag and drop stuff.

PR #1967, opened 2019-11-07
Fix #1909 Searchable table Header disappear or loses focus

### LINKED ISSUE #1909: Strange behavior on searchable table
### Overview of the problem

**Buefy** version: [0.8.6]
**Vuejs** version: [2.6.10]
**OS/Browser**: Chrome v78 on windows 10

### Description

Searchable table emits a blur event after 1 or 2 chars have been pressed.
Furthermore, if no results are found, the entire table is gone and a page refresh is needed to show the table again.

I am using it as a an attribute on the column itself, like so:

```
<template slot-scope="columns">
     <b-table-column field="first_name"
      label="first name"
      searchable>
        {{columns.row.first_name}}
    </b-table-column>
</template>
```
What can be the couse for this behavior?
Thanks.

dauriata:
Fixes #1909

## Proposed Changes

While using custom template with searchable table : 
- Prevent column from being destroyed at each cell destruction but rather only at the last one of the column => prevent focus lose on header that are destroyed and recreated at each data change when rows are filtered
- Prevent column header removal when there is no data visible ( all rows filtered)

### DIFF
--- a/src/components/table/TableColumn.vue
+++ b/src/components/table/TableColumn.vue
@@ -64,6 +64,7 @@ export default {
     },
     beforeDestroy() {
         if( !this.$parent.visibleData.length ) return
+        if ( this.$parent.$children.filter(vm => vm.$options._componentTag === 'b-table-column' && vm.$data.newKey===this.newKey).length !== 1 ) return
         const index = this.$parent.newColumns.map(
             (column) => column.newKey).indexOf(this.newKey)
         if (index >= 0) {
--- a/src/components/table/TableColumn.vue
+++ b/src/components/table/TableColumn.vue
@@ -63,6 +63,7 @@ export default {
         this.addRefToTable()
     },
     beforeDestroy() {
+        if( !this.$parent.visibleData.length ) return
         const index = this.$parent.newColumns.map(
             (column) => column.newKey).indexOf(this.newKey)
         if (index >= 0) {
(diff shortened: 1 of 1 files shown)


jtommy:
I've just checked it but doesnt' fix the issue

dauriata:
Did you checkout the `fix_searchable_table` branch ? 

Here is a codepen from the doc modified with the searchable prop:
https://codepen.io/zouloufr/pen/eYYPabJ?&editable=true&editors=101

It is referencing the buefy.min.js built from the PR : 
https://filebin.net/23ayv7fgqb3aefm0/buefy.min.js?t=syeo1ysc

jtommy:
I checked your PR with this https://github.com/buefy/buefy/pull/1975

dauriata:
This PR fixes #1967 while losing focus with custom template.

Problem with #1975 is different and lies with the column being reset (and then focus lost)  when data changes : 
`  *   2. Reset newColumns (thead), in case it's on a v-for loop.`
` this.newColumns = []` 
When you remove this line , #1975 is fixed,  I don't really understand what case with v-for that covers.

jtommy:
Ok, I'll add your fix but probably only in searchable mode. Thanks

Maxiride:
Any update on when this fix will be released?

jtommy:
@Maxiride already fixed in the previous releases

Issue #1946, opened 2019-11-02
Should we remove wrong extending?


afwn90cj93201nixr2e1re:
https://github.com/buefy/buefy/blob/899053abfea15069de85bf29f226f5bdeaa1351b/src/scss/components/_dropdown.scss#L30

for example
it's gonna extend each .dropdown-item rule with has-link a, it's really overusing for customizing.
Or for example .input extending, it's gonna also extend each .input class inside any other stuff with that rules, sometimes we doesnt need that.

https://github.com/buefy/buefy/blob/c4d4e4c059e50b8e75d1d908fc02909ad74cd7b9/src/scss/components/_taginput.scss#L6

afwn90cj93201nixr2e1re:
![изображение](https://user-images.githubusercontent.com/9306996/68073243-66f1d000-fda7-11e9-8872-ae6aac0b345a.png)
For example i have dropdown component, which is using for making another one logic.
But i can't add dropdown-item deep, coz it's adding has-link by extend, that's why i should append rules to `.special` instead `.dropdown-item`.
![изображение](https://user-images.githubusercontent.com/9306996/68073237-58a3b400-fda7-11e9-803a-97d976ce9262.png)

![изображение](https://user-images.githubusercontent.com/9306996/68073231-3f9b0300-fda7-11e9-86bd-ef062c45454a.png)

afwn90cj93201nixr2e1re:
```css
.autocomplete {
    &._custom {
        .dropdown-item {
            white-space: unset !important;
        }
    }
}
```

gonna be compiled to:
```css
.autocomplete._custom .dropdown-item, .autocomplete._custom .dropdown .dropdown-menu .has-link a, .dropdown .dropdown-menu .has-link .autocomplete._custom a {
  white-space: unset !important; }
```

useless additions

afwn90cj93201nixr2e1re:
Also this doesnt care about order:
![изображение](https://user-images.githubusercontent.com/9306996/68807529-63045e80-0681-11ea-9699-a11be821d27e.png)
before/after - same thing.

afwn90cj93201nixr2e1re:
https://github.com/buefy/buefy/blob/899053abfea15069de85bf29f226f5bdeaa1351b/src/scss/components/_dropdown.scss#L30
https://github.com/buefy/buefy/blob/c4d4e4c059e50b8e75d1d908fc02909ad74cd7b9/src/scss/components/_taginput.scss#L6

[...]

PR #1980, opened 2019-11-13
./del

afwn90cj93201nixr2e1re:
./del

### DIFF
--- a/src/components/autocomplete/Autocomplete.vue
+++ b/src/components/autocomplete/Autocomplete.vue
@@ -32,13 +32,13 @@
                 <div class="dropdown-content" v-show="isActive">
                     <div
                         v-if="hasHeaderSlot"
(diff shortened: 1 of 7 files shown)


afwn90cj93201nixr2e1re:
Not sure about this line: 
https://github.com/afwn90cj93201nixr2e1re/buefy/blob/patch-3/src/components/autocomplete/Autocomplete.vue#L344

@yxngl can you check this and https://github.com/buefy/buefy/issues/1946

Fixes #1946

afwn90cj93201nixr2e1re:
@jtommy or we can create has-link mixin and add it to each dropdown-item instead this.

Thru @include and @mixin. Whach u prefer?

afwn90cj93201nixr2e1re:
@jtommy another solution in #1981

PR #1984, opened 2019-11-15
fixes #1982, searchable able to filter booleans

### LINKED ISSUE #1982: b-table searchable bugs
### Overview of the problem

**Buefy** version: [0.8.6]
**Vuejs** version: [2.6.10]
**OS/Browser**: Chrome 78.0.3904.97 Windows 10

### Description

3 separate issues: 
1.) when laying out the table using a template to define the columns searching for a non existent value causes the table to disappear.
2.) when using the template again, the search box loses focus when the table contents are updated
3.) when searching a boolean column there is an exception thrown when we try to do pattern matching on a bool when we expected a string.

### Steps to reproduce

See `App.vue` at https://codesandbox.io/s/buefy-table-search-bug-n05i2

willhatchVista:
Fixes #

Fixes #1982

## Proposed Changes

- handle special cases when filtering:
-1.) able to filter on booleans
-2.) null values are excluded!

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -841,7 +841,7 @@ export default {
                 if (Number.isInteger(value)) {
                     if (value !== Number(this.filters[key])) return false
                 } else {
-                    var re = new RegExp(this.filters[key])
+                    const re = new RegExp(this.filters[key], 'i')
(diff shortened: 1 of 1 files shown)


jtommy:
Thanks

PR #1997, opened 2019-11-19
Component Proposal: Carousel

apolokaklab:
### Proposed Changes
- Add a new compenent for Carousel

[Demo Carousel](https://buefy-dev.netlify.com/documentation/carousel/)


afwn90cj93201nixr2e1re:
Pause is blinking.
Dont merge it to dev from dev, you should create another branch. For avoid 
изображение this stuff.

afwn90cj93201nixr2e1re:
Pause is blinking on hover.

src/components/carousel/Carousel.vue
```diff
@@ -159,11 +156,11 @@ export default {
     },
     methods: {
         startTimer() {
-            if (!this.autoplay || this.interval <= 0 || this.timer) return
+            if (!this.autoplay || this.timer) return
             this.isPause = false
             this.timer = setInterval(() => {
                 this.next()
-            }, this.interval)
+            }, (this.interval || config.defaultCarouselInterval))
```
afwn90cj93201nixr2e1re: what about situation when interval is 0?

jtommy: @afwn90cj93201nixr2e1re you shouldn't set it to zero

apolokaklab: @afwn90cj93201nixr2e1re Then back to use default config, this also work and then why you use interval 0?

afwn90cj93201nixr2e1re: Why not? immediately change on some event.

service-paradis: Interval of 0 would not make any sense for autoplay.

afwn90cj93201nixr2e1re: I know. Read https://github.com/buefy/buefy/pull/1997#discussion_r348090273 again.

service-paradis: Yeah, I read it and it does not make sense anyway. The interval is for the autoplay feature. Not for some event.

jtommy:
@apolokaklab Thanks a lot!

src/components/carousel/Carousel.vue
afwn90cj93201nixr2e1re: Seems like props which are reactive is nothing for you. :/

service-paradis:
Good job @apolokaklab!

PR #2009, opened 2019-11-21
New feature: Table subheadings

arambert:
## Proposed Changes

- Adds a `subheading` prop to columns
- Adds a `subheading` scope to columns
- Adds a `subheading` scope to tables

When present, another row in `thead` is created with the value of `subheading`.

## Why?

When using Buefy tables with long tables you cannot rely on the footer to display the table summary (totals, averages, etc.).
You can now use subheadings!! 🎉 

I've been inspired by the Google Ads & Google Analytics tables. If someone knows about big data tables, it's them! ;)

### DIFF
--- a/docs/pages/components/table/Table.vue
+++ b/docs/pages/components/table/Table.vue
@@ -69,6 +69,15 @@
             </p>
         </Example>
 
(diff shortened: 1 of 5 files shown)


jtommy:
@arambert Nice addtion! Probably we should think a way to show it on mobile (mobile-cards) but at the moment is ok for me

PR #2022, opened 2019-11-25
 Sync values on input 

afwn90cj93201nixr2e1re:
Fixes #

## Proposed Changes

-
-
-


afwn90cj93201nixr2e1re:
Need review.

src/components/input/Input.vue
```diff
@@ -101,8 +107,22 @@ export default {
...
+                if(this.forcedSync){
+                    if(this.trim)value=value.trim();
+                    else if(this.number)value = this.toNumber(value);
+                    //add your handlers and preparers here, for forced sync on conditional setters/getters inside parent computed/event's logic
+					// if(this.onlyIntegers){//allow input only integers
+						// if(!Number.isInteger(value))return;
+					// }
+					
+                }
+                if(this.prepare){//not related to v-model stuff
```
afwn90cj93201nixr2e1re: Not sure about this one, is this real needed?
Coz if we gonna update local value inside local setter with changes it's gonna be reactive too, but if it's focused right now, its not gonna be updated instantly.

src/components/input/Input.vue
```diff
@@ -101,8 +107,22 @@ export default {
...
+                    value = this.prepare(value);
+                }
+                if(this.prepareValid && !this.prepareValid(value))return;
```
afwn90cj93201nixr2e1re: Pre setter for local setters with condition's without syncOnBlur stuff, also for instant syncs. return true - to allow, false - to stop. It's before input coz we don't need to call input for setter, we can just make something like: this.computedValue = value and setter gonna be called.

src/components/input/Input.vue
```diff
@@ -181,10 +201,12 @@ export default {
     watch: {
         /**
         * When v-model is changed:
-        *   1. Set internal value.
+        *   1. Set internal value if it's needed right now
         */
         value(value) {
-            this.newValue = value
+            if (this.shouldUpdateLocalValue(this.newValue, value)) {
```
afwn90cj93201nixr2e1re: For disabling on focus update

src/components/input/Input.vue
```diff
@@ -218,6 +240,47 @@ export default {
             this.$nextTick(() => {
                 this.$refs.input.focus()
             })
+        },
+		
+		shouldUpdateLocalValue (localValue, checkVal, fromBlur = false) {
```
afwn90cj93201nixr2e1re: Default v-model stuffs.

src/components/input/Input.vue
```diff
@@ -218,6 +240,47 @@ export default {
...
+		
+        isNotInFocusAndDirty (localValue, checkVal, fromBlur) {
+            if(!fromBlur){//from value (v-model) watcher
+                if(this.isFocused && !this.forcedSync){
+                    if(this.trim||this.number)return false;//don't update value if it's trim/number and it's still focused
+                }
+                return true;//update value if it's trim/number and unfocused
+            }
+            //sync local variable on blur, default v-model behaviour
+            return localValue !== checkVal;
```
afwn90cj93201nixr2e1re: If it's gonna be called from localOnBlur and values not same we gonna force update them.

src/components/input/Input.vue
```diff
@@ -218,6 +240,47 @@ export default {
...
             })
+        },
+		
+		shouldUpdateLocalValue (localValue, checkVal, fromBlur = false) {
+            return this.isNotInFocusAndDirty(localValue, checkVal, fromBlur)||this.isDirtyWithModifiers(localValue, checkVal, fromBlur);
+        },
+		
+        isNotInFocusAndDirty (localValue, checkVal, fromBlur) {
+            if(!fromBlur){//from value (v-model) watcher
+                if(this.isFocused && !this.forcedSync){
```
afwn90cj93201nixr2e1re: If its focused and not forcedSync, for example on typing, then check if there trim or number stuff and then ignore local updates, same as v-model.trim/.number, they are updates onBlur event, but what about situations where we need instan updates, so, v-model with computed also support it, that's why it's here.

src/components/input/Input.vue
```diff
@@ -218,6 +240,47 @@ export default {
...
+                    if(this.trim||this.number)return false;//don't update value if it's trim/number and it's still focused
+                }
+                return true;//update value if it's trim/number and unfocused
```
afwn90cj93201nixr2e1re: If it's not focused or instanUpdate then update values instantly.

src/components/input/Input.vue
```diff
@@ -218,6 +240,47 @@ export default {
...
+            }
+            //sync local variable on blur, default v-model behaviour
+            return localValue !== checkVal;
+        },
+		
+        isDirtyWithModifiers (localValue, checkVal, fromBlur) {//if input have modifiers
+            if (this.number) {
+                return this.toNumber(localValue) !== this.toNumber(checkVal);//update value if it's num
+            }
+            else if(this.trim){
```
afwn90cj93201nixr2e1re: else if for disabling stupid stuff like calling .trim() on numbers and other stuff which doesnt have this methods.

src/components/input/Input.vue
```diff
@@ -218,6 +240,47 @@ export default {
...
+                return localValue.trim() !== localValue.trim();//update value if it's trim
+            }
+            //sync local variable on blur, default v-model behaviour, seems like checks before isnot needed
+            return localValue !== checkVal;
+        },
+		
+        toNumber (val) {
```
afwn90cj93201nixr2e1re: Typical v-model _n translator. Return original value if can't be translated to number.

src/components/input/Input.vue
```diff
@@ -218,6 +240,47 @@ export default {
...
+            let n = parseFloat(val);
+            return isNaN(n) ? val : n
+        },
+		
+		localOnBlur() {
+			if(this.syncOnBlur){
+				//force update local variable to value prop, same as $forceUpdate in v-model
+                //not sure about this check, but v-model doing this too
+				// if(this.shouldUpdateLocalValue(this.newValue,this.value, true)){
```
afwn90cj93201nixr2e1re: v-model adding this checks too, but it's can be heavy, that's why this is commented. So, if this.syncOnBlur then we should instantly sync value on blur.

src/utils/FormElementMixin.js
```diff
@@ -94,6 +94,7 @@ export default {
             this.isFocused = false
             this.$emit('blur', $event)
             this.checkHtml5Validity()
+			if (this.localOnBlur) this.localOnBlur()
```
afwn90cj93201nixr2e1re: Coz we can't add one more onBlur listener to template. So, it's not nice in this case to use mixins.

afwn90cj93201nixr2e1re:
```
  function shouldUpdateValue (elm, checkVal) {
	console.log('-------------------------------',elm,checkVal);
	if(elm.composing)return false;
	let a = isNotInFocusAndDirty(elm, checkVal);
	let b = isDirtyWithModifiers(elm, checkVal);
	console.log('a,b',a,b);
	return (elm.tagName === 'OPTION' || a || b);
  }

  function isNotInFocusAndDirty (elm, checkVal) {
    // return true when textbox (.number and .trim) loses focus and its value is
    // not equal to the updated value
    // usually return true on special blur event only for trim and number
    var notInFocus = true;
    // #6157
    // work around IE bug when accessing document.activeElement in an iframe
    try { notInFocus = document.activeElement !== elm; } catch (e) {}
    return notInFocus && elm.value !== checkVal
  }

  function isDirtyWithModifiers (elm, newVal) {
	//usually return true on value changes(not input but directly value changes, ex. this.value = "new string")
	//return true conditions
	//if trim and val.trim != newval.trim
	//this.val = "123" and this.val = "   123 " - false,
	//this.val = "123" and this.val = " 1234 " - true
	//if number and parseFloat(val) != parseFloat(newval)
	//this.val = "1e" and this.val = "1" - false
	//this.val = "1e1" and this.val = "10" - false
	//this.val = "1e1" and this.val = "100" - true
	//this.val = "100" and this.val = "101" - true
    var value = elm.value;
	console.log(`elm.value = "${elm.value}"`);
    var modifiers = elm._vModifiers; // injected by v-model runtime
	console.log(`value.trim() = "${value.trim()}"`);
	console.log(`newVal.trim() = "${newVal.trim()}"`);
	if (isDef(modifiers)) {
      if (modifiers.number) {
        return toNumber(value) !== toNumber(newVal)
      }
      if (modifiers.trim) {
        return value.trim() !== newVal.trim()
      }
    }
    return value !== newVal
  }
```

afwn90cj93201nixr2e1re:
Seems like code should be little more refactored(not spacing, but the logic) but still need the answer is that needed in buefy or not.

@jtommy.

src/components/input/Input.vue
afwn90cj93201nixr2e1re: should be if only

PR #2039, opened 2019-11-29
Update _modal.scss

francoism90:
Use background variable to control modal animation

Fixes #

## Proposed Changes

- This helps when using a dark theme
- It stops white flashes instead of dark

### DIFF
--- a/src/scss/components/_modal.scss
+++ b/src/scss/components/_modal.scss
@@ -6,7 +6,7 @@
             height: 100%;
             max-height: 100vh;
             margin: 0;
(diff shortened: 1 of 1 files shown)


jtommy:
It's ok for me but i'm thinking if it's a breaking change or not

francoism90:
@jtommy `$background` is used in most elements, although maybe `body-background-color` would be a better choice? At the moment it flashes (white screen for a brief moment) when using a dark background.

francoism90:
@jtommy Thanks for the merge. :)

PR #2042, opened 2019-11-30
Fixed typo in datepicker api

adrianrudnik:
Example and api doc diverge on `show-week-number` and `show-number-week`, only former works.


jtommy:
Thanks

PR #2051, opened 2019-12-03
Fix change step when stepsItem is not present

re2005:
Fixes #

## Proposed Changes

- When using the steps inside pagination, prevent error.
-
-

### DIFF
--- a/docs/pages/components/loading/variables/loading.js
+++ b/docs/pages/components/loading/variables/loading.js
@@ -9,4 +9,14 @@ export default [
         description: 'The loading background color for non-rgba compatible browsers',
         default: '<code>#7f7f7f</code>'
     },
+    {
+        name: '<code>$loading-icon-size</code>',
+        description: 'The loading icon size',
+        default: '<code>3em</code>'
+    },
+    {
+        name: '<code>$loading-full-page-icon-size</code>',
+        description: 'The loading icon size when loader is displayed over full page',
+        default: '<code>5em</code>'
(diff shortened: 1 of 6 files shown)


src/components/steps/Steps.vue
```diff
@@ -199,7 +199,7 @@ export default {
         * Change the active step and emit change event.
         */
         changeStep(newIndex) {
-            if (this.activeStep === newIndex) return
+            if (this.activeStep === newIndex || this.stepItems[newIndex]) return
```
service-paradis: I'm not sure this would keep the correct behavior...

re2005: I'm also not 100% sure.
The use case is: Having the steps inside pagination, when you go to page two, seems that the stepsArray is empty and "this.stepItems[newIndex].activate()" will fail.
It seems save to escape if there are no stepsItems.

service-paradis: But right now, the function exit when the newindex exists in the stepItems. So the code below would never run.

The condition to return should probably be something like `newIndex < this.stepItems.length` instead. So the function will return early only if the new index is not part of the step items.

re2005: You're totally right **newIndex** can never be bigger than the total index.
I've changed the PR and also added error message.
Thanks

jtommy:
Thanks

jtommy:
@re2005 now a test is failing, please check it

PR #2061, opened 2019-12-06
Fix #2014 : TagInput Autocomplete with latest Vue (2.6?)

### LINKED ISSUE #2014: Tag input is broken since latest update
### Overview of the problem

**Buefy** version: 0.8.8
**Vuejs** version: 2.6.10
**OS/Browser**: Linux/Firefox Developer Edition

### Description

See [documentation](https://buefy.org/documentation/taginput) on Autocomplete.
The options stay empty:

image

### Steps to reproduce

1. Add `taginput`
2. Select options from `autocomplete`

### Expected behavior

Show option names

### Actual behavior

Empty options

dauriata:
Fixes #2014 

TagInput seems to be broken since recent Vue update ( around 2.6 ? )
Autocomplete component evaluate hasDefaultSlot as true even if there is none.

Looks like a bug with Vue 2.6 with the change of slot semantic but simply moving the default slot after the named slots (empty, header and footer) seems to fix the issue. 

It looks like the named templates declared after the default one stay as ghost and are found in $slots and then $scopedSlots, maybe during component creation when template name is not defined yet ?

### DIFF
--- a/src/components/taginput/Taginput.vue
+++ b/src/components/taginput/Taginput.vue
@@ -46,18 +46,18 @@
                 <template :slot="headerSlotName">
                     <slot name="header" />
                 </template>
+                <template :slot="emptySlotName">
+                    <slot name="empty" />
+                </template>
+                <template :slot="footerSlotName">
+                    <slot name="footer" />
+                </template>
                 <template
                     :slot="defaultSlotName"
                     slot-scope="props">
                     <slot
                         :option="props.option"
                         :index="props.index" />
-                </template>
-                <template :slot="emptySlotName">
-                    <slot name="empty" />
-                </template>
-                <template :slot="footerSlotName">
-                    <slot name="footer" />
                 </template>
             </b-autocomplete>
         </div>


jtommy:
I tried to revert vue-template version and it works but I didn't investigate more.
Your fix is strange, did you find a similar issue?

dauriata:
For me this is a Vue bug, I needed a quick fix so I did not investigate much. I did not narrow down which change made the break.
Per my findings I can tell : 
- the new v-slot syntax works
- the old slot syntax breaks if the default is in the middle of other named template slot, but works if declared in the beginning or end
- I thought it was because multiple slots could not have the same name ( they are all named "dontrender" when not set), I gave them different names to try but that did not fix the problem

afwn90cj93201nixr2e1re:
> the old slot syntax breaks if the default is in the middle of other named template slot, but works if declared in the beginning or end

It's expected.

> I thought it was because multiple slots could not have the same name ( they are all named "dontrender" when not set)

Yes it is, but you should give diff name to 2 components, not only one.

dauriata:
> 
> 
> > the old slot syntax breaks if the default is in the middle of other named template slot, but works if declared in the beginning or end
> 
> It's expected.
Why would it be expected ? is it specified somewhere ?
> 
> > I thought it was because multiple slots could not have the same name ( they are all named "dontrender" when not set)
> 
> Yes it is, but you should give diff name to 2 components, not only one.
Well yes I tested a unique name for each of the 4 template slot and that was not the problem
> 
> We should use v-slot as expected, so @jtommy said that he works on splitting repo, just waiting for that.

jtommy:
Close https://github.com/buefy/buefy/commit/8c2ea38cedc909c7ec1c647f1b9f445311b4a3c7

PR #2063, opened 2019-12-07
Patch 3

afwn90cj93201nixr2e1re:
Fixes #

## Proposed Changes

-
-
-


afwn90cj93201nixr2e1re:
Explaining for localVueInstances - just for local using and diff Vue instances using.

jtommy:
Probably it's ok and works well but I don't love the "fix"

afwn90cj93201nixr2e1re:
But it's should works like that by default, i think  it's typo that you don't use bindings before, coz every vue module, like i18n, router and other's also using this binding style, coz we can have more than one vue instance on page and then you gonna fail with window.Vue which can be different.

jtommy:
Thanks

glumb:
Im getting an error after updating to this version. I use programmatic notifications like this:
```
Notification.open(options)
```

Some initial debugging shows that https://github.com/buefy/buefy/pull/2063/files#diff-2b7db94c0a400ab26161e55b4a5267b4R28 localVueInstance is indeed undefined.

`install(Vue)` ist not called.

I register Buefy as usual 
```
Vue.use(Buefy, {
  options
})
```

jtommy:
@glumb sure, https://buefy.org/documentation/notification#from-outside-vue-instance

glumb:
@jtommy this is how I’m using it. Unfortunately this does not work after the Buefy code changes.

jtommy:
@glumb you're right, I'm sorry..
I think a workaround is to set Vue instance to window.Vue in the main.js for example

afwn90cj93201nixr2e1re:
@glumb show an a real example, repo. I gonna investigate.

glumb:
https://github.com/glumb/test-buefy
fresh Vue CLI project, only 1 commit for the test.
Works with buefy 0.8.8, but not 0.8.9

afwn90cj93201nixr2e1re:
I'm not familiar with ts, can you show example with default js? Or with pure js everything is ok?

jtommy:
@glumb you have to import NotificationProgrammatic in this way `import { NotificationProgrammatic as Notification } from "buefy";` in the testPlugin.ts

glumb:
@jtommy indeed. thanks for the help :D

israelmuca:
Hi, I'm getting problems with the programmatic Snackbar after this update.

`var vm = typeof window !== 'undefined' && window.Vue ? window.Vue : localVueInstance || VueInstance;`
In my case, vm returns `undefined`

So I get an error:
Cannot read property 'extend' of _undefined_

On the next line:
`var SnackbarComponent = vm.extend(Snackbar);`


In my main.js I'm importing individual Buefy components:

`import Field from "buefy/dist/components/field";`
`import Input from "buefy/dist/components/input";`
`import Switch from "buefy/dist/components/switch";`
`import Dropdown from "buefy/dist/components/dropdown";`
`import Datepicker from "buefy/dist/components/datepicker";`
`import Autocomplete from "buefy/dist/components/autocomplete";`
`import { SnackbarProgrammatic as snackbar } from "buefy";`
`Vue.use(Field);`
`Vue.use(Input);` 
`Vue.use(Switch);`
`Vue.use(Dropdown);`
`Vue.use(Datepicker);`
`Vue.use(Autocomplete);`
`Vue.prototype.$snackbar = snackbar;`

I'm not sure what could be happening, before this update it was working as expected.

jtommy:
Why are you importing them from 'dist'?

israelmuca:
To be honest, I'm not sure, I think when I started this project some months ago that's how I managed to import individual components after reading some other issues here and maybe a blog post? Can't recall honestly 😟 
I'm importing their CSS individually as well but from src
`@import "~buefy/src/scss/utils/_all.scss";`
`@import "~buefy/src/scss/components/_autocomplete.scss";`
etc...

Is that incorrect?
Worth mentioning those all work great, and also the Snackbar was working great right before this update 🤔

jtommy:
Yes it worked but we removed Vue reference in Snackbar component.
Please follow my instructions and let me know

israelmuca:
Thank you so very much!

I ended up fixing my import in `main.js`
I now import everything from "buefy" directly, including the `SnackbarProgrammatic as snackbar` and also importing and registering the Snackbar plugin too.

Everything works again! Thank you so very much.

PR #2069, opened 2019-12-09
Fix: handleDetailKey null index case

Cl3MM:
This fix a bug that appears when you delete an element from the table data, in conjunction with a custom `detail-key` property.

In that case, the `index` param of the `handleDetail` method is `null` while we try to return `index[key]`

## Proposed Changes

- handle index param null case in handleDetailKey method


jtommy:
Thanks

Cl3MM:
Thank you for this awesome library :)

PR #2080, opened 2019-12-14
Fixed wrong default value on message component's API doc

hirokiky:
# Proposed Changes

Fixed wrong default value on message's API documentation.
The default value is `2000` but `5000` on the doc.

https://github.com/buefy/buefy/blob/dev/src/utils/MessageMixin.js#L29


jtommy:
Thanks

PR #2087, opened 2019-12-16
Add range start and end events to datepicker component

lunfel:
## Overview

This is not a fix, it is an enhancement. This PR targets the datepicker component when used with the `range` prop only.

## Details

When the range prop is used, it will add 2 new events: range-start and range-end which will respectively correspond to when a user selects the first date of a date range and when he selects the second date. Each of the events will receive the selected date as a parameter (as stated in the updated documentation).

## Concerns

As far as I know, the values passed in the events are Date objects, which inherently are passed by reference. As of now (not tested), I could modify the internal state of the component through the date value from the events. Should we not emit `new Date(date.getTime())` instead? I'm asking because I used the pass-by-ref as the `input` event.

```
<b-datepicker @range-start="handleRangeStart" />

[methods]

handleRangeStart (date) {
   date.setDate(25) // This would change the internal state of the datepicker component
}
```

### DIFF
--- a/docs/pages/components/datepicker/api/datepicker.js
+++ b/docs/pages/components/datepicker/api/datepicker.js
@@ -269,6 +269,16 @@ export default [
                 name: '<code>change-year</code>',
                 description: 'Triggers when calendar year is changed',
                 parameters: '<code>year: Number</code>'
+            },
+            {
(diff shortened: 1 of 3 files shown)


jtommy:
Why not ? Thanks

PR #2091, opened 2019-12-18
Fix snapshots after vue-template-compiler revert

service-paradis:


afwn90cj93201nixr2e1re:
Emmmm... We should rename our component names? Or what?

service-paradis:
Nope

Issue #1748, opened 2019-09-06
IE11 indeterminate progress styling has extra animation


service-paradis:
### Overview of the problem

**Buefy** version: 0.8.3
**OS/Browser**: IE11, Edge

### Description

There is a bug with Bulma progress bar animation when using IE11 or Edge.
Ref: https://github.com/jgthms/bulma/issues/2539

A PR is made for Bulma:
https://github.com/jgthms/bulma/pull/2540

Should we wait a Bulma update? 
Should we make a temporary fix?
Should we suggest a workaround?

afwn90cj93201nixr2e1re:
Sometimes bulma doesnt accept PR's for 2 or more years, i think you should just write wrapper inside buefy, and then disable it on pr merge.

service-paradis:
Yeah, Bulma is ont really reactive unfortunately.
What do you think @jtommy?

jtommy:
Ok add it but please add a comment too

PR #2105, opened 2019-12-24
Fix #1889: Body padding for fixed spaced navbar

### LINKED ISSUE #1889: Overlap issue when using :fixed-top and :spaced options together
### Overview of the problem

**Buefy** version: 0.8.5
**Vuejs** version: 2.6.10
**OS/Browser**: Linux/Chrome

### Description

When you use `<b-navbar :fixed-top="true" :spaced="true">` buefy adds `has-navbar-fixed-top` to `body` which causes the navbar to clip over the content at large screen widths. It should instead add `has-spaced-navbar-fixed-top` for the correct padding rules.

### Steps to reproduce

[See this codepen](https://codepen.io/tzar/pen/bGGBJGy) - make sure you're running it above 1300px wide.

### Expected behavior

Navbar does not clip over title

### Actual behavior

Above ~1300px wide, the title "Test" disappears due to the navbar overlapping it.

a-kriya:
Fixes #1889 for tablets and under.

## Proposed Changes

- Apply padding on the body for tablets and below when using a fixed spaced navbar. See the gif demoing the issue in jgthms/bulma#2737. It seems @jgthms intends for the two classes to be applied to html/body to set appropriate padding for all viewport widths.
- Add a link to the contributing guidelines in the main README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,10 @@ Buefy is an open source MIT project if you are interested in supporting this pro
   </a>
 </p>
 
+## Contributing
+
+Please refer to the contribution guidelines [here](.github/CONTRIBUTING.md).
+
 ## License
 
 Code released under [MIT](https://github.com/buefy/buefy/blob/master/LICENSE) license.
--- a/src/components/navbar/Navbar.vue
+++ b/src/components/navbar/Navbar.vue
@@ -83,24 +83,30 @@ export default {
         fixedTop: {
             handler(isSet) {
                 this.checkIfFixedPropertiesAreColliding()
-                const className = this.spaced
-                    ? BODY_SPACED_FIXED_TOP_CLASS : BODY_FIXED_TOP_CLASS
                 if (isSet) {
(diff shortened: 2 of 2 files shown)


jtommy:
I think it's better wait the merge into Bulma core code before

afwn90cj93201nixr2e1re:
Bulma is dead. We shouldn't depend on main fork. We should just fork it into buefy-bulma repo and merge some updates.

a-kriya:
@jtommy Sure.. I can also put in a TODO if you'd prefer similarly: https://github.com/buefy/buefy/issues/1748#issuecomment-529467664. Although I doubt they're going to be addressed.

jtommy:
@a-kriya all right

a-kriya:
@jtommy Updated.

tomas-kulhanek:
What does it look like with the repair? The client wants me to spaced navbar and unfortunately I am facing this error.

PR #2119, opened 2020-01-06
2118 increment hours time picker

### LINKED ISSUE #2118: Add a granularity props to datetime picker and time picker.
### Description

Add granularity props for time picker. One for minutes an other for hours at least. This will change the minutes or hours list available from the date picker.
For example if I set a granularity to 15 min, I should only be able to select 00, 15, 30, 45 in minute list.

I just get a use case for this I don't know if it's a requested feature so it's open to discussion :smile_cat:

LeoMouyna:
Fixes #2118 

## Proposed Changes

- Add an incrementHours prop in Timepicker Mixin
- Add some documentation example for Timepicker Mixin props usage in Time and DateTime pickers


jtommy:
Nice!

PR #2125, opened 2020-01-09
Support for timezone offset in Datetimepicker

### LINKED ISSUE #2013: Add timezone support to Datetimepicker
### Description

Despite what I initially thought, the support for a date creator introduced by #2000 does not allow the support of timezones in DatetimePicker/etc.

I think it would be quite useful, since the current alternative is to convert all dates sent/returned by the component, between the desired timezone and local time (used by the component), which can create quite a mess of code. A much nicer way to handle it, would be for the component to keep taking/returning universal Date objects, but allow the display to be done in a timezone other than local time.

I have come up with what seems to be [a working version of a TZ-aware `Datetimepicker` component](https://github.com/buefy/buefy/compare/dev...zedrdave:datetimepicker-with-timezone), where all input/output dates are shifted back/forth by a given offset (this also requires shifting props like `minDate` etc).

If you want, I'll be happy to submit a PR (as soon as I'm done completing support for the Native version of the component: still missing at the moment).

Two things to consider:

* Instead of providing a numeric `tzOffset` prop, I could implement a more user-friendly `timezone` ID prop (since the offset can only be accurately calculated from a TZ ID, due to DST issues, the calling code would most likely have both). This would require adding `moment-timezone` as a dependency to handle the conversions to offset.

* We could also improve the `defaultDatetimeFormatter` function to manually append the correct timezone (either as the abbreviated name, eg 'CET', or as an offset: '+01:00'). 

Let me know your thoughts!

zedrdave:
Fixes #2013

## Proposed Changes

- Adds a `tzOffset` property (in ms) that allows user to set a specific timezone for the datetime to be displayed in.


jtommy:
Why do you need it? You should be able to set your date (mindate etc) as you want as prop

zedrdave:
@jtommy this is different from mindate etc (it works fine with it): this is for the case where the date that needs to be displayed, is in a different timezone from the local computer (indicated by a tzOffset). At the moment, the only alternative, would be to do a lot of back-and-forth conversion between local time and the desired time zone to display. This prop allows to keep handling a regular timestamp, while still displaying the datetime in the desired TZ (and potentially changing the TZ as needed, without changing the time itself)…

jtommy:
@zedrdave there's something strange in your last merge, please fix it so I can merge it

zedrdave:
@jtommy I probably messed up at some point when I rebased on the latest dev.
Anyway, I generated a new pull request from scratch that should be clean enough to merge (the commit history is a mess, but you can squash it when merging):
https://github.com/buefy/buefy/pull/2177

Let me know if there's any question…

PR #2126, opened 2020-01-10
navbar - remove body fixed class on destroy

user4i:
Fixes #

## Proposed Changes

- when `navbar` is fixed and is destroyed it leaves space at top (or bottom) because of not removed class, e.g. `.has-navbar-fixed-top`
- example - when clicks on "Logout" and it is needed to hide (by v-if) top-fixed navbar = space will be on top
-  `<b-navbar fixed-top v-if="isNavBarVisible" />`

image


jtommy:
Thanks

PR #2134, opened 2020-01-14
BUGFIX: Calculate correct table column count in edge case

### LINKED ISSUE #2133: Bug in Table component: colspan miscalcualted for detailed :show-detail-icon="false"
### Overview of the problem

**Buefy** version: 0.8.9
**Vuejs** version: 2.6.10
**OS/Browser**: all

### Description

The table details row calculates a wrong colspan value, which breaks the table when used with `table-layout: fixed;`

You can find a test case here, make sure to see the demo in Desktop mode:
https://codesandbox.io/s/buefy-bug-in-table-details-rendering-2edec

### Steps to reproduce

1. Rendering a table with `detailed :show-detail-icon="false"`
2. Using `table-layout: fixed;`

### Expected behavior

Since [showDetailRowIcon](https://github.com/buefy/buefy/blob/dev/src/components/table/Table.vue#L191) decides if the td is rendered, the [columnCount method](https://github.com/buefy/buefy/blob/dev/src/components/table/Table.vue#L579) should also consider `showDetailRowIcon`instead of `detailed`.

### Actual behavior

The colspan value is to big.

rolandschuetz:
Fixes #2133

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -576,7 +576,7 @@ export default {
         columnCount() {
             let count = this.newColumns.length
             count += this.checkable ? 1 : 0
-            count += this.detailed ? 1 : 0
+            count += (this.detailed && this.showDetailIcon) ? 1 : 0
 
             return count
         }


jtommy:
@rolandschuetz Good catch!

rolandschuetz:
Wow, thanks for the super fast merge :)

PR #2138, opened 2020-01-15
Prop to trap focus inside Datepicker

service-paradis:
## Proposed Changes

- Add a Prop to trap focus inside Datepicker


jtommy:
Thanks

PR #2139, opened 2020-01-15
BUGFIX: Modal should overlay tooltip

### LINKED ISSUE #1473: Tooltip  using  'always'  is Overlay on modal
I set a tooltip using  'always' props in my main view, it work very well. 

` <b-tooltip
              label="開始年齢より終了年齢が大きい"
              :active="errCheck.age"
              type="is-danger"
              :always="errCheck.age"
            >`
image

But When I show a modal over my main view,  the tooltip overlay on the modal.
How can I resoled it?
image

rolandschuetz:
Fixes #1473

## Proposed Changes

The `.b-tooltip` `z-index` should be lower than the `.modal` one.

### DIFF
--- a/src/scss/components/_tooltip.scss
+++ b/src/scss/components/_tooltip.scss
@@ -94,7 +94,7 @@ $tooltip-multiline-sizes: (
         pointer-events: none;
     }
     &:before {
-        z-index: 889;
+        z-index: 39;
     }
     &:after {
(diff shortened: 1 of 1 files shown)


jtommy:
Probably it's ok, but we should check that it doesn't break something with other components

rolandschuetz:
Sure! In my own project, this works like a charm.

PR #2143, opened 2020-01-17
Allow "tag" prop in PascalCase

nevmn:
Following [the naming convention](https://vuejs.org/v2/guide/components-registration.html#Name-Casing) and since it's implemented in [vue-router v3.0.2](https://github.com/vuejs/vue-router/releases/tag/v3.0.2) it should be possible to pass the value of "tag" property as "RouterLink".


jtommy:
Thanks

PR #2147, opened 2020-01-17
Fix #2114

### LINKED ISSUE #2114: Limit the select options of month in Datepicker with max-date
### Description

Datepicker `max-date` should limit month as well. For example, `max-date` is set to `1-1-2020`, then the month selection should only have January as the only option.

Improvements to the UI / UX

service-paradis:
Fix #2114

## Proposed Changes

- Add a computed property for list of months
- Compare the focused year with min/max year
- Check if some months need to be disabled


jtommy:
Thanks

PR #2150, opened 2020-01-18
Bind table mouseenter and mouseleave only when table has listeners.

zzyyxxww:
## Proposed Changes
This change modifies the behaviour implemented by https://github.com/buefy/buefy/pull/1569.

Here's a gif of the current behaviour :
Peek 2020-01-17 22-12

(This is an exaggerated example, but a couple of mouse movements is enough to needlessly ruin the vue extension experience).

With this change, the Table `mouseenter` and `mouseleave` events are emitted **only** if there were listeners declared for them. Without said listeners, the events aren't bound and in consequence the "reemission" doesn't happen.

The purpose of this is making the vue extension easier to deal with and to avoid unnecessary processing usage.

I used [this comment](https://github.com/vuejs/vue/issues/7349#issuecomment-458405684) as reference.


zzyyxxww:
@jtommy thank you

Issue #74, opened 2017-05-25
[FEATURE] select with icons or dropdown that changes handler


darkylmnx:
i came accross a use case where neither the select component nor the dropdown component solves it.

If you are on facebook, you probably came accross the following image 
![image](https://cloud.githubusercontent.com/assets/1426357/26464841/6dd434bc-4189-11e7-9485-734bfbd5409f.png)

It gives you the ability to change the privacy of a post but also explains them. 

Currently, a dropdown would be the way to go (as there is html inside the options), but here's the problem now, when you make your choice, the dropdown handler / trigger does not change value, therefore the user doesn't have feedback on what he choosed.

In this sens, the behaviour seems more like a select, but because of the html issue (icons and stuff) i can't use the select component.

### Api proposal
Could you add on the drop down component a prop **replace** which would be a boolean, and if set, would replace & show the selected dropdown option instead of the handler / trigger.

rafaelpimpa:
Well, I leave the trigger opened for that purpose. You can change the trigger based on the v-model of the dropdown.

darkylmnx:
well i tried something like 
![image](https://cloud.githubusercontent.com/assets/1426357/26496431/cf29c04e-4227-11e7-81bb-92d0860ebcee.png)

and it doesn't seem to work, and btw, why don't you use a slot for html in option ? and use a label attr ? It makes code not really readable

rafaelpimpa:
Could you copy paste your code so I can test?

What you mean by html in option? Give more detailed stuff so I can help you better!

darkylmnx:
sorry, i didn't as it wasn't an issue but a feature request.

i'm using latest vue version and latest buefy version

This is what i'd like to acheive :

```
<b-dropdown v-model="is_public">
        <button class="button is-flat" type="button" slot="trigger">
            <b-icon icon="globe" v-if="is_public"></b-icon>
            <b-icon icon="users" v-if="!is_public"></b-icon>
            <span v-if="is_public">{{ $t('hit_form.public_title') }}</span>
            <span v-if="!is_public">{{ $t('hit_form.private_title') }}</span>
        </button>

        <b-dropdown-option class="columns" :value="true">
          <div class="column">
            <b-icon icon="globe"></b-icon>
          </div>
          <div class="column">
            <h3>{{ $t('hit_form.public_title') }}</h3>
            <p>{{ $t('hit_form.public_text') }}</p>
          </div>
        </b-dropdown-option>
[...]

PR #2155, opened 2020-01-21
Added multi column sorting. Fully compatible with regular and backend…

suneab:
… sorting and pagination

Fixes #
#1694

## Proposed Changes

- Adds the posibility of sorting on multiple columns
- Compatible with both frontend and backend sorting
- Compatible with both frontend and backend pagination


jtommy:
@suneab, please check my review.
I think we should add a way to reset the sort of a single column without resetting all.

suneab:
@jtommy i just pushed a change to fix the .find IE11 issue
I also removed the lodash package and added my own custom sorting method. i added it to the helper class because .sort method was interfering with array.sort

How would you like the column sort resetting to work? 

the only idea i can come up with is
click column for sorting asc <- how it works now
click same coulmn for sorting desc <- how it works now
click same column to reset it <- new change (not implimented yet!)

but i don't like it.. 

would appreciate your input.

suneab:
btw i left lodash orderBy for my fake backend sorting in the example. hope thats ok.

jtommy:
@suneab I think it's a good example https://handsontable.com/docs/7.1.1/demo-multicolumn-sorting.html

suneab:
I don't like it.. 
We keep the sorting priority when clicking the columns multiple time while only changing sorting direction, they change the priority every click.
If we added the reset on third click we would change the priority. 
Is there any other way of achieving this?

suneab:
Hi @jtommy 
I would appreciate some feedback :)

jtommy:
@suneab I don't know to be honest. Btw you should add a way to support multiple sort on mobile too (for example a dropdown)

suneab:
@jtommy i just added mobile support (IE11 is supported as well). Please let me know whats missing for this feature to be approved.

suneab:
oh btw, i couldnt figure out how to add the custom sorting icon to inside of the select options so i just added the HTML arrows
i found this thread, but its outdated and can't find anything in the current documentation about it
https://github.com/buefy/buefy/issues/74

jtommy:
@suneab any news about reset sorting column ?

suneab:
@jtommy nope, still dont know how to fit it in there

jtommy:
We have to find a way to reset or a way to enable multi sort on each column

suneab:
@jtommy i added a way to reset a single column by clicking a button. also works for mobile.

jtommy:
@suneab Nice idea! I'm checking your code and `custom-key` prop is too generic name, please change it to `sort-multiple-key`

suneab:
i fixed the naming of the custom key. Not sure what would be the correct versions in the package file

jtommy:
@suneab i don't see your commit and please reset your package-lock.json

src/components/table/Table.vue
```diff
@@ -91,7 +94,29 @@
                                 </template>
                                 <template v-else>{{ column.label }}</template>
 
+                                <template
+                                    v-if="sortMultiple &&
+                                        sortMultipleDataComputed.find(i =>
```
jtommy: Please don't use find method

src/components/table/Table.vue
```diff
@@ -91,7 +94,29 @@
...
+                                    i.field === column.field)">
+                                    <b-icon
+                                        :icon="sortIcon"
+                                        :pack="iconPack"
+                                        both
+                                        :size="sortIconSize"
+                                        :class="{ 'is-desc': sortMultipleDataComputed.find(i =>
+                                        i.field === column.field).order === 'desc'}"
+                                    />
+                                    {{ sortMultipleDataComputed.findIndex(i =>
```
jtommy: Please don't use findIndex

src/components/table/Table.vue
```diff
@@ -91,7 +94,29 @@
...
+                                    i.field === column.field) + 1 }}
+                                    <button
+                                        style="margin-left:10px"
```
jtommy: Please add a class that set margin and apply on it

suneab:
> @suneab i don't see your commit and please reset your package-lock.json

lol i forgot to press commit.. i fixed the package as well

jtommy:
@suneab please resolve conflicts so I'll merge it, thanks :)

suneab:
conflicts resolved

ms-saltation:
Thank you guys. Will try it out soon.

Issue #1751, opened 2019-09-09
inputs with left and right icon


BenjaminVockenberg:
### Description
It would be nice to have the possibility to put more than one icon into an input field. For example to show a search-icon on the left and a clear-icon on the right side of the field. In some input fields are more than one icons already (Loading, Password-reveal) but they are not customizable.
In Bulma it is possible to use the "controll"-class to add those icons:

![bulma](https://user-images.githubusercontent.com/7655131/64536222-0307d980-d319-11e9-887d-5defcac22c0c.png)

### Why Buefy need this feature
With this feature you provide more customizable form elements.

Thanks a lot and best wishes

ashimasingla20:
I will be working on this issue

BenjaminVockenberg:
thanks a lot and good success

service-paradis:
#2169 will make props available to customize the Input right icon

PR #2169, opened 2020-01-24
added clear icon button for autocomplete

### LINKED ISSUE #1954: Clear icon for autocomplete
### Description
It would be great to have a way of clearing the input of an autocomplete like for example google search does:
image but it is not exactly what I want.

UX improvement.

melogustavo:
## Proposed Changes
- [x] Fixes #1954 
- [x] Implemented button feature to clear the input search text for Autocomplete

### Preview 
ezgif com-video-to-gif

### DIFF
--- a/docs/pages/components/autocomplete/examples/ExSimple.vue
+++ b/docs/pages/components/autocomplete/examples/ExSimple.vue
@@ -8,7 +8,7 @@
                 :data="filteredDataArray"
                 placeholder="e.g. jQuery"
                 icon="magnify"
-                :clearable=true
+                clearable
                 @select="option => selected = option">
                 <template slot="empty">No results found</template>
             </b-autocomplete>
--- a/docs/pages/components/autocomplete/examples/ExSimple.vue
+++ b/docs/pages/components/autocomplete/examples/ExSimple.vue
@@ -4,10 +4,13 @@
         <b-field label="Find a JS framework">
             <b-autocomplete
                 rounded
+                :clearable=true
                 v-model="name"
                 :data="filteredDataArray"
                 placeholder="e.g. jQuery"
                 icon="magnify"
+                icon-right="times-circle"
+                icon-pack-right="fas"
                 @select="option => selected = option">
                 <template slot="empty">No results found</template>
             </b-autocomplete>
--- a/docs/pages/components/autocomplete/examples/ExSimple.vue
+++ b/docs/pages/components/autocomplete/examples/ExSimple.vue
@@ -4,13 +4,11 @@
         <b-field label="Find a JS framework">
             <b-autocomplete
                 rounded
-                :clearable=true
                 v-model="name"
                 :data="filteredDataArray"
                 placeholder="e.g. jQuery"
                 icon="magnify"
-                icon-right="times-circle"
-                icon-pack-right="fas"
+                :clearable=true
                 @select="option => selected = option">
                 <template slot="empty">No results found</template>
             </b-autocomplete>
--- a/src/components/autocomplete/Autocomplete.vue
+++ b/src/components/autocomplete/Autocomplete.vue
@@ -1,5 +1,8 @@
 <template>
-    <div class="autocomplete control has-icons-right" :class="{'is-expanded': expanded}">
+    <div
+        class="autocomplete control"
+        :class="{'is-expanded': expanded, 'has-icons-right': showClearButton}"
+    >
         <b-input
             v-model="newValue"
             ref="input"
@@ -25,9 +28,8 @@
             <template v-if="showClearButton" slot="right-icon">
                 <b-icon
                     class="is-right is-clickable"
-                    icon="times-circle"
-                    pack="fas"
-                    type="is-secondary"
+                    :icon="iconRight"
+                    :pack="iconPackRight"
                     :size="iconSize"
                     both
                     @click.native="clearInputText"/>
@@ -106,7 +108,9 @@ export default {
         customFormatter: Function,
         checkInfiniteScroll: Boolean,
         keepOpen: Boolean,
-        clearIconButton: Boolean
+        clearable: Boolean,
+        iconRight: String,
+        iconPackRight: String
     },
     data() {
         return {
@@ -182,7 +186,7 @@ export default {
          * Check if has input text
          */
         showClearButton() {
-            return this.clearIconButton && this.newValue
+            return this.clearable && this.newValue
         }
     },
     watch: {
--- a/src/components/autocomplete/Autocomplete.vue
+++ b/src/components/autocomplete/Autocomplete.vue
@@ -1,7 +1,7 @@
 <template>
     <div
         class="autocomplete control"
-        :class="{'is-expanded': expanded, 'has-icons-right': showClearButton}"
+        :class="{'is-expanded': expanded}"
     >
         <b-input
             v-model="newValue"
@@ -11,6 +11,7 @@
             :loading="loading"
             :rounded="rounded"
             :icon="icon"
+            :icon-right="clearButton ? 'close-circle' : ''"
             :icon-pack="iconPack"
             :maxlength="maxlength"
             :autocomplete="newAutocomplete"
@@ -24,17 +25,8 @@
             @keydown.native.enter.prevent="enterPressed"
             @keydown.native.up.prevent="keyArrows('up')"
             @keydown.native.down.prevent="keyArrows('down')"
-        >
-            <template v-if="showClearButton" slot="right-icon">
-                <b-icon
-                    class="is-right is-clickable"
-                    :icon="iconRight"
-                    :pack="iconPackRight"
-                    :size="iconSize"
-                    both
-                    @click.native="clearInputText"/>
-            </template>
-        </b-input>
+            @icon-right-click="clearInputText"
+        />
 
         <transition name="fade">
             <div
@@ -108,9 +100,7 @@ export default {
         customFormatter: Function,
         checkInfiniteScroll: Boolean,
         keepOpen: Boolean,
-        clearable: Boolean,
-        iconRight: String,
-        iconPackRight: String
+        clearable: Boolean
     },
     data() {
         return {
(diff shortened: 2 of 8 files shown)


afwn90cj93201nixr2e1re:
Specific, should'nt be merged.
Also your algorithm's are shitty.

jtommy:
@melogustavo i think it's probably better add a name to the new slot

melogustavo:
@jtommy hey, just added the slot name. Also changed the clear button to a property in case the user doesn't want to use it.
Thanks..

src/components/autocomplete/Autocomplete.vue
```diff
@@ -21,7 +21,18 @@
             @keydown.native.enter.prevent="enterPressed"
             @keydown.native.up.prevent="keyArrows('up')"
             @keydown.native.down.prevent="keyArrows('down')"
-        />
+        >
+            <template v-if="showClearButton" slot="right-icon">
+                <b-icon
+                    class=" is-right is-clickable"
```
sobrinho: extra space

melogustavo: thanks

src/components/autocomplete/Autocomplete.vue
```diff
@@ -21,7 +21,18 @@
             @keydown.native.enter.prevent="enterPressed"
             @keydown.native.up.prevent="keyArrows('up')"
             @keydown.native.down.prevent="keyArrows('down')"
-        />
+        >
+            <template v-if="showClearButton" slot="right-icon">
+                <b-icon
+                    class="is-right is-clickable"
+                    icon="times-circle"
+                    pack="fas"
```
jtommy: It works only with FontAwesome 5, and material icon?

src/components/autocomplete/Autocomplete.vue
```diff
@@ -94,7 +105,8 @@ export default {
         openOnFocus: Boolean,
         customFormatter: Function,
         checkInfiniteScroll: Boolean,
-        keepOpen: Boolean
+        keepOpen: Boolean,
+        clearIconButton: Boolean
```
jtommy: Please change name to 'clearable'

src/components/autocomplete/Autocomplete.vue
```diff
@@ -21,7 +21,18 @@
...
             @keydown.native.up.prevent="keyArrows('up')"
             @keydown.native.down.prevent="keyArrows('down')"
-        />
+        >
+            <template v-if="showClearButton" slot="right-icon">
+                <b-icon
+                    class="is-right is-clickable"
+                    icon="times-circle"
+                    pack="fas"
+                    type="is-secondary"
```
jtommy: Please, remove it

src/components/autocomplete/Autocomplete.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <div class="autocomplete control" :class="{'is-expanded': expanded}">
+    <div class="autocomplete control has-icons-right" :class="{'is-expanded': expanded}">
```
jtommy: Add it only if clearable prop is true

src/components/autocomplete/Autocomplete.vue
```diff
@@ -21,7 +21,18 @@
             @keydown.native.enter.prevent="enterPressed"
             @keydown.native.up.prevent="keyArrows('up')"
             @keydown.native.down.prevent="keyArrows('down')"
-        />
+        >
+            <template v-if="showClearButton" slot="right-icon">
+                <b-icon
+                    class="is-right is-clickable"
+                    icon="times-circle"
```
jtommy: Add the icon to https://github.com/buefy/buefy/blob/39e918aa1fe574bcc8096436080d99a7029f8279/src/utils/icons.js#L24 in order to automatically swap from MDI to FA and vice versa

service-paradis:
It would be great to have the option to add a custom right icon to `b-input`. This could be used by the autocomplete to add a clear icon.
Would be a little less specific and could be useful for other cases.

jtommy:
@service-paradis yes, using the new slot "right-icon"

service-paradis:
I was talking to have properties that have the same behavior as the left icon.
It is a little weird to have the possibilities to add 2 icons in the `b-input` component with 2 different ways. Using prop for left and using slot for right...

jtommy:
Probably you're right but we should decide if the right icon overwrite status icon too.
Btw we should introduce a breaking change for this request

service-paradis:
Sure, would probably need to be discussed and would probably be breaking

melogustavo:
Looking at the b-icon tag in the Input.vue file I realized that for the riight icon, you dont let the user decide wich icon to use from. It only checks for the passwordVisibleIcon or statusTypeIcon.

Captura de Tela 2020-01-28 às 09 03 16

There is no way to add a separated icon right without changing the FormElementMixin or having to add a separated property on the autoComplete.

@jtommy  what are your thoughts on this? Should it no be set pre-defined?

jtommy:
@melogustavo ok please add icon-right prop

melogustavo:
@jtommy Just sent the new commit with the icon-right prop as asked.  Also, I edited the doc and changed the default first view on the Autocomplete page, so that people can see how it's like with the clear button. In case I wasn't supposed to do this, just tell me and I will update the code.

Captura de Tela 2020-01-28 às 11 36 16

service-paradis:
I really think the `right-icon` prop should be on `b-input` component. It would help close more issues (like #1751).

jtommy:
@melogustavo i agree with @service-paradis and check my review

melogustavo:
I'm finishing up putting the right icon on the input... 

I have one question, is it to leave it only to erase the text in the input or is it to just emit an event that they can then use for different kinds of manipulation?

jtommy:
When there is a right-icon you can use it, without check empty input or other

jtommy:
@melogustavo please check my review. I don't understand your last question

src/utils/icons.js
```diff
@@ -30,7 +30,8 @@ const faIcons = () => {
             'chevron-down': 'angle-down',
             'eye-off': 'eye-slash',
             'menu-down': 'caret-down',
-            'menu-up': 'caret-up'
+            'menu-up': 'caret-up',
+            'times-circle': 'times-circle'
```
jtommy: Wrong mapping, it has to be 'close-circle':'times-circle'

docs/pages/components/autocomplete/examples/ExSimple.vue
```diff
@@ -8,6 +8,7 @@
                 :data="filteredDataArray"
                 placeholder="e.g. jQuery"
                 icon="magnify"
+                icon-right-clear="close-circle"
```
jtommy: Please remove it and add a default into autocomplete

src/components/autocomplete/Autocomplete.vue
```diff
@@ -94,7 +99,8 @@ export default {
         openOnFocus: Boolean,
         customFormatter: Function,
         checkInfiniteScroll: Boolean,
-        keepOpen: Boolean
+        keepOpen: Boolean,
+        iconRightClear: String
```
jtommy: Add a default 'close-circle'

src/components/input/Input.vue
```diff
@@ -176,6 +189,9 @@ export default {
                 return this.computedValue.toString().length
             }
             return 0
+        },
+        showRightButton() {
+            return this.iconRight && this.newValue
```
jtommy: Why input? It will be visible only when there is a icon right as prop

src/components/autocomplete/Autocomplete.vue
```diff
@@ -404,6 +410,9 @@ export default {
             const currentValue = this.getValue(this.selected)
             if (currentValue && currentValue === this.newValue) return
             this.$emit('typing', this.newValue)
+        },
+        clearInputText() {
+            this.$emit('input', '')
```
jtommy: You should clear internal value too

melogustavo:
Hey @jtommy, I just finished the changes you requested.

docs/pages/components/autocomplete/examples/ExSimple.vue
```diff
@@ -8,6 +8,7 @@
                 :data="filteredDataArray"
                 placeholder="e.g. jQuery"
                 icon="magnify"
+                :clearable=true
```
hjJunior: When is a boolean you can use like this:  

```suggestion
                clearable
```

src/components/autocomplete/Autocomplete.vue
```diff
@@ -8,6 +11,7 @@
             :loading="loading"
             :rounded="rounded"
             :icon="icon"
+            :icon-right="clearButton ? 'close-circle' : ''"
```
hjJunior: Maybe this could be a computed function, WDUT?

melogustavo: Yeah, just changed it and it looks a lot cleaner. Thanks

docs/pages/components/autocomplete/examples/ExSimple.vue
melogustavo: thanks :)

jtommy:
@melogustavo I've just added some general improvements, anyway thanks a lot!

service-paradis:
This is great!
But there is some warning in the tests:
```
console.error node_modules/vue/dist/vue.runtime.common.js:603
      [Vue warn]: $listeners is readonly.
      
      found in
      
      ---> <BInput>
             <BAutocomplete>
               <Root>
    console.error node_modules/vue/dist/vue.runtime.common.js:603
      [Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "value"
      
      found in
      
      ---> <BInput>
             <BAutocomplete>
               <Root>
    console.error node_modules/vue/dist/vue.runtime.common.js:603
      [Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "iconRight"
      
      found in
      
      ---> <BInput>
             <BAutocomplete>
               <Root>
```

Issue #2152, opened 2020-01-21
[TABS] Icons aren't in same line as the header in Custom Headers


nainardev:
### Overview of the problem

**Buefy** version: 0.8.10
**Vuejs** version: 2.6.11
**OS/Browser**: Chrome 79

### Description

Icons don't align with the Headers in Custom Headers [TABS]

### Steps to reproduce
https://buefy.org/documentation/tabs#custom-headers

### Expected behavior
Icons should be in same line as the Header

### Actual behavior
Icons not aligning with Headers

service-paradis:
I think it is related to the issue #1891. Buefy is adding a span around slot content when using the slot `header`.

The rendered element right now looks like this:
```
<li class="is-active">
  <a>
    <span>
      <span class="icon"><i class="mdi mdi-information-outline mdi-24px"></i></span> 
      <span> Issues <span class="tag is-rounded"><span class=""> 3 </span> </span></span>
    </a>
  </span>
</li>
```
instead of:
```
<li class="is-active">
  <a>
    <span class="icon"><i class="mdi mdi-information-outline mdi-24px"></i></span> 
    <span> Issues <span class="tag is-rounded"><span class=""> 3 </span> </span></span>
  </a>
</li>
```
Not sure why exactly this is the current behavior. Would probably be a breaking change to fix it.

nainardev:
So it would be fixed?

service-paradis:
Will see with @jtommy what he is thinking about this and we could fix it in the next breaking release.

jtommy:
Close https://github.com/buefy/buefy/pull/2171

PR #2171, opened 2020-01-27
Fixes #1891 and #2152

### LINKED ISSUE #1891: Tab header with tag is 1px higher than without
### Overview of the problem

**Buefy** version: [0.8.5]
**OS/Browser**: Firefox

### Description

Custom Headers tabs with a slot and a `<b-tag> `on it have 1px more in height than slots without <b-tag> so it looks weird when combined. 

### Steps to reproduce

```
<template>
    <b-tabs>
        <b-tab-item>
            <template slot="header">
                <b-icon icon="information-outline"></b-icon>
                <span> No tag </span>
            </template>
        </b-tab-item>
        <b-tab-item>
            <template slot="header">
                <b-icon icon="source-pull"></b-icon>
                <span> With tag <b-tag rounded> 2 </b-tag> </span>
            </template>
        </b-tab-item>
    </b-tabs>
</template>
```

### Expected behavior

Same height

### Actual behavior

42.5px height vs 43.5px height when having <b-tag>

service-paradis:
Fixes #1891, #2152

## Proposed Changes

- refactor the way the header slot is managed
- prevent adding an unnecessary wrapper around custom header content

### DIFF
--- a/src/components/tabs/Tabs.vue
+++ b/src/components/tabs/Tabs.vue
@@ -7,21 +7,21 @@
                     :key="index"
                     v-show="tabItem.visible"
                     :class="{ 'is-active': activeTab === index, 'is-disabled': tabItem.disabled }">
-                    <a @click="tabClick(index)">
-                        <template v-if="tabItem.$slots.header">
-                            <b-slot-component
-                                :component="tabItem"
-                                name="header"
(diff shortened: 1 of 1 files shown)


jtommy:
Great!

PR #2177, opened 2020-01-29
Datetimepicker with timezone

### LINKED ISSUE #2013: Add timezone support to Datetimepicker

zedrdave:
Fixes #2013


src/components/datetimepicker/Datetimepicker.vue
```diff
@@ -1,6 +1,6 @@
 <template>
     <b-datepicker
-        v-if="!isMobile || inline"
+        v-if="!isMobile || inline || tzOffset != 0"
```
jtommy: Why?

src/components/datetimepicker/Datetimepicker.vue
```diff
@@ -21,7 +21,7 @@
         :placeholder="placeholder"
         :range="false"
         :disabled="disabled"
-        :mobile-native="mobileNative"
+        :mobile-native="mobileNative && tzOffset == 0"
```
jtommy: And this?

zedrdave: Sorry, I should have explained: as far as I can tell, it is not possible to make the native API play along with this support. As such, `mobileNative` and `tzOffset` are mutually exclusive… Perhaps adding a warning would be a better way to address this.

jtommy: Yes, please add only warning

zedrdave: Sure. Any model of how you'd like that warning to be handled?

jtommy: In the docs?

zedrdave: Do you mean there should just be a warning in the docs? I was rather talking about a runtime warning or error, that prevents this use (otherwise the dev could end up with unpredictable behaviours without noticing).

Note that `v-if="!isMobile || inline || tzOffset != 0"` should still remain, since we don't want the code to automatically use a mobile native display if a `tzOffset` is provided…

jtommy: Ok but please add this check in s computed prop and not in the template

zedrdave: OK, I added:
```
isMobileNative() {
          return this.mobileNative && tzOffset == 0
        },
```
What do you think?

jtommy: Ok

vakrolme:
This implementation seemingly ignores the fact that the offset between two zones changes during the year because of DST.  So you really gotta have the timezone info and query the offset for a specific date like [here](https://momentjs.com/timezone/docs/#/zone-object/parse/)

zedrdave:
@vakrolme indeed (this was discussed in my initial [proposal for the PR](https://github.com/buefy/buefy/issues/2013#issue-526702347))… As long as @jtommy does not mind the addition of `moment-timezone` to the dependencies, it should not be too hard to change.

I won't have time to write the PR until later next week, though. But feel free to have a stab at it.

BuffMcBigHuge:
As per @vakrolme's note, that is indeed the case.

As a result, I did not use the tz-offset parameter in the calendar, but instead:

Initializing:

```
birthday = new Date(
    moment.unix(user.birthday)
    .utc()
    .tz(user.birthTimeZoneId)
    .format('YYYY-MM-DDTHH:mm:ss')
);
```

Saving:

```
new Date(
    moment(birthday)
        .utcOffset(0, true)
        .tz(user.birthTimeZoneId, true)
        .utc()
        .format();
);
```

where `birthday` is the value used in the datetimepicker.

zedrdave:
@BuffMcBigHuge Perhaps you can wrap the change into a new PR (merely need to add `moment-tz` in the project dependencies)?

PR #2178, opened 2020-01-29
Allow Vue Router to function cleanly with Navbar

elken:
## Proposed Changes

- Allows Vue Router to be used with Navbar
Currently, as a NavbarItem directly uses `this.$parent`, it breaks when wrapped in a `<router-item>` when trying to call `this.$parent.closeMenu` which it obviously lacks.
I could have just done a lazy `this.$parent.$parent` but I felt this is more future-proofed in case another level of indentation occurs.
This PR attempts to resolve this by passing a direct reference to the navbar as props to each navbar item. I'm relatively new to the Vuejs ecosystem so if there's a more preferable way to do this I'd be happy to amend.

### DIFF
--- a/docs/pages/components/navbar/Navbar.vue
+++ b/docs/pages/components/navbar/Navbar.vue
@@ -2,6 +2,13 @@
     <div>
         <Example :component="ExSimple" :code="ExSimpleCode"/>
 
+
+        <Example title="Vue Router">
+            <p>To work with Vue Router properly, you will need to pass <code>parent</code> to each navbar item</p>
+            <p>You can then safely wrap <code>b-navbar-item</code>s in <code>router-link</code> tags</p>
+            <Example :component="ExRouter" :code="ExRouterCode"/>
+        </Example>
+
         <ApiView :data="api"/>
     </div>
 </template>
@@ -11,13 +18,17 @@
     import api from './api/navbar'
     import ExSimple from './examples/ExSimple'
     import ExSimpleCode from '!!raw-loader!./examples/ExSimple'
+    import ExRouter from './examples/ExRouter'
+    import ExRouterCode from '!!raw-loader!./examples/ExRouter'
     
     export default {
         data() {
             return {
                 api,
                 ExSimple,
-                ExSimpleCode
+                ExSimpleCode,
+                ExRouter,
+                ExRouterCode
             }
         },
     }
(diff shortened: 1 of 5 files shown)


elken:
Confused on why codecov is complaining?

jtommy:
Why are you wrapping a Navbar item with router link ?

elken:
> 
> 
> Why are you wrapping a Navbar item with router link ?

Cleanest way to correctly manage `is-active`

jtommy:
There is already a tag and active props

elken:
> 
> 
> There is already a tag and active props

And removes the error specified complaining about parent lacking closeMenu, which also breaks the ability to close menu on mobile.

jtommy:
Can you provide a fiddle on codepen or similar?

elken:
https://codepen.io/elken/pen/qBEGPOz?editors=1011

elken:
https://codepen.io/elken/pen/KKwLXzQ?editors=1011

Pen with my version, showing the fix

service-paradis:
Hum, `navbar-item` have a prop to act as a `router-link` directly... Not sure to understand why we would wrap it inside a `router-link`...

elken:
An active prop does not a navbar-item make

service-paradis:
You mean the properties available using `v-slot` like `isExactActive`?

I think it would be preferable to improve the navbaritem component to make those available using `slot-scope` when the tag is `router-link` instead of having to wrap a `router-link` around.

So you could use something like 
```
<b-navbar-item 
    tag="router-link"
    to="/foo"
    slot-scope="{navigate, isExactActive}"
    :active="isExactActive" 
    @click="navigate"
>
    <p class="is-capitalized">Foo</p>
</b-navbar-item>
```
instead of:
```
<router-link 
    to="/foo"
    exact 
    exact-active-class="is-active"
    v-slot="{navigate, isExactActive}"
>
    <b-navbar-item :active="isExactActive" @click="navigate">
        <p class="is-capitalized">Foo</p>
    </b-navbar-item>
</router-link>
```

service-paradis:
Even easier for the end user would be to improve navbaritem to apply the `is-active` class when using the tag `router-link`.
This way, the end user code would look like this:
```
<b-navbar-item 
    tag="router-link"
    to="/foo"
>
    <p class="is-capitalized">Foo</p>
</b-navbar-item>
```
And the `is-active` class would be applied if /foo is the current page.

jtommy:
@elken did you read @service-paradise comments? What do you think about?

jtommy:
@elken did you read @service-paradis comments? What do you think about?

PR #2193, opened 2020-02-03
fix "Avoid mutating a prop directly"

service-paradis:
Fixes the Vue warning ("Avoid mutating a prop directly") when using a container. The `isFullPage` prop was mutated directly.


jtommy:
Thanks

PR #2200, opened 2020-02-05
Refactor a little bit the Input Component

service-paradis:
This is a proposal. Not mandatory. Feel free to not use it if you dont find it useful enough.

## Proposed Changes

- Simplify the Input template code
- Bring more logic to the code
- Improve separation of concerns
- Add an example using the input right icon
- Add `icon-right-click` to the doc


jtommy:
I wanted to refactor it, thanks!

PR #2207, opened 2020-02-06
WIP: Improve datepicker a11y

service-paradis:
There are probably more tests needed before merging.

## Proposed Changes

- Accept `dialog` for dropdown `aria-role`
- Refactor dialog a little bit to use a validator
- Add `aria-role="dialog"` when not inline
- Add `aria-modal="true"` when not inline
- Add `aria-label` for next/previous months
- Sync the focused date between the inner components
- Accept arrow keys to navigate between dates

### DIFF
--- a/docs/pages/components/datepicker/api/datepicker.js
+++ b/docs/pages/components/datepicker/api/datepicker.js
@@ -254,6 +254,20 @@ export default [
                 default: '<code>false</code>'
             },
             {
+                name: '<code>aria-next-label</code>',
+                description: 'Accessibility label for the next month button.',
+                type: 'String',
+                values: '—',
+                default: '—'
+            },
+            {
+                name: '<code>aria-previous-label</code>',
+                description: 'Accessibility label for the previous month button.',
+                type: 'String',
+                values: '—',
+                default: '—'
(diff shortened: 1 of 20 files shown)


jtommy:
@service-paradis Good job! Let me know when you think it's ready :)

annam002:
@service-paradis cool that it is picked up!
I checked out your branch and tested it. Nice that you can now move the dates with the arrow keys. I would recommend removing the possibility to move the dates by tab key then. Apart from that, I didn't notice any improvements with the keyboard navigation, is there more planned?

service-paradis:
@annam002 I dont want to add every accessibility improvements in a single PR.p
It is still in WIP, since I wanted to remove the possibility to change dates using tab. It could be a little breaking change if people expect it to work with tabs. Maybe, I should add a prop, but I'm not sure yet what would be the better approach for this.

annam002:
It is an anti-pattern to make it work by tab, so I don't know if people would really expect it to work that way.

service-paradis:
I just made the change to prevent changing days using tabs

jtommy:
Good job!

Issue #2208, opened 2020-02-07
Making TagInput's Tag content  customisable


jacobgoh101:
### Description

The ability to make TagInput's Tag content customisable.

### Why Buefy need this feature

Use Case: 

To implement this design, I'll need to be able to edit the HTML content for each tag. I believe this would not be an uncommon design.

![image](https://user-images.githubusercontent.com/15951104/74002507-3f179700-49aa-11ea-996f-ffddcd20b503.png)

jtommy:
4fdbf495

PR #2209, opened 2020-02-07
Making TagInput's Tag content customisable 

jacobgoh101:
Referring to https://github.com/buefy/buefy/issues/2208

## Proposed Changes

- Turning TagInput's Tag content into slot

### DIFF
--- a/src/components/taginput/Taginput.vue
+++ b/src/components/taginput/Taginput.vue
@@ -18,7 +18,10 @@
                 :closable="closable"
                 :title="ellipsis && getNormalizedTagText(tag)"
                 @close="removeTag(index)">
-                <slot name="tag" :text="getNormalizedTagText(tag)">
+                <slot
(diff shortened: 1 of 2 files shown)


src/components/taginput/Taginput.vue
```diff
@@ -18,7 +18,12 @@
                 :closable="closable"
                 :title="ellipsis && getNormalizedTagText(tag)"
                 @close="removeTag(index)">
-                {{ getNormalizedTagText(tag) }}
+                <slot
+                    name="tag"
+                    :text="getNormalizedTagText(tag)"
```
jtommy: In this case I think we might remove it because there is already tag parameter

jacobgoh101: I've removed it in the latest commit.

jtommy:
@jacobgoh101 thanks ;)

PR #2215, opened 2020-02-10
Added missing `footer` slot documentation

jacobgoh101:
Proposed Changes

- Added missing `footer` slot documentation
-
-


jtommy:
Thanks

Issue #2195, opened 2020-02-03
The switch works crookedly in desktop Safari 13.0.5


FlymeDllVa:
### Overview of the problem

**Buefy** version: buefy.org
**Vuejs** version: buefy.org
**OS/Browser**: MacOS Safari 13.0.5

### Description

<img width="105" alt="Снимок экрана 2020-02-04 в 00 17 29" src="https://user-images.githubusercontent.com/33265875/73691614-bc5db600-46e3-11ea-8911-b3b95faca99b.png">
<img width="108" alt="Снимок экрана 2020-02-04 в 00 17 44" src="https://user-images.githubusercontent.com/33265875/73691632-c54e8780-46e3-11ea-968a-1651cce97d1d.png">

Checked on your site. [Video](https://www.youtube.com/watch?v=kv7iPsIcuTg)

### Steps to reproduce

1. Go to the site via desktop Safari
2. Try clicking

jtommy:
I'm not able to test on MacOS, any contributions are welcome

yxngl:
Seems to be the problem [here](https://stackoverflow.com/questions/45586256/transition-translate-width-and-height-at-same-time-in-safari). Using `scale` indeed works fine on Safari, but it is not the same effect for rounded switches.

I think a workaround is to disable the width change of the `is-elastic` class, maybe even worth an option for people who does not like the effect.

FlymeDllVa:
I checked it in my browser. It seems to work well. I don't know how to solve it better in the end.

service-paradis:
Instead of changing the width when `is-elastic` class is there, we should try to use the transform property `scaleX(1.5)`.
Would need to rewrite part of the SCSS for this component, but it could work cross-browser. Would need to test it. Unfortunately, I'm not able to test on Safari right now.

PR #2221, opened 2020-02-10
Fix switch for Safari

FlymeDllVa:
Fix switch for Safari #2195


yxngl:
Changing the width somehow prevents the overflow. But keeping mouse down and you will notice the gap to the right when switching off is much larger than that when switching on (in all browsers). Another transform of `is-elastic` class when checked seems controlling  it.

bender-a:
I've proposed a [different change](https://github.com/buefy/buefy/pull/2263)
If we do not transition the width but transform via using scaleX, we still have the desired effect cross browser.

PR #2223, opened 2020-02-11
refactor: prevent table thead jumping when sorting is changed

re2005:
**Prevent the 'thead' items to jump when changing the sorting.**

Instead of remove the icon using 'v-show', it change visibility. So the icon is not removed with display none causing the jumping of the elements

**Preview of the problem:**
table-problem

**Preview of the fix:**
table-fixed

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -92,12 +92,12 @@
                                 <template v-else>{{ column.label }}</template>
 
                                 <b-icon
-                                    v-show="currentSortColumn === column"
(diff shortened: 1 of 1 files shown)


jtommy:
@re2005 what happened?

re2005:
Dummy me, noticed that if you set the "<b-table-column" a width this problem won't happen :)
So, this change may be not necessary.

Cheers

jtommy:
It's right but without a width could be a good idea add your class

re2005:
Let's do it!

PR #2237, opened 2020-02-14
item.first_name is undefined

LaurenceJJones:
Change item.first_name to item as data[evt.newIndex] returns a string of the moved object instead of an object

Fixes #

Instead of toast saying undefined moved from 1 to 2 , it will say the name of the item you are moving
## Proposed Changes
Change item.first_name to item as data[evt.newIndex] returns a string of the moved object instead of an object
-
-
-


jtommy:
Thanks

PR #2242, opened 2020-02-18
Add defaultDateRange prop

sebastianks:
I was in a situation where I needed the date-picker to be rendered with a date already selected. I could workaround this by adding a `ref` to the `<b-datepicker ...>` component, and then set the `dateSelected`.

But, I feel this should be part of the component itself, since there are cases where you might have some data shown on the screen that are based on a date-range, but the date-picker shows the user haven't selected anything, so the user doesn't know about the "invisible" date-range.

### DIFF
--- a/src/components/datepicker/Datepicker.vue
+++ b/src/components/datepicker/Datepicker.vue
@@ -306,6 +306,7 @@ export default {
             default: () => { return config.defaultUnselectableDaysOfWeek }
         },
         selectableDates: Array,
+        defaultDateRange: Array,
         dateFormatter: {
             type: Function,
             default: (date, vm) => {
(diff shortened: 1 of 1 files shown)


jtommy:
Why don't you use value or v-model ?

sebastianks:
@jtommy I've tried both. I can't get `:value` to work, but maybe I am using it wrong? And setting the `v-model` to my default range does not trigger the `@input` event.

I am using `value` like this:

```
<b-datepicker
   placeholder="Select a date"
    v-model="dateRange"
    @input="dateChanged"
    :value="dateRange"
    range
/>
```
I've also tried without `v-model` and with `:value` and vice versa.

jtommy:
@sebastianks what's wrong ? https://codepen.io/jtommy/pen/bGdeXmd?editable=true&editors=101

sebastianks:
Well, the `dateChanged` function that I passed to `@input` is not triggered when using `v-model` and setting it to my default date. That should be fixed.

jtommy:
It works well, btw you should use `v-model` or `:value` + `@input`

sebastianks:
@jtommy I am not sure I understand your last line, but wouldn't it make sense to have the `@input` event fired on init too?

Ah I see in regards to `:value` and `v-model`. Would it make sense to update the docs with this then? To tell that `v-model` is used on it's own, and `:value` + `@input` is used together?

PR #2255, opened 2020-02-21
Fixed typos in autocomplete and taginput docs

harrison9149:
I have updated the spelling for the miss spelt event names in the two files below.


jtommy:
Thanks

PR #2263, opened 2020-02-24
Fix #2195, fixed transition of box shadow

### LINKED ISSUE #2195: The switch works crookedly in desktop Safari 13.0.5

bender-a:
Fixes #2195

## Proposed Changes
The switch did not animate / transition correctly in safari.
Switched from transitioning the width to scaling.
Works cross browser (tested on safari, chrome, firefox).

Also:
The box shadow of the .check was only applied if the input was focused. When clicking users experienced a flash , because styles were not applied on active input state.
Changed this and added transform to box-shadow property

### DIFF
--- a/src/components/datepicker/DatepickerTable.vue
+++ b/src/components/datepicker/DatepickerTable.vue
@@ -5,7 +5,7 @@
                 v-for="(day, index) in visibleDayNames"
                 :key="index"
                 class="datepicker-cell">
-                <span>{{ day }}</span>
+                {{ day }}
             </div>
         </header>
         <div class="datepicker-body" :class="{'has-events':hasEvents}">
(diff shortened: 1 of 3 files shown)


service-paradis:
This is a great beginning! The transition is not exactly as smoother as before though.

captured (2)

When clicking slowly to see the elastic transition, it goes a little bit to the left before transitioning to the right.

bender-a:
@service-paradis thank you for the tip.
It should be working correctly now

jtommy:
@bender-a thanks a lot!

PR #2264, opened 2020-02-24
[Fix] BTable Sort Icon - Change to is-hidden

runyasak:
## Scope of work
Change hiding table sort icon from `is-invisible` to `is-hidden`

## Proposed Changes
`is-invisible` will not let label of column to be centered because the area of sort icon would be displayed.
So, I change to `is-hidden` to hide sort icon.

If you have any feedback or suggestion, please let me know. 😎

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -123,7 +123,7 @@
                                     :size="sortIconSize"
                                     :class="{
                                         'is-desc': !isAsc,
-                                        'is-invisible': currentSortColumn !== column
+                                        'is-hidden': currentSortColumn !== column
                                     }"
                                 />
                             </div>
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -92,40 +92,45 @@
                                         :index="index"
                                     />
                                 </template>
-                                <template v-else>{{ column.label }}</template>
+                                <template v-else>
+                                    <span class="is-relative">
+                                        {{ column.label }}
+                                        <template
+                                            v-if="sortMultiple &&
+                                                sortMultipleDataComputed &&
+                                                sortMultipleDataComputed.length > 0 &&
+                                                sortMultipleDataComputed.filter(i =>
+                                            i.field === column.field).length > 0">
+                                            <b-icon
+                                                :icon="sortIcon"
+                                                :pack="iconPack"
+                                                both
+                                                :size="sortIconSize"
+                                                :class="{
+                                                    'is-desc': sortMultipleDataComputed.filter(i =>
+                                                i.field === column.field)[0].order === 'desc'}"
+                                            />
+                                            {{ findIndexOfSortData(column) }}
+                                            <button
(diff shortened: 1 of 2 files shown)


service-paradis:
This will bring back the problem solved in #2223

pflirae:
I agree with service-paradis, but if the column is not sortable it should not include an invisible icon, it should be hidden, so the column is correctlly centered, this waya table can have sortable columns with the invisible icon and non sortable columns with the hidden icon.
This sounds more as an improvement than a fix to a problem.

service-paradis:
Ideally, we should have a solution that cover both cases.
- Being able to center the column label
- Prevent jumping when changing column sort

pflirae:
As a proposal, instead of changing is-invisible, try to modify from
<b-icon v-else 
to
<b-icon v-else-if="column.sortable"
this way, the icon tag will not be generated for non sortable columns, but will be generated for sortable columns, covering both cases and preventing the "jumping" columns.

runyasak:
@service-paradis Sorry, I have not seen this issue #2223 
However, I agree with this solution.

> Ideally, we should have a solution that cover both cases.
> 
> * Being able to center the column label
> * Prevent jumping when changing column sort

I have thought like @pflirae that column that not sortable should not be included with icon, even invisible.

I will find the solution, then I will update my PR soon.
Thank you 😁

runyasak:
Hi, @service-paradis @pflirae 
I have updated by position `relative` & `aboslute` with label and icon.
However, I have rearrange by icon to inside label and add some css class.

Please, review my update. 😁

jtommy:
@runyasak please check your last commit because it isn't working well

runyasak:
Hi, @jtommy 
I will check my PR and update as soon as possible.
Can you give me suggestion about which case is not working properly or any part of my code is not correct?

Thank you, for reviewing my code. 😁

jtommy:
I saw the same sort arrow icon when desc and asc :)

jtommy:
Add @pflirae solution https://github.com/buefy/buefy/commit/f9ee09f732f1dd73d02afda16309d6dad1166aae

runyasak:
> I saw the same sort arrow icon when desc and asc :)
> Btw I don't understand your PR, probably @pflirae idea is good and easy

@jtommy Sorry, for lack of my explanation. 😂
I want to fix the label of table column by move it centered. Then, move `sort-icon` to the side and make it `absolute` position.

For my desc & asc arrow icon glitch, I need to fix the class style overriding for icon.

I agree with @pflirae solution that simpler than mine but it would make it would make both of label and icon to be centered. By the way, it is depended on your design and intention that would be.

By the way, thank you for commenting.
Buefy is my first and favourite Vue UI Component. 😁

jtommy:
@runyasak ok! Do you have a ready and working solution ?

runyasak:
@jtommy I have submitted as new PR #2287.

This PR would be just "Make it work version".
If you have any feedback or suggestion, please let me know and I will fix as soon as possible. 😁

PR #2268, opened 2020-02-25
WIP: Scoped styles for component

### LINKED ISSUE #1749: Question: How do I import components and their style individually?
Hi,
I'm new to Buefy. And I'm sorry to ask this as an "issue", but how do I add components individually, without bringing extra and global css styles?

Like I want to use the Tooltip component, but in the [documention](https://buefy.org/documentation/customization/) it says I have to do:
```scss
@import "~bulma";
@import "~buefy/src/scss/buefy";
```
which brings all the extra css styles that I don't wan't. It completely messes up my page layout, my font sizes and everything.
Before using buefy, I used to import bulma like this :
```scss
@import "~bulma/sass/utilities/_all";
@import "~bulma/sass/form/_all.sass";
@import "~bulma/sass/elements/button.sass";
@import "~bulma/sass/elements/icon.sass";
```
This would allow me to have control and only import necessary components I need.

Is it possible to use Buefy like I explained?

### LINKED ISSUE #2235: CSS scoping accross the whole framework
### Overview of the problem

**Buefy** version: all

### Description

CSS across the whole framework is scattered and creates unexpected behavior. 

Example one, buefy styles creep into non-buefy components:
https://codesandbox.io/s/buefy-css-creep-2edec

Example two, some in buefy defined clases do not work scoped:
https://codesandbox.io/s/buefy-scoped-css-2kb3i

service-paradis:
Fixes #2235 
Fixes #1749

## Proposed Changes

- Scope style in every component
- Make variables available in every component scoped style

This is in WIP and will probably be better to introduce this in a breaking release. For now, as an example, I scope the style for only 2 components:

- Tooltip
- Upload


bender-a:
Have you thought about putting the styles in a seperate file?
I think if we progress that way, we will have a better directory structure and also even better auto-completition in IDEs

jtommy:
@service-paradis I agree with @bender-a

service-paradis:
Sure!

jtommy:
It's better use a class based strategy

Tofandel:
Yep, scoped styles are a big no no for me

Big dom, poor performance and unintuitive

Class based scoping is way better

PR #2275, opened 2020-02-26
Dependency updates

bender-a:
Fixes #

## Proposed Changes / Updates

- vue
- vue-router
- vue-template-compiler
- vue-loader

vueLoaderPlugin was included into the plugins of the router. 
Had to update the router to handle esModule breaking change of vue-loader (v.14)


afwn90cj93201nixr2e1re:
so, maybe now you gonna add \r \n\n\n and other stuff removing inside rollup too?

Issue #2253, opened 2020-02-21
Horizontal layout for BDatetimePicker


tecaz:
### Description

Consider moving to a more horizontal layout for the `BDatetimePicker` dropdown.
eg from this:

![before](https://user-images.githubusercontent.com/47509484/75028344-5ec9b780-54a0-11ea-941f-4bb8cbbbc7c9.png)

to this:

![after](https://user-images.githubusercontent.com/47509484/75028348-612c1180-54a0-11ea-8d1d-02b72a2d3adb.png)

### Why Buefy need this feature

After having used `BDateTimePicker` for a while, I noticed two problems my users seems to be having:
1. On devices with limited screen estate (eg laptops, notebooks), the time picker often ends up outside of the window screen (meaning you need to scroll down or up to make it visible), unless carefully positioned
2. The time picker is sometimes entirely ignored (the users only focus on the date picker)

Having the date picker and the time picker side to side could solve both of these problems.

melogustavo:
Hey, I will take a look at this and see if I can help you.

jtommy:
Close https://github.com/buefy/buefy/pull/2280#merged-event

PR #2280, opened 2020-02-28
feature Horizontal layout for BDatetimePicker

melogustavo:
## Proposed Changes

- adds feature requested in #2253
- Implements prop horizontalTimePicker that when set to true will move the time picker from the bottom to the right side of the calendar.
- Whenever the size of the screen is to small that the pop up opens, the layout will move back to default.

## Preview
Gravacao-de-Tela-2020-02-28-as-1


jtommy:
@melogustavo nice work but please unit tests

melogustavo:
Working on it now. For some reason, the local test did not fail but the build did. As soon as I fix this, I`ll commit.

PR #2287, opened 2020-03-03
[Fix] BTable Sort Icon - Fix label of table column to centered

runyasak:
## Scope of work
Move label of table column to centered.

## Proposed Changes
Fixing the label of table column by move it centered. Then, move sort-icon to the side and make it absolute position.

**Note**
- This would be continued from #2264 and I have submit PR with my working solution.
- My code is "Make it work" version. If it is not proper with standard, please let me know and I will fix as soon as possible.


jtommy:
@runyasak it works but i'm seeing the same behaviour of current implementation, what should it fix ?

runyasak:
@jtommy I don't not carried PC tonight.
I will explain with screenshot later. 😂

runyasak:
@jtommy I have recorded and export as gif.
Alt Text

Noticing to `Date` column, this column is `sortable` and the label is not centered because the `sort-icon` is invisible. There will be the space for the icon.

Alt Text
For my solution, I make the label to be centered and make `sort-icon` to be abosolute. So, the label would be always centered for `centered` prop. The `sort-icon` would be placed at the right of label.

If you have any feedback or comments, please let me know. 😁

Issue #2269, opened 2020-02-25
Is it possible to increase the slider's focus?


francoism90:
### Description

I really like the slider, but sometimes it's hard to actually click somewhere on the slider without using the thumb especially on small sizes:

![image](https://user-images.githubusercontent.com/5028905/75261136-e68a2b80-57ea-11ea-857b-bb3f2f3314d4.png)

### Why Buefy need this feature

It would be great to allow a bit of threshold so a change can be triggered, even when the user isn't actually on the slider.

I'll try to take a look at the SCSS and make a PR if this isn't possible yet.

yxngl:
Do you want to increase the width of the track or just make clickable region wider?

francoism90:
@yxngl I would like to increase the last one, the clickable region.

jtommy:
@francoism90 Probably you're right.
@yxngl @francoism90 a PR is welcome 👍

melogustavo:
Hey @francoism90 I would like to help you out.

@yxngl did you already start working on a PR for this?

jtommy:
@melogustavo make your PR :)

yxngl:
What are your thoughts and approaches to this? I didn't start. My initial idea is not to change HTML.  For workaround in a single use case, increasing the height of the track and using color gradient to fake a thinner look of track and fill will do. But these are in SCSS, do you want an API for it?

jtommy:
@yxngl if we can manage it with a prop why not ?

melogustavo:
Hey @francoism90, Just opened a PR :)

francoism90:
@melogustavo Thanks, great stuff!

PR #2294, opened 2020-03-05
Feature Increase Focus On Slider

melogustavo:
## Proposed Changes

- adds feature request in #2269 
- Implements prop biggerSliderFocus that will make the clickable area larger.
- The prop is default false, so if there is no need for making the click area bigger, it will be exactly the same as before

## Preview
Gravacao-de-Tela-2020-03-05-as-1


jtommy:
Thanks

PR #2302, opened 2020-03-09
Make BTable sortable column header text non-selectable

### LINKED ISSUE #2301: BTable sortable table header has selectable text
### Overview of the problem

**Buefy** version: [0.8.12]
**Vuejs** version: [2.6.16]
**OS/Browser**: Chrome

### Description

Double-clicking on a sortable table header selects the header text. From a standard UX perspective, text in control-based elements (buttons etc.) that can be clicked multiple times should have non-selectable text. A sortable table header is a control-based element (which is also double-clicked quite often to change sorting direction).

Demo:
buefy_table_selection_none_demo

### Steps to reproduce

Double-click a sortable table header.

### Expected behavior

Text should not have been selected.

### Actual behavior

Text was selected.

raimondlume:
Fixes #2301 

## Proposed Changes
Add `user-select: none;` to sortable table header to make it non-selectable.

### DIFF
--- a/src/components/table/TableColumn.vue
+++ b/src/components/table/TableColumn.vue
@@ -27,6 +27,10 @@ export default {
         subheading: [String, Number],
         customSort: Function,
         sticky: Boolean,
+        headerSelectable: {
+            type: Boolean,
+            default: true
+        },
         headerClass: String,
         cellClass: String,
         internal: Boolean // Used internally by Table
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -71,7 +71,8 @@
                             :class="[column.headerClass, {
                                 'is-current-sort': currentSortColumn === column,
                                 'is-sortable': column.sortable,
-                                'is-sticky': column.sticky
+                                'is-sticky': column.sticky,
+                                'is-unselectable': !column.headerSelectable
                             }]"
                             :style="{
                                 width: column.width === undefined ? null :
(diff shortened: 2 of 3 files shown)


service-paradis:
You can also use this [mixin](https://github.com/jgthms/bulma/blob/master/sass/utilities/mixins.sass#L133) for mobile and cross-browser support.

jtommy:
@raimondlume Ok but add a prop on table column component because it's a breaking change

src/components/table/TableColumn.vue
```diff
@@ -27,6 +27,10 @@ export default {
         subheading: [String, Number],
         customSort: Function,
         sticky: Boolean,
+        headerSelectable: {
+            type: Boolean,
+            default: true
```
jtommy: No It's a breaking, so you have to set to false at the moment

service-paradis: Seems OK, since the prop determine if the header is selectable. `true` is the current behavior.

raimondlume: The style binding is inverted (`'is-unselectable': !column.headerSelectable`), so if default is True, no style is added. This should not be breaking?

jtommy:
You are right, I'm sorry

raimondlume:
@jtommy Should this prop default to `false` with the next breaking update? If yes, should I make a PR for it?

jtommy:
Yes

raimondlume:
> Yes

Added the breaking change: #2306

PR #2304, opened 2020-03-09
docs: correct input event param type in datepicker

aquaminer:
Fixes #

## Proposed Changes

- change parameters in docs (api/datepicker.js) 
-
-


jtommy:
Thanks

PR #2306, opened 2020-03-09
Make tableColumn headerSelectable prop default to false

raimondlume:
## Proposed Changes
Make the headerSelectable prop of TableColumn default to false, as discussed in #2302.

### DIFF
--- a/docs/pages/components/table/api/table.js
+++ b/docs/pages/components/table/api/table.js
@@ -592,7 +592,7 @@ export default [
             },
             {
                 name: '<code>custom-sort</code>',
-                description: 'Custom sort method, works when is <code>sortable</code>',
+                description: 'Custom sort method, works when column is <code>sortable</code>',
                 type: 'Function (a: Object, b: Object, isAsc: Boolean)',
                 values: '—',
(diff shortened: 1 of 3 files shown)


service-paradis:
Thanks for the contributions!
I'm asking myself a question about this. I agree that header text should not be selectable by default when they are clickable (ex. for sorting).
I'm not sure about this default when header are not clickable though. What do you think?

jtommy:
@service-paradis you are right, looking other libs it's no a default behaviour

raimondlume:
> I'm not sure about this default when header are not clickable though. What do you think?

I agree as well, will change this when I have a moment again.

raimondlume:
How's this? This line is getting a bit long, should the conditon be moved to a computed property?

service-paradis:
Could be a good idea to add this in a computed property. Would maybe need to change the name or improve the documentation to warn user that this is affecting only sortable column.

raimondlume:
"sortableHeaderSelectable" feels a bit too verbose, so maybe let's stick with better documentation? :smile:

PR #2312, opened 2020-03-11
Prevented the regular sorting icon to appear if multi sorting is enab…

suneab:
…led on tables

Fixes #

Fixes an issue where an arrow would show up under the following condition:
When multi sort AND backend sorting is enabled
User presses column A, then column B and then pressing the delete button for column B, an arrow will show up because its the most recently sorted column.
This PR fixes that issue. 
## Proposed Changes

-
-
-


jtommy:
Thanks

PR #2330, opened 2020-03-19
Fix #595 : Add dropdownPosition prop to Autocomplete

### LINKED ISSUE #595: Autocomplete dropdown position (especially for mobile)
### Overview of the problem
Would be nice if we could specify the dropdown position (top, bottom, left, right). I think it is automatically determined right now, which works good enough most of the time, but sometimes it is positioned inconvenient or outside the screen.

For a mobile/small screens: when the dropdown is active it would be more user friendly if it was positioned underneath the autocomplete input or even better, scrollable on mobile.

**Buefy** version: 0.6.3
**Vuejs** version: 2.5.13
**OS/Browser**: Mozilla/5.0 (iPhone; CPU iPhone OS 10_3 like Mac OS X)

### Expected behavior

The dropdown positioned underneath the autocomplete input or scrollable on mobile.
 

### Actual behavior

The dropdown positioned above the autocomplete input and half of its content floats outside the screen.

schermafbeelding 2018-02-09 om 15 53 14

dauriata:
Fixes #595 

## Proposed Changes
Add a `dropdown-position` prop to Autocomplete
Value can be set to `top`, `bottom `or `auto` with a default to current behavior `auto`

Sidenote: the current implementation (and not modified with this PR) of the calculated behavior would move the dropdown up if there is no room downside, but will not move it back down if the window is resized bigger

### DIFF
--- a/src/components/autocomplete/Autocomplete.vue
+++ b/src/components/autocomplete/Autocomplete.vue
@@ -32,7 +32,8 @@
         <transition name="fade">
             <div
                 class="dropdown-menu"
-                :class="{ 'is-opened-top': !isListInViewportVertically }"
+                :class="{ 'is-opened-top': dropdownPosition==='top' ||
+                (dropdownPosition==='auto' && !isListInViewportVertically ) }"
                 v-show="isActive && (data.length > 0 || hasEmptySlot || hasHeaderSlot)"
                 ref="dropdown">
                 <div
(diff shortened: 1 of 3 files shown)


src/components/autocomplete/Autocomplete.vue
```diff
@@ -32,7 +32,8 @@
         <transition name="fade">
             <div
                 class="dropdown-menu"
-                :class="{ 'is-opened-top': !isListInViewportVertically }"
+                :class="{ 'is-opened-top': dropdownPosition==='top' ||
+                (dropdownPosition==='auto' && !isListInViewportVertically ) }"
```
service-paradis: Would be better to use a computed property here.

dauriata: sure, will update it

jtommy:
@dauriata please fix because there is an error in a unit test so I can merge it. Thanks

jtommy:
Thanks

PR #2384, opened 2020-04-08
Use $radius SASS variable for Step marker

service-paradis:
## Proposed Changes

- Use $radius SASS variable for Step marker
- Add `rounded` prop (default `true`)
- This adds a squared markers variant

captured (1)


jtommy:
Nice!

PR #2386, opened 2020-04-09
Document that table column width could use any unit, not only pixels

j-a-m-l:
Fixes #

## Proposed Changes

- Documents the improvement added on https://github.com/buefy/buefy/pull/1827


jtommy:
Thanks

PR #2396, opened 2020-04-13
Fixing Icon type to account for colors that are hyphenated. (Ex. "is-…

### LINKED ISSUE #2395: Icon type does not support hyphenated colors
### Overview of the problem

**Buefy** version: [0.8.15]
**Vuejs** version: [2.6.11]
**OS/Browser**: Windows/Chrome

### Description

When using the `type` attribute for `b-icon`, colors that are more than one word and hyphenated do not apply the corresponding "has-text-[type]" class to the element.

### Steps to reproduce

1. Create icon element with `type="is-grey-dark"`

### Expected behavior

Icon should have class "has-text-grey-dark" and be displayed with "grey-dark" color.

### Actual behavior

Icon only has first part of type in class. (Ex. class="has-text-grey")

BrauBeast:
…grey-dark")

Fixes #

Fixes #2395 

## Proposed Changes

- Use ES6 spread operator to get all parts of `splitType` except "is" prefix.
- Join resulting parts into return value for Icon type.

### DIFF
--- a/src/components/icon/Icon.vue
+++ b/src/components/icon/Icon.vue
@@ -67,7 +67,8 @@ export default {
             }
             if (splitType.length <= 1) return
 
-            return `has-text-${splitType[1]}`
+            const [, ...type] = splitType
+            return `has-text-${type.join('-')}`
         },
         newCustomSize() {
             return this.customSize || this.customSizeByPack


jtommy:
Thanks

PR #2407, opened 2020-04-16
Add rounded prop to mobile date and time pickers

jeromelebleu:
## Proposed Changes

`rounded` prop is set to the `<input>` of the shipped date and time pickers. For consistency it is also set to the native date and time `<input>` on mobile.

### DIFF
--- a/src/components/datepicker/Datepicker.vue
+++ b/src/components/datepicker/Datepicker.vue
@@ -184,6 +184,7 @@
             :size="size"
             :icon="icon"
             :icon-pack="iconPack"
+            :rounded="rounded"
             :loading="loading"
             :max="formatNative(maxDate)"
             :min="formatNative(minDate)"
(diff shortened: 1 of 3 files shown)


jtommy:
Why? There is already v-bind on input

jeromelebleu:
> Why? There is already v-bind on input

Sorry if I missed or misunderstood something in Vue, I am just learning and using it recently.

Anyway, the input is currently not rounded with the following code:
```vue
<b-datepicker placeholder="Date…" rounded />
```

... which renders on mobile:
```html
<div class="datepicker control is-filter is-empty">
  <div class="control is-clearfix">
    <input type="date" autocomplete="off" placeholder="Date…" max="" min="" class="input">
    <!----> 
    <!---->
    <!---->
  </div>
</div>
```

It works however when I add `rounded` as proposed here - and as it is already done for the `<input>` inside the dropdown [here](https://github.com/buefy/buefy/blob/v0.8.15/src/components/datepicker/Datepicker.vue#L29) despite v-bind. I thought so this was missing but if not, tell me I will open an issue instead.

PR #2417, opened 2020-04-18
DatePicker fix for Safari

### LINKED ISSUE #1957: Datepicker selected wrong date on mobile
Tested datepicker component on iOS. When selected a date with the datepicker, the wrong date is being added to the component. I've attached a screen capture of the video I took on my device.
In this image, I selected November 9, 2019, but November 8, 2019 was displayed in the datepicker component.

ScreenGrab

davidmeeker:
Update the defaultDateFormatter in the datepicker to handle dates far in the past on Safari (iOS and OS X). 

Fixes #1957

## Proposed Changes

- In DatePicker.vue:defaultDateFormatter - when constructing a new Date object for the purposes of locale formatting it, construct it against noon local time instead of midnight so that DST bugs in browser JS implementations (Safari currently) that cause the time to be off by an hour or so don't actually change the date.

### DIFF
--- a/src/components/datepicker/Datepicker.vue
+++ b/src/components/datepicker/Datepicker.vue
@@ -682,7 +682,7 @@ export default {
                 const year = parseInt(s[0], 10)
                 const month = parseInt(s[1]) - 1
                 const day = parseInt(s[2])
-                this.computedValue = new Date(year, month, day, 12)
+                this.computedValue = new Date(year, month, day)
             } else {
                 this.computedValue = null
             }
--- a/src/components/datepicker/Datepicker.vue
+++ b/src/components/datepicker/Datepicker.vue
@@ -677,9 +677,16 @@ export default {
         */
         onChangeNativePicker(event) {
             const date = event.target.value
-            this.computedValue = date ? new Date(date + 'T00:00:00') : null
+            const s = date ? date.split('-') : []
+            if (s.length === 3) {
+                const year = parseInt(s[0], 10)
+                const month = parseInt(s[1]) - 1
+                const day = parseInt(s[2])
+                this.computedValue = new Date(year, month, day, 12)
+            } else {
+                this.computedValue = null
+            }
         },
-
         updateInternalState(value) {
             const currentDate = Array.isArray(value)
                 ? (!value.length ? this.dateCreator() : value[0])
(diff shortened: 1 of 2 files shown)


jtommy:
Did you test it on all other browsers?

davidmeeker:
Tested on Firefox, Chrome and Safari under OS X. Your question prompted me to test it from my phone though, and unfortunately it doesn't look like this fixes the issue on mobile. Let me take a crack at that and I'll update the PR.

davidmeeker:
I unfortunately don't have access to an Android device to verify the native date picker on that platform. If someone can verify that this should be good to merge.

src/components/datepicker/Datepicker.vue
```diff
@@ -680,9 +677,16 @@ export default {
         */
         onChangeNativePicker(event) {
             const date = event.target.value
-            this.computedValue = date ? new Date(date + 'T00:00:00') : null
+            const s = date ? date.split('-') : []
+            if (s.length === 3) {
+                const year = parseInt(s[0], 10)
+                const month = parseInt(s[1]) - 1
+                const day = parseInt(s[2])
+                this.computedValue = new Date(year, month, day, 12)
```
jtommy, line 685: Why 12 ? It isn't a correct date. Keep in mind that value will be emitted to parent component

davidmeeker: 12 makes it noon as opposed to midnight. If folks are relying on a specific time of day within the emitted Dates, then I suppose that's problematic. The issue is that Safari is broken in such a way that you can't reliably construct a date with midnight local time when the date in question predates the most recent set of Daylight Savings rules. For instance, in the USA, we last changed DST rules in 2006, so dates constructed within range of what is now DST but was previously Standard Time will be off by an hour:

> new Date(2020,3,1).toLocaleDateString()
> "4/1/2020"
> new Date(2005,3,1).toLocaleDateString()
> "3/31/2005"
> new Date(2005,3,1).toLocaleString()
> "3/31/2005, 11:00:00 PM"

Constructing them at noon instead of midnight works around this. 

Chrome had this issue as well for many years, but it was resolved a year or two ago. 

There's a good reason most date controls emit ISO 8601 strings instead of JS Date objects.

jtommy: It's a breaking change because in this way the date object has a different time part and I don't know if it's really correct to get a not 00:00 time

davidmeeker: It doesn't, unfortunately. It's only once you do a locale format on it that you can see a problem. 

I've updated the PR to construct the dates against midnight again; given we're emitting a Date object this seems like the least worst option. Most consumers of that Date object will (hopefully) be interrogating it to get the numeric date elements (where it behaves correctly) as opposed to locale formatting it, so this should be the least surprising implementation.

PR #2419, opened 2020-04-19
Safari (OS X and iOS) fixes to DateTimePicker

### LINKED ISSUE #2418: DateTimePicker on iOS is off by UTC offset of user's timezone
On head of dev against iOS 13.4.1, the native datetimepicker exhibits the same problems that the native datepicker showed in #1957 -- the chosen datetime is being interpreted as being in UTC, but only on iOS Safari, not under Chrome / Android. 

Time mismatch in US Central (UTC-5)

The date component of the datetimepicker also exhibits the same off-by-one-day issues for spring dates prior to 1990 under OS X Safari, likely for the same reason.

davidmeeker:
Fixes #2418.

### DIFF
--- a/src/components/datetimepicker/Datetimepicker.spec.js
+++ b/src/components/datetimepicker/Datetimepicker.spec.js
@@ -20,6 +20,17 @@ describe('Datetimepicker', () => {
         expect(valueEmitted).toContainEqual(date)
     })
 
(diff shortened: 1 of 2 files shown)


davidmeeker:
If someone can check out this PR on Android that would be great.

PR #2433, opened 2020-04-24
Add sidebar content background variable

OskarKinomoto:
## Proposed Changes

- Add separate variable for sidebar content background so it can be changed separately from other elements


jtommy:
@OskarKinomoto you should use `type` prop to edit the background color

PR #2481, opened 2020-05-05
shallow copy in watcher

lunfel:
Fixes #1598 

## Proposed Changes

It fixes the mutation of vuex stores when using TagInput along with Vuex. As mentioned in this PR by ManUtopiK, it did not entirely solved the problem the first time. I had missed the watcher for cloning the array. Now I am fixing the last part.

* Make a shallow copy of the tags array when the watcher of the v-model is triggered


jtommy:
Thanks

PR #2492, opened 2020-05-09
Feature: close-type prop added into b-tag component

kaangokdemir:
## Description
This Pull Requests adds `close-type` prop into `b-tag` component to allow customize type of close icon.

## Why Buefy needs this?
Sometimes developers should want to customize close button without writing custom classes. In most of situations, `is-danger` type preferred to increase accessibility and consistency instead of using a gray one.

## Examples
Ekran Resmi 2020-05-09 13 41 09

## Documentation
Ekran Resmi 2020-05-09 13 42 53

### DIFF
--- a/src/components/tag/Tag.vue
+++ b/src/components/tag/Tag.vue
@@ -31,7 +31,7 @@
             role="button"
             :aria-label="ariaCloseLabel"
             class="delete is-small"
-            :class="[{'tag is-paddingless is-rounded' : closeType}, closeType]"
+            :class="closeType"
             :disabled="disabled"
             :tabindex="tabstop ? 0 : false"
             @click="close"
--- a/src/components/tag/Tag.vue
+++ b/src/components/tag/Tag.vue
@@ -13,7 +13,7 @@
             :aria-label="ariaCloseLabel"
             :tabindex="tabstop ? 0 : false"
             :disabled="disabled"
-            :class="[size, { 'is-rounded': rounded }]"
+            :class="[size, closeType, { 'is-rounded': rounded }]"
             @click="close"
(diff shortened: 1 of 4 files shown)


src/components/tag/Tag.vue
```diff
@@ -31,6 +31,7 @@
             role="button"
             :aria-label="ariaCloseLabel"
             class="delete is-small"
+            :class="[{'tag is-paddingless is-rounded' : closeType}, closeType]"
```
jtommy, line 34: Why?

kaangokdemir: For changing color, it should be a `.tag` element first like attached one. Otherwise Bulma's own .delete class' color would be active only https://github.com/jgthms/bulma/blob/master/css/bulma.css#L55
And `is-paddingless` + `is-rounded` combination needed for making tag rounded.

jtommy: It looks like a workaround.. i think it's better add a custom implementation in the tags.scss, for example:
```
.tag {
    .delete  {
        @each $name, $pair in $colors {
            $color: nth($pair, 1);
            &.is-#{$name}:before {
              background: $color;
            }
        }
    }
}
```

kaangokdemir: I aggree. However :before and :afters for cross' design by default in Bulma. Though we can still write custom css for `is-delete` and `delete` classes. See my commit and let me know your opinons. https://github.com/buefy/buefy/pull/2492/commits/f0673d9afcba61a17242cb45c9cda859e6d64319 @jtommy

jtommy:
Perfect!

PR #2508, opened 2020-05-13
Skeleton fixes and improvements

EmmanuelVlad:
## Proposed Changes
  
Here is my take on skeletons

- Changed `$skeleton-line-height` from `1` to `$body-line-height` so it matches the default line-height _(I don't think this variable is needed anymore)_
- Added `position` prop to align the skeleton
- Skeletons are now vertically centered

There is a small issue with `vertical-align: middle;` sometimes it'll add less than 1px to the total height of the skeleton div

### DIFF
--- a/docs/pages/components/skeleton/api/skeleton.js
+++ b/docs/pages/components/skeleton/api/skeleton.js
@@ -49,7 +49,14 @@ export default [
                 type: 'Number',
                 values: '—',
                 default: '<code>1</code>'
-            }
+						},
+            {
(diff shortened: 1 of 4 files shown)


jtommy:
@EmmanuelVlad Good catch but it's a breaking change

jtommy:
I've just added https://github.com/buefy/buefy/commit/a81b8d7d873b8d0ef737c2d1d0fe2bfacce29229 size prop (in your case you should use `is-large`).
About position prop you might submit an other PR

PR #2511, opened 2020-05-15
Keep table search filtering when data are updated

### LINKED ISSUE #2509: Search filtering is lost after updating table data
### Overview of the problem

**Buefy** version: 0.8.18
**Vuejs** version: 2.6.11
**OS/Browser**: All

### Description

After updating the data of the table, the filters are not applied on the new data.

### Steps to reproduce

1. Go to the [codesandbox demo](https://codesandbox.io/s/buefy-table-searchable-bug-7hpj5)
2. Filter the `Name` column (for example: type "J" to display only "Jesse" and "John")
3. Click on the `Refresh data` button (it just re-assigns `data`)
4. The filter is not applied anymore

### Expected behavior

I think the filters should be applied on the refreshed data.

### Actual behavior

The filters are not applied on the refreshed data.

I think I've found how to fix it, then I'll be glad to open a PR if needed.

VictorCazanave:
Fixes #2509

## Proposed Changes

- Apply filtering before sorting and pagination when `data` prop changes

### DIFF
--- a/src/components/table/Table.spec.js
+++ b/src/components/table/Table.spec.js
@@ -1,4 +1,5 @@
 import { shallowMount } from '@vue/test-utils'
+import BInput from '@components/input/Input'
 import BTable from '@components/table/Table'
 
 describe('BTable', () => {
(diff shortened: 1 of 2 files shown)


VictorCazanave:
Should I add unit tests to not decrease the code coverage?

jtommy:
Thanks!

PR #2515, opened 2020-05-16
Feature: passive-type prop for b-switch component

kaangokdemir:
## Description
This Pull Requests adds `passive-type` prop into `b-switch` component to allow customize type of close icon.

## Why Buefy needs this?
Currently Buefy is using $grey-light variable for all switchs while they are passive. Developers who wants to seperate passive type of their switch components and grey-light variables are need to write some custom CSS to make them different. This prop would mostly used by while building a 
modern Light Mode / Dark Mode switch.

## Examples
gif1

gif2

## Documentation
documentation

## Checklist

### DIFF
--- a/docs/pages/components/switch/api/switch.js
+++ b/docs/pages/components/switch/api/switch.js
@@ -12,6 +12,16 @@ export default [
                 default: '—'
             },
             {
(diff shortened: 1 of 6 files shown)


jtommy:
Nice addition!

Issue #2478, opened 2020-05-04
Close prevention for Dialog component


denis-goncharenko:
### Description

It would be good to have a possibility to prevent closing dialog modal. For example, you would like to create a confirmation modal for delete action. In this dialog, you need to enter some word like 'DELETE' to confirm that you really want to delete a record.

![image](https://user-images.githubusercontent.com/52198869/80955126-58907980-8e07-11ea-91a7-0713a84660ee.png)

### Why Buefy need this feature

At the moment on onConfirm we can't prevent dialog closing if a user has entered the word that doesn't equal to 'DELETE'.

fabiofdsantos:
Are you using `$buefy.dialog.prompt` with `canCancel` prop?

jtommy:
Duplicated https://github.com/buefy/buefy/issues/1953

Issue #1953, opened 2019-11-04
Feaure request: return promise from dialogs


Christilut:
### Description

As a heavy Buefy user, I would really like it when the confirm dialog would return a Promise.

My workaround now is simple but not very pretty:
```js
    const confirmed = await new Promise((resolve, reject) => {
      this.$buefy.dialog.confirm({
        message: `Are you sure?`,
        onCancel: () => reject(false),
        onConfirm: () => resolve(true)
      })
    })
```

Perhaps for a non-API breaking method we could do it like this (AWS SDK added promises to their library in the same way):
```js
    const confirmed = await this.$buefy.dialog.confirm({
        message: `Are you sure?`
      }).promise()
```

### Why Buefy need this feature

This is quite common nowadays. SWAL uses this method and it's very nice to work with. I ditched SWAL when I found out Buefy now has similar programmatic dialog/modal functionality.

jtommy:
@Christilut sure, it's a good idea

travisfisher:
Just ran into this issue today. I agree that returning a Promise would be a nice addition.

jtommy:
https://github.com/buefy/buefy/pull/2539

Christilut:
@jtommy I think this issue is not related to #2539 

This one is specifically about `this.$buefy.dialog` returning a Promise so that you can await it. You can achieve the same behavior with `closeOnConfirm` but that was already possible by wrapping it in a Promise as in the first post.

This issue is more for a convenience so that you can do simple stuff like 
```
const isConfirmed = await this.$buefy.dialog.confirm({
        message: `Are you sure?`
})
// isConfirmed = true if OK pressed. False if canceled by button or Escape key
```
without all the extra boilerplate code in `onConfirm`

jtommy:
@Christilut i'm working on a similar way for the next version 0.9.0 , for example:
```
const dialog = this.$buefy.dialog.confirm({
        message: `Are you sure?`,
        closeOnConfirm: false
});
dialog.result.then((value, dialog) => ...).catch(...)
```
https://github.com/buefy/buefy/blob/promise-programmatic/src/components/dialog/index.js#L16

Christilut:
I think the following will be possible with 0.9.0 then?
```ts
const isConfirmed = (await this.$buefy.dialog.confirm({
        message: `Are you sure?`,
        closeOnConfirm: false
}).result).value
```

[...]

PR #2539, opened 2020-05-23
Feature: Dialog closeOnConfirm prop

kaangokdemir:
## Description
This Pull Requests adds `closeOnConfirm` prop into `b-dialog` component to allow async requests in dialog component and adds `dialog` param as second param of `onConfirm` hook to make close request easier.

## Why Buefy needs this?
Currently only way to handle async requests is using `b-modal` component. With that change, it would be possible to close dialog component depending on async responses.

## Related Issues
https://github.com/buefy/buefy/issues/1953
https://github.com/buefy/buefy/issues/2478
And maybe many more
## Simple Usage
```javascript
  checkPassword() {
    this.$buefy.dialog.prompt({
      message: "Your password",
      inputAttrs: {
        type: "password",
        placeholder: "Your password",
        value: ""
      },
      trapFocus: true,
      closeOnConfirm: false, // This prop is a must if we want keep dialog open.
      onConfirm: (value, { close }) => {
        fetch("my/api/login")
          .then(res => {
            if (res === value) {
              this.$buefy.toast.open("Password true!");
              close(); // Instead of auto-close, construct from onConfirm hook's second dialog param.
            } else {
              this.$buefy.toast.open("Password false!");
              // Dialog remains open
            }
          })
          .catch(err => this.$buefy.toast.open("Server error.."));
      }
    });
  }
```
## Changes at `onConfirm()`

```javascript
onConfirm(value) {} // Before
onConfirm(value, dialog) {} // Now could use dialog as second parameter.
onConfirm(value, { close }) // Better to destruct.
```
## Examples
gif1

## Documentation
documentation

## Checklist
- [x] Documentation updated
- [x] New Example added

### DIFF
--- a/docs/pages/components/dialog/api/dialog.js
+++ b/docs/pages/components/dialog/api/dialog.js
@@ -91,9 +91,16 @@ export default [
             {
                 name: '<code>onConfirm</code>',
                 description: 'Callback function when the confirm button is clicked',
-                type: 'Function (value: String)',
+                type: 'Function (value: String, dialog: VueInstance)',
                 values: '—',
                 default: '—'
+            },
+            {
+                name: '<code>closeOnConfirm</code>',
+                description: 'Turning this prop into false allows to make async requests in <code>onConfirm</code> callback',
+                type: 'Boolean',
+                values: '<code>true</code>, <code>false</code>',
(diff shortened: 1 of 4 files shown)


jtommy:
Nice!

PR #2540, opened 2020-05-23
docs: add bulma-css-vars extension to docs

wtho:
## Proposed Changes

- adds [`bulma-css-vars`](wtho/bulma-css-vars) to extensions


jtommy:
Top! In the next breaking version i'll update some dependencies in order to support your work better

PR #2546, opened 2020-05-25
fixes navbar conflicts

apolokaklab:
Fixes navbar conflicts on body class when use navbar fixed with (bulma) pure css and a b-navbar without fixed.


apolokaklab:
example navbar
@jtommy I will try another solution.

Issue #2557, opened 2020-05-27
[Tabs] Improvement: Use provide/inject feature rather than direct ancestry properties 


Tofandel:
The idea is to replace the current tab registration behavior using provide inject rather than scanning the vnodes and using the `$parent` property in the tab item

This will allow to use the tab in a component and the tab item in another component

Eg:

App.vue
```html
<b-tabs v-else v-model="activeTab">
  <template v-for="group in groups">
    <Group :key="group.id" v-bind="group"/>
  </template>
</b-tabs>
```

Group.vue
```html
<b-tab-item :label="name">
  <div>Content</div>
</b-tab-item>
```

PR will follow

jtommy:
As you can see in 0.9.0 branch there is small refactoring. Your idea is correct but if you want to make a PR you have to start from that branch please:)

service-paradis:
Would be great to do this for every component using $parent (ex. Dropdown).

jtommy:
Sure, I'm adding it to table component too

Tofandel:
@jtommy I've had a look at the table component implementation and it's incomplete so it doesn't work, is it okay if I make a new PR to fix thatl?

jtommy:
I was working on it. If you want help me you might take a look to roadmap or new feature requests in the issues list. Thanks

Tofandel:
Okay then, I'll wait for the table, I have an improvement that I can PR now, basically the provide inject functionality I made a factory mixin wrapper for it, so it's super easy for any component to implement those features (Added it to dropdown as well)

I also made a super simple timeline component, it's clean enough, using bulma variables and lightweight enough to be added into buefy if you're interested (20 lines of vue and 30 lines of scss)

This is what it looks like on my dark theme
![image](https://user-images.githubusercontent.com/6115458/85327706-b5d3ad80-b4cf-11ea-8f67-782f9cb94a7b.png)
The circles are customizable b-icons

There isn't much left in the todo list for v1 but I can tackle #2214 and #2575 by removing the default value of the number input until user action on the input, I can also add a prop 'exponential' to increase the incrementation speed when button is left clicked as it's currently quite slow, it's something I did for one of my components

I can also do a few test units

Issue #1026, opened 2018-10-11
tabs nav is not active by v-model


pakkaparn:
### Overview of the problem

**Buefy** version: [0.7.0]
**Vuejs** version: [2.5.16]
**OS/Browser**:
MacOs/Chrome
### Description

tab nav is not active by v-model

![alt text](https://i.imgur.com/fsYvB3G.png)

```
<b-tabs type="is-boxed" class="column" v-model="tabIndex">
```
```
computed: {
    tabIndex: {
      get() {
        return this.$route.query.tab;
      },
      set(index) {
        this.$router.replace({ query: { tab: index } });
      },
    },
}
```

jtommy:
@pakkaparn The component accepts string value but it's no the correct way, we'll fix it.
Try to use `parseInt `function in the `get `of `tabIndex`

PR #2558, opened 2020-05-28
Change the tabs logic to provide/inject and add support for ids in v-model

Tofandel:
Fixes ##2498 #2557 #1026

## Proposed Changes
The logic is now simplified by Vue provide/inject
The internal state has been modified to work with ids, the external state can work with indexes or ids as such keeping backward compatibility

If a number is provided the input will always emit a number, if an id is provided it will always emit a string id

Only breaking change:
The change event was removed, as it's not useful and was acting as a duplicate of input

Thoroughly tested with unit test and doc included


jtommy:
Why did you add Id prop ? There is already value prop on tabitem to use a string value instead of an index.

Tofandel:
Mm, guess I completely missed it, I'll just change id to value

Tofandel:
Quick question, is it okay if I also include a version change of node-sass in package.json
It doesn't compile with the latest node with the current version (https://github.com/nodejs/node-gyp/issues/1960) and I have to update it every time

jtommy:
About node-sass I think to remove it and use dart-sass

Tofandel:
That's a good idea, I'm using it for projects using buefy and it does solve some problems at the price of some compilation time, there is also quite a few other libraries that need an update (eg @vue/test-utils)
So I won't include it here, PR is done

PR #2580, opened 2020-06-06
closeType prop added into BTagInput

kaangokdemir:
## Description
With #2492, close icons in `tag` components were became customizable. This Pull Requests adds `close-type` prop into `b-taginput` component to allow customize type of close icon at Tags in b-taginput.

## Why Buefy needs this?
Sometimes developers should want to customize close button without writing custom classes. In most of situations, `is-danger` type preferred to increase accessibility and consistency instead of using a gray one.

## Examples
examples

## Documentation
documentation

## Checklist
- [x] Documentation updated
- [x] New Example added

### DIFF
--- a/docs/pages/components/taginput/api/taginput.js
+++ b/docs/pages/components/taginput/api/taginput.js
@@ -40,6 +40,16 @@ export default [
                 default: '<code>is-light</code>'
             },
             {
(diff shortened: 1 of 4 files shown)


jtommy:
Thanks

PR #2590, opened 2020-06-09
Dropdown close on window scroll feature

cain:
Hey! 

This feature closes an active dropdown if the window is scrolled.

We needed this functionality for our project, so I thought I might as well propose it here also if you think it would be helpful for others.

If you don't think this would be useful, I'm happy to close this PR. Otherwise, I'm happy to complete the PR with docs and tests.

Thanks :)

### DIFF
--- a/src/components/dropdown/Dropdown.vue
+++ b/src/components/dropdown/Dropdown.vue
@@ -116,7 +116,8 @@ export default {
         },
         expanded: Boolean,
         appendToBody: Boolean,
-        appendToBodyCopyParent: Boolean
+        appendToBodyCopyParent: Boolean,
+        closeOnWindowScroll: Boolean
     },
     data() {
         return {
@@ -174,6 +175,11 @@ export default {
             if (this.appendToBody) {
                 this.$nextTick(() => {
                     this.updateAppendToBody()
+                })
+            }
+            if (this.closeOnWindowScroll && value) {
+                this.$nextTick(() => {
+                    this.activeWindowScrollClose()
                 })
             }
         }
@@ -336,6 +342,18 @@ export default {
                     zIndex: '99'
                 }
             }
+        },
+
+        activeWindowScrollClose() {
+            // Use a local state to ensure function isnt called more then once
+            let triggered = false
+            const func = () => {
+                if (triggered) return
+                if (this.$refs.optionsDropdown) this.toggle()
+                triggered = true
+                window.removeEventListener('scroll', func)
+            }
+            window.addEventListener('scroll', func)
         }
     },
     mounted() {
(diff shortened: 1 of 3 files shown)


jtommy:
Sure! It's ok for me

service-paradis:
Good idea. Instead of a new prop, it could also be a `scroll` value in `can-close` prop?

service-paradis:
Good idea. Instead of a new prop, it could also be a `scroll` value in `can-close` prop?

jtommy:
@service-paradis good

cain:
@service-paradis will do!

jtommy:
@cain i've just converted it to draft so when it's ready you change it to "open", thanks

src/components/dropdown/Dropdown.vue
```diff
@@ -336,6 +342,18 @@ export default {
                     zIndex: '99'
                 }
             }
+        },
+
+        activeWindowScrollClose() {
```
Tofandel, line 347: The `const func` should be added as a method called `onScroll` instead
Because the event listener needs to be cleaned in the `beforeDestroy` hook
(Or you may pollute the windows with event listeners if the dropdown is destroyed while it was opened)

cain: The problem I see with `onScroll` is that window can only be assigned one value? I'm assuming this is going to cause issues if anyone else is wanting to use `onScroll` or accidentally doesn't realize it's being used.

"Only one onscroll handler can be assigned to an object at a time." https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onscroll

I'm trying to find a clean solution to remove the addEventListener, not sure yet.

cain: Also thanks for the help, appreciate it

Tofandel: Sorry if my comment wasn't clear, I wasn't talking about the windows method, but the component methods and lifecycle hooks instead of having a const func, you need to reference `window.removeEventListener('scroll', this.onScroll)`
and `window.addEventListener('scroll', this.onScroll)`

And you will add a hook to the component
```
beforeDestroy() {
  window.removeEventListener('scroll', this.onScroll)
}
```
So that it cleans the event listener correctly

Tofandel:
For your tests to work, you need to make your test function async and add
`await wrapper.vm.nextTick()` before and after the scroll action

afwn90cj93201nixr2e1re:
And why he is using global func for only one prop? What about situation when we gonna add new features for scrolling?

You don't remember when i suggested make auto position adj. on scrolling and u declined it? Lol...

src/components/dropdown/Dropdown.vue
```diff
@@ -336,6 +342,18 @@ export default {
...
+            // Use a local state to ensure function isnt called more then once
+            let triggered = false
+            const func = () => {
+                if (triggered) return
+                if (this.$refs.optionsDropdown) this.toggle()
```
service-paradis, line 352: If maxHeight is used, we should not close the dropdown if we're scrolling into it

PR #2603, opened 2020-06-15
Handle events for customElement components

vvkk77:
Hi,

This PR adds support of handling auto-close feature in components that are wrapped in a HTML5 custom element created via @vue/web-component-wrapper. 

**Problem:** If we use components like dropdown and autocomplete, the handling of auto-close on clicking outside the element currently depends on `event.target`. In the case of customElement the `event.target` returns the customElement instead of the actual target element, due to which the component closes/toggles on clicking even on the element itself, which seems like a bug in the case of customElement.

**Solution:** [`Event.composePath()`](https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath) provide the event's path list. We can check if the Vue instance is wrapper in a customElement. And get the composedPath of the event and treat the first element as the target. This would solve the above problem.

### DIFF
--- a/src/components/autocomplete/Autocomplete.vue
+++ b/src/components/autocomplete/Autocomplete.vue
@@ -80,7 +80,7 @@
 </template>
 
 <script>
-import { getValueByPath, removeElement, createAbsoluteElement } from '../../utils/helpers'
+import { getValueByPath, removeElement, createAbsoluteElement, isCustomElement } from '../../utils/helpers'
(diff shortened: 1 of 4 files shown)


Dharmendra205111014:
Really Needed this as Every time we wrap buefy dropdown inside a custom web component created by vue cli, The exact same issue happen to be there. 

Thanks for the PR

vvkk77:
@jtommy any thought/comment on this?

jtommy:
I think it might be a good solution

PR #2605, opened 2020-06-16
Fixing carousel active items

### LINKED ISSUE #2604: Carousel not always setting the correct carouselItem index when using multiple instance on the same page
### Overview of the problem

**Buefy** version: <=[0.8.20]
**Vuejs** version: <=[2.6.11]
**OS/Browser**: Ubuntu Linux / Chrome 83.0.4103.61  (64-bits)

### Description

Carousel not always setting the correct carouselItem index when using multiple instance on the same page especially with different amount of items per carousel. When autoplay is enabled and you have let's say 2 carousels. Carousel A has 3 items, Carousel B has 2 items. Displaying them both works fine and auto play is changing the items for each carousel. 

Before we set a filter to only show Carousel B we have the following scenario: Carousel A is displaying item on index 2 and Carousel B is displaying item on index 0. When the filtering is performed, Carousel B is loaded on top of Carousel A (VueJS is reusing the objects or something like that) the index of the carouselItem is still on index 2, but Carousel B only has 2 items so the max index is 1.

This happens when state is not maintained by VueJS. https://vuejs.org/v2/guide/list.html#Maintaining-State

### Steps to reproduce

1. Add at least 2 Carousels on a single page and make sure that they have different amount of items to display. 
2. Forget to set the :key (to maintain state) on the element your filtering on.
2. Filter on one of the carousels with the smallest amount of items
3. You will now see that the filtered carousel will not display the correct image.

### Expected behavior

Checking if the index at carouselItems exists or fallback to the first index 0. When index of activeItem is not defined, fallback to first index and display that item.

### Actual behavior

Not showing the correct image and throwing undefined errors when autoplay is enabled.

### Screenshots

Screenshot 1: correct working carousels
Screenshot from 2020-06-16 08-57-48

Screenshot 2: Filtered on carousels and incorrect indexes sellected.
Screenshot from 2020-06-16 08-58-16

Screenshot 3: console errors after filtering with autoplay on
Screenshot from 2020-06-16 08-58-39

ivdmeer:
Fixes #2604

## Proposed Changes

- falling back to first index in carouselItems to make as active item or at change event when autoplay is active.

### DIFF
--- a/src/components/carousel/Carousel.vue
+++ b/src/components/carousel/Carousel.vue
@@ -228,8 +228,8 @@ export default {
         carouselItems() {
             if (this.activeItem < this.carouselItems.length) {
                 this.carouselItems[this.activeItem].isActive = true
-            } else if(this.carouselItems[0] !== undefined) {
-                this.carouselItems[0].isActive = true;
+            } else if (this.carouselItems[0] !== undefined) {
+                this.carouselItems[0].isActive = true
             }
         },
         /**
@@ -239,6 +239,7 @@ export default {
             status ? this.startTimer() : this.pauseTimer()
         }
     },
+
     methods: {
         startTimer() {
             if (!this.autoplay || this.timer) return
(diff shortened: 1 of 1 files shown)


jtommy:
Please fix lint error

ivdmeer:
I will fix the lint errors

ivdmeer:
Thanks for reviewing and accepting! Cheers.

Issue #2591, opened 2020-06-09
Carousel does not continue to slide even pause-hover is false


tungtex-it:
### Overview of the problem

**Buefy** version: [0.8.20]
**Vuejs** version: [2.6.11]
**OS/Browser**: Windows 10 1909/Firefox 77.0.1

### Description

I was following the documentation and created a carousel with autoplay = true, it was rolling smoothly except even pause-hover was set to false, the carousel stopped when mouse moved over it.

I wonder if I misunderstood anything in the guideline. Please refer to the example originally provided in the official guideline: https://codepen.io/pen?&editable=true&editors=101=https%3A%2F%2Fbuefy.org%2Fdocumentation%2Fcarousel%2F

Switch on Autoplay and leave Pause on Hover off.

### Expected behavior

Carousel continues rolling even mouse moves over it.

### Actual behavior

Carousel pauses when mouse moves over it.

Tofandel:
I can confirm there is a bug, it always pauses on hover even if set to false

Tofandel:
This is a simple fix but only made for the next major, so if it's urgent I could make a PR for a minor (it's only one function to rename in the template)

jtommy:
You can make a breaking change if you want

jtommy:
Close https://github.com/buefy/buefy/pull/2619

Issue #2196, opened 2020-02-03
Carousel v-model is not in sync when using arrow navigation buttons


lunfel:
### Overview of the problem

**Buefy** version: [0.8.10] (0.8.9 is also affected)
**Vuejs** version: [2.5.18]
**OS/Browser**: Linux Ubuntu

### Description

The v-model of Carousel component is not reactive on slide changes through previous and next arrow navigation buttons.

### Steps to reproduce

1. Create a simple b-carousel component with a v-model using this code (code from simple example from docs with added v-model):
```
<template>
    <b-carousel v-model="selected">
        <b-carousel-item v-for="(carousel, i) in carousels" :key="i">
            <section :class="`hero is-medium is-${carousel.color}`">
                <div class="hero-body has-text-centered">
                    <h1 class="title">{{carousel.text}}</h1>
                </div>
            </section>
        </b-carousel-item>
    </b-carousel>
    <span>{{ selected }}</span>
</template>

<script>
export default {
    data(){
        return {
            selected: 0,
            carousels: [
                { text: 'Slide 1', color: 'primary' },
                { text: 'Slide 2', color: 'info' },
                { text: 'Slide 3', color: 'success' },
                { text: 'Slide 4', color: 'warning' },
                { text: 'Slide 5', color: 'danger' }
            ]
        }
    }
}
</script>
```
2. Click on navigation buttons
3. The value of selected will stay at 0
4. (optional) When using indicators, the value will update

### Expected behavior

I would expect my v-model to be in sync with the actual selected slide regardless of the method I choose to change my slide

### Actual behavior

My v-model is only in sync when using indicators

### Discussion / Solution

In the docs, it states those two events:

Name | Description | Parameters
-- | -- | --
change | Triggers when carousel item value is changed | value: Boolean, value: String
input | Triggers when indicator-mode value is changed | value: Boolean, value: String

I feel like `change` should be `input` in fact. That way, v-model would always stay in sync with the component.

This is the api I would suggest, but it is a breaking change for current users

Name | Description | Parameters
-- | -- | --
input | Triggers when carousel item value is changed | value: Boolean, value: String
indicator-change | Triggers when indicator-mode value is changed | value: Boolean, value: String

### Workaround

The current workaround (not tested yet) would be to update the v-model manually through `@change` event.

```
// Untested
<b-carousel @change="value => selected = value"></b-carousel>
```

apolokaklab:
[...]

PR #2619, opened 2020-06-23
0.9.0 Improvements and bugfixes

Tofandel:
This PR moves the provide inject features to factory mixins and adds this feature to the dropdown and carousel as well

Fixed: #2591 #2196

Refactored the carousel list to respect vue guidelines (use computed properties rather than events and methods) and fixed some bugs + added some test units, I removed the config prop because it's redundant and the v-bind directive already does what this was trying to do very badly

Improved the doc on card modal to vue best practice using `$emit` rather than `$parent` and added the close icon to the card modal rather than in the overlay (like shown on the bulma doc)

Added a prop headerClass to Tabs and Steps the same way as it exists on Table

Added a prop exponential to NumberInput + documentation + examples + test units + improved coverage (There was actually a bug in the onclick handler that got uncovered)
Peek 2020-06-23 16-53

### DIFF
--- a/docs/pages/components/carousel/api/carousel.js
+++ b/docs/pages/components/carousel/api/carousel.js
@@ -369,7 +369,7 @@ export default [
             {
                 name: '<code>item</code>',
                 description: 'Custom item',
-                props: '<code>list: Array</code>, <code>index: Number</code>, <code>active: Number</code>'
+                props: '<code>...item</code>, <code>index: Number</code>, <code>active: Number</code>, <code>scroll: Number</code>'
             }
         ],
         events: [
--- a/docs/pages/components/carousel/api/carousel.js
+++ b/docs/pages/components/carousel/api/carousel.js
@@ -46,7 +46,7 @@ export default [
             },
             {
                 name: '<code>pause-info</code>',
-                description: 'Show infomation pause when <code>autoplay</code> and <code>pause-hover</code>',
+                description: 'Show information pause when <code>autoplay</code> and <code>pause-hover</code>',
                 type: 'Boolean',
                 values: '—',
                 default: '<code>true</code>'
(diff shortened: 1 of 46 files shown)


docs/pages/components/carousel/api/carousel.js
```diff
@@ -261,7 +261,7 @@ export default [
             },
             {
                 name: '<code>config</code>',
-                description: 'An object to pass all configs',
+                description: 'This property was removed, if you were using it, simply rename <code>:config</code> to <code>v-bind</code>',
```
service-paradis, line 264: Since 0.9.0 will be considered a breaking release for Buefy, I think this part can be removed from the doc. Every breaking changes will need to be mentioned in the Changelog.

Tofandel: Yep, there is an error thrown by the component if the prop is provided anyways, I'll remove it

src/components/carousel/CarouselList.vue
```diff
@@ -114,160 +119,180 @@ export default {
...
-        this.$nextTick(() => {
-            this.update()
-        })
+        this.observer = new ResizeObserver(this.refresh)
+        this.observer.observe(this.$el)
+        window.addEventListener('resize', this.resized)
+        this.resized()
+        if (this.$attrs.config) {
+            throw new Error('The config prop was removed, you need to use v-bind instead')
+        }
```
service-paradis, line 290: I think you can remove this

Tofandel: This is important when releasing breaking changes, not everybody reads the changelog and if you don't trigger an error people may not notice. It should be removed on the following major

service-paradis:
Thanks a lot for your work on those improvements! Really like the Inject/Provide mixins!

Once @jtommy approve your works, it could be good to merge this into the current 0.9.0 branch. I'm working on this branch and it will be difficult to merge it all if we keep adding things on this one. If you have other improvements to make, it could be in another PR. What do you think?

Tofandel:
Yes I completely agree, it's getting big and hard to review. But having multiple PR's open is a bit worse if there is conflicts between the two PR's.. So I'll just wait for this to be merged and then start a new PR from the same branch, that will clean the change history

I still have the menu to refactor to the provide inject mixin, anything I missed?

In the next PR I'll attack the scss tree shaking, I just got the best idea for it

jtommy:
@Tofandel It's a huge and awesome work. I checked all and i think it's ready to be merged, what do you think ?

Tofandel:
@jtommy Now it's ready to be merged ;) I'll do the menu in a following PR as it's another complex item
For SCSS it's possible to make the imports in each vue file we also need to import once dependencies (either dart-sass or defining a variable in each file to assert it's not output twice)

Then in buefy.scss we import the scss of the vue component rather than the actual scss

Other solution is to declare only a variable in the scss of each component and work from that in buefy.scss, I do need to do some tests regarding import orders and see what works and what doesn't

We should also make tree shaking available with individual vue components, as some components don't import their dependencies

PR #2630, opened 2020-06-27
Tabs (is-vertical) - Allowing the wrapping of content when there is n…

bawa93:
This commit fixes the issue of vertical Tabs, For now, when use vertical tabs feature, content is not properly visible due to lack of space for both navigation and content, here is before and after screenshot.

Before: https://nimb.ws/SIiWwD

After: https://nimb.ws/mxbDWw


jtommy:
Thanks

PR #2638, opened 2020-06-30
Taginput : Added fix for 2608 bug.

### LINKED ISSUE #2608: TagInput value wrong when allow-duplicates is disabled
### Overview of the problem

**Buefy** version: [0.8.20]
**Vuejs** version: [2.6.11]
**OS/Browser**: Windows 10/Chrome

### Description

When `allow-duplicates` is disabled (default), and autocomplete is enabled, entering a value twice will remove the tag from UI. 
However the variable bound with v-model is not updated. 

### Steps to reproduce

1. Navigate to https://buefy.org/documentation/taginput#autocomplete
2. Ensure `Allow new tags` is disabled
3. Enter `George`, press enter button
4. Repeat step 3
5. Variable bound to v-model still retains the data for George, but the tag is not present.

### Expected behavior

The variable bound in v-model to reflect what is indicated in the UI.

### Actual behavior

The data for George only disappears if more tags are added or removed - but the state is inconsistent unless action is taken.

nithesh247:
hey,

Fixes #2608 
On adding same tag when allow-duplicates is disabled the tag is getting cleared .

## Proposed Changes
Have removed the code for allowDuplicates check.
we are already checking the allowDuplicates in the next line.

Let me know if any changes has to be done. cheers :)

thanks,

### DIFF
--- a/src/components/taginput/Taginput.vue
+++ b/src/components/taginput/Taginput.vue
@@ -271,14 +271,6 @@ export default {
                         return
                     }
                 }
-                // Remove the tag input previously added (if not allowDuplicates).
-                if (!this.allowDuplicates) {
-                    const index = this.tags.indexOf(tagToAdd)
-                    if (index >= 0) {
-                        this.tags.splice(index, 1)
-                        return
-                    }
-                }
                 // Add the tag input if it is not blank
                 // or previously added (if not allowDuplicates).
                 const add = !this.allowDuplicates ? this.tags.indexOf(tagToAdd) === -1 : true


jtommy:
I don't understand.. why are you removing it?

nithesh247:
> I don't understand.. why are you removing it?

when you add same tag second time the same tag that was added at first time is getting removed from the view.
since in the code we are just removing if the tag is already exits in the array.so i have removed it.

nithesh247:
Hi @jtommy 

Should i close the PR?

jtommy:
Ok but I'll add a prop to restore the feature

PR #2661, opened 2020-07-09
Implements debounce for Table component

### LINKED ISSUE #2636: [Table] Add a delay between text input and the filter event for big tables
I have a pretty large table (using a large per-page prop), so triggering a search is sluggish, maybe there can be a prop to set a delay to the filter-event until I can enter the full text I want to search.

EDIT: Typo, the --> there.

Sc0ra:
Fixes #2636

Implemented the underscore debounce function and use it the table component to allow debouncing filtering.


jtommy:
Thanks

PR #2671, opened 2020-07-12
[fix] Registered iconSize prop for input to pass onto icon

riqwan:
##  Proposed Changes

Currently there is a method called `iconSize` that is being passed to the `b-icon` component but its not being initialized anywhere or being accepted as a prop.

This commit allows the `input` component to receive iconSize as a prop


jtommy:
At the moment the size follows input size, like button and other component. I don't think it's a good addition

Issue #2686, opened 2020-07-17
switch toggles on interaction even if bound value has not been changed


kmohrf:
### Overview of the problem

**Buefy** version: 0.8.20
**Vuejs** version: 2.6.11
**OS/Browser**: Chrome/Chromium

### Description

When toggling a switch that is bound to a computed property with a custom setter via v-model, the switch changes its visual state even though the new switch state might not have been written back to the source. 

In my case I wanted the user to confirm the decision to turn the switch off before the change is applied. I solved this by creating a computed property with a custom setter that updates the value of the source property once the user has confirmed their decision in a dialog. You can see this in action [here](https://codepen.io/kmohrf/pen/JjGeGEW). I’m not entirely sure, if the setter is a good place for this kind of interaction with the user, but the behavior seems wrong either way.

The root cause seems to be that the switch toggles its local state and then emits the input event. As bound values are only picked up on change, but the bound value never changes in the process of the input event, the switch and its bound value are no longer in sync. IMHO it should suffice to delete [the line that toggles the local state](https://github.com/buefy/buefy/blob/dev/src/components/switch/Switch.vue#L75).

### Steps to reproduce

1. Add a switch to your page
2. Bind it to a computed field with a custom setter
3. Let the setter decide when to write back the new value

### Expected behavior

The switch should only toggle once the new value has been written back to the source of the computed property.

### Actual behavior

The switch toggles right after a click and no longer matches the state that is bound via v-model.

jtommy:
Use a computed prop for your scope is not a good idea and i think it's better use a watcher, for example https://codepen.io/jtommy/pen/QWNLmZz

kmohrf:
Uhm… your solution has the weird side-effect of toggling the switch component before the dialog is answered. I’d say that this might be confusing to the user as the switch is then turned off even though they haven’t answered the question yet.

I’m also not really sure *why* it’s not a good idea to handle it with a computed property/setter. The vuex docs explicitly mention [computed setters for store commits](https://vuex.vuejs.org/guide/forms.html#two-way-computed-property). That seems like a similar use-case although store commits are admittedly synchronous whereas my example utilizes an async mechanism for storing data.

I really don’t see any advantage of the current state-management in the switch component. And even though it doesn’t technically break the one-way-data-flow paradigm of vue, because it doesn’t change the parents state, it’s obviously prone to get detached from the parents state as shown in my example.

I kindly ask you to reconsider closing this issue without a fix in Buefy.

[...]

PR #2687, opened 2020-07-17
make local switch state entirely dependent on bound value

kmohrf:
This fixes the switch behavior in cases where the state emitted by the switch is not applied by the parent component. Until now the switch would toggle even if the bound value never changed after an input event from it.

Feel free to change any of this (including the commit description… I’m kind of struggling with the words to adequately describe the problem and solution today...).

Fixes buefy#2686

### DIFF
--- a/src/components/switch/Switch.spec.js
+++ b/src/components/switch/Switch.spec.js
@@ -29,10 +29,11 @@ describe('BSwitch', () => {
     })
 
     it('emit input event when computedValue is set', () => {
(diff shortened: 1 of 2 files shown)


jtommy:
I'm checking it because I don't like the solution but in general there is an issue updating v-model

kmohrf:
Yeah feel free to close this or implement another solution. As I said in #2686 I struggled a little to adequately describe the problem and thought that a PR would help to illustrate it.

PR #2688, opened 2020-07-18
docs: change taginput confirm-key-codes default value

felipelincoln:
`TAB` key (code 9) is actually not in the default `confirm-key-codes` in Taginput component:
https://github.com/buefy/buefy/blob/dev/src/components/taginput/Taginput.vue#L143


jtommy:
In the next version there aren't key-codes (breaking)

PR #2701, opened 2020-07-22
Adds Snackbar.close() example to docs.

chrisleyva:
The documentation did not make clear that snackbar.open() returned
a reference to the Snackbar, which can then be closed manually at
a later time.

## Proposed Changes

- Adds to the Snackbar docs an example of calling `.close()` on a previously opened Snackbar.

---
This change is [Reviewable](https://reviewable.io/reviews/buefy/buefy/2701)


jtommy:
It's quite confused example, better add close method to the api docs

Issue #2640, opened 2020-07-02
Remove newlines where necessary to avoid leading/trailing whitespaces


daniandl:
### Overview of the problem

**Buefy** version: 0.8.20
**Vuejs** version: 2.6.11
**OS/Browser**: Safari 13.1

### Description
Table cells and input message/hint fields (probably also more) have a leading and trailing space on **Safari**.
This causes **inconsistent** layout issues like in the screenshots below. This does not happen on **Chrome, Firefox & Edge.**

**Example from the docs Table page:**
![image](https://user-images.githubusercontent.com/2960785/86344041-11353680-bc5a-11ea-93a7-a38088f40910.png)
_The header cells and body cells are not aligned._

**Example from docs Input page:**
![image](https://user-images.githubusercontent.com/2960785/86344208-49d51000-bc5a-11ea-99c3-34133a9e8017.png)
_The message is not aligned with the input on the left._ 

This is due to lines such as:
https://github.com/buefy/buefy/blob/dev/src/components/table/Table.vue#L284

(From my quick research, this behavior is intentional as defined in the HTML 2.0 spec.)

#### Suggested solution
Wrap code like this in a `span` tag or move them to a single line of code to avoid newlines being converted to spaces.
Fixing it per component would be less destructive/bug-prone than changing compiler options, like in #1944 

### Steps to reproduce

- can be seen on docs page

### Expected behavior

There should not be a leading/trailing space.

### Actual behavior

There is a leading/trailing space.

PR #1988, opened 2019-11-17
Remove extra spacing inside pagination

afwn90cj93201nixr2e1re:
Fixes #

## Proposed Changes

-
-
-

PR #1987, opened 2019-11-17
Fix extra spaces inside table :/

afwn90cj93201nixr2e1re:
Fixes #

## Proposed Changes

-
-
-

PR #1944, opened 2019-11-02
Fix single spaces in templates.

afwn90cj93201nixr2e1re:
Fixes #1927, #1703

## Proposed Changes

-
-
-

PR #2710, opened 2020-07-26
0.9.0

jtommy:


src/components/table/Table.vue
```diff
@@ -265,25 +241,21 @@
...
-                                <BTableColumn
-                                    v-for="column in newColumns"
-                                    v-bind="column"
-                                    :key="column.customKey || column.label"
-                                    internal>
-                                    <span
-                                        v-if="column.renderHtml"
-                                        v-html="getValueByPath(row, column.field)"
+                            <template v-for="(column, index) in visibleColumns">
+
```
afwn90cj93201nixr2e1re, line 245: If you already doing this, remove extra lines everywhere, they're useless there.

src/components/carousel/Carousel.vue
```diff
@@ -2,39 +2,39 @@
...
         @mouseleave="startTimer">
         <progress
             v-if="progress"
             class="progress"
             :class="progressType"
-            :value="activeItem"
-            :max="carouselItems.length - 1">
-            {{ carouselItems.length - 1 }}
+            :value="activeChild"
+            :max="childItems.length - 1">
```
afwn90cj93201nixr2e1re, line 12: It's better to move it inside computed stuff instead raw templates.

src/components/modal/Modal.vue
```diff
@@ -24,11 +24,14 @@
...
                     :is="component"
-                    @close="close"/>
-                <div
-                    v-else-if="content"
-                    v-html="content"/>
-                <slot v-else/>
+                    :can-cancel="canCancel"
+                    @close="close"
+                />
+                <div v-else-if="content"> {{ content }} </div>
```
afwn90cj93201nixr2e1re, line 30: Then we can't use html inside it? Why? :/ Maybe then it's should be `{{{ }}}`?

src/components/notification/Notification.vue
```diff
@@ -21,11 +21,14 @@
...
-                        class="text"
-                        v-if="message"
-                        v-html="message" />
-                    <slot v-else/>
+                    <p class="text">
+                        <template v-if="$slots.default">
+                            <slot />
+                        </template>
+                        <template v-else>
+                            {{ message }}
```
afwn90cj93201nixr2e1re, line 29: Same here, but as i can see now you are using slot's instead html? Ok, whatever.

src/components/numberinput/Numberinput.vue
```diff
@@ -193,27 +195,27 @@ export default {
         },
         onControlClick(event, inc) {
             // IE 11 -> filter click event
-            if (event.detail !== 0 || event.type === 'click') return
+            if (event.detail !== 0 || event.type !== 'click') return
```
afwn90cj93201nixr2e1re, line 198: You are using `{key, blabla}` rn, so, maybe there's also should be `{detail, type}`?

src/components/numberinput/Numberinput.vue
```diff
@@ -193,27 +195,27 @@ export default {
...
             if (inc) this.increment()
             else this.decrement()
         },
+        longPressTick(inc) {
+            if (inc) this.increment()
+            else this.decrement()
+
+            this._$intervalRef = setTimeout(() => {
+                this.longPressTick(inc)
+            }, this.exponential ? (250 / (this.exponential * this.timesPressed++)) : 250)
```
afwn90cj93201nixr2e1re, line 208: Hardcoded, make it configurable (250).

src/utils/helpers.js
```diff
@@ -7,12 +7,44 @@ function signPoly(value) {
...
 export const sign = Math.sign || signPoly
 
+/**
+ * Checks if the flag is set
+ * @param val
+ * @param flag
+ * @returns {boolean}
+ */
+function hasFlag(val, flag) {
+    return (val & flag) === flag
```
afwn90cj93201nixr2e1re, line 17: xD, once serviceparadice said that he don't give a damn about speed/performance. Also it's should be `==`, not `===`, there's no ways to get diff type than integer.

afwn90cj93201nixr2e1re:
https://github.com/buefy/buefy/pull/1988#issuecomment-554772143
https://github.com/buefy/buefy/pull/1987
https://github.com/buefy/buefy/pull/1981 - it's important too... 
https://github.com/buefy/buefy/issues/2640

(https://github.com/buefy/buefy/pull/1944)

src/utils/helpers.js
service-paradis: The fact that the type wont be different is not a really good reason to use `==` instead of `===`. Do you have any source to support this? For me, the only good reason to use `==` is if you absolutely need type coercion. In every other cases, we should prefer strict equality `===`. Talking about performance here is also weird since the difference between the two will always be negligible and `===` will be faster in most cases.

Here are some references about this:
https://codeahoy.com/javascript/2019/10/12/==-vs-===-in-javascript/
https://stackoverflow.com/a/359509/2774496
https://stackoverflow.com/a/8045526/2774496

service-paradis:
There are still some conflicts that would need to be addressed.

src/utils/helpers.js
afwn90cj93201nixr2e1re: Do you know bit operations return type? (Yes, just check typeof) It's everytime int, you don't really need === there, but whatever. Instead providing js links, which are describe js, check c bit operations which are implemented inside js engines there's no way to get anything else than integer, js doesn't work with 64, 128 bits numbers on bitwise operations.

service-paradis: I know what a bitwise operation is and I know its return type. There are still no reasons to not use strict equality.

afwn90cj93201nixr2e1re: What the reason use strict equality if there's no way to get anything instead number?)
let a = 16;
b = 2;
alert(typeof (a&b));

Idk why u sent these links.

afwn90cj93201nixr2e1re:
Check out these issues too. Coz we waited more than one year for those breaking changes.

afwn90cj93201nixr2e1re:
Ok nothing was done? Seems like you don;t give a damn about stability and issues resolivng on stable/brekable releases.

Tofandel:
@afwn90cj93201nixr2e1re Instead of complaining and criticizing on the huge work that was done for free, you know you can make pull requests to fix those issues as well? None of the contributors currently have the capacity to fix them and 0.9 was a long awaited release so unless you wanted to wait another year for it to come out, well it was released with still plenty of issues resolved and new features, and it doesn't mean there can't be a new major fixing those and new issues in a few months

src/components/numberinput/Numberinput.vue
Tofandel: The exponential is configurable making the timer speed effectively configurable more than well enough in term of changing speed

src/utils/helpers.js
Tofandel: The reason is it's safer coding for a lot of other operations so linters and code formatting tools will always force/highly recommend you to use them. Btw you can't assume the type of flag, it should be a number and an actual flag but it's not type checked so if you passed `true` in there well you'd be surprised by what you'd get
```
(1 & true) === true
false
(1 & true) == true
true
```

Tofandel:
Lol @afwn90cj93201nixr2e1re blocked me so I can't comment on his PR and issues, great mentality that goes with a great personnality
Is it possible to ban a user from a repo? Because I believe this kind of behavior should not be tolerated

PR #2719, opened 2020-07-30
Update navbar docs: wording

rubjo:


rubjo:
Up to you, but it does correct a couple of typos. 🙂

PR #2730, opened 2020-08-06
Fix a console error since we cannot destructure preventDefault

service-paradis:
## Proposed Changes

- remove `preventDefault` destructuring

It was causing an error in the console that prevented adding a tag in the taginput component.
> [Vue warn]: Error in v-on handler: "TypeError: Illegal invocation

The cause is explain here:
https://stackoverflow.com/a/49616808/2774496


jtommy:
Thanks

PR #2734, opened 2020-08-07
Datepicker: Be sure $refs.input exists onChange

service-paradis:
Fixes Unit tests and prevent errors if user are using custom trigger for Datepicker


jtommy:
Thanks

PR #2735, opened 2020-08-08
Fixed table row comparison

hikaruworld:
## Proposed Changes

Since the comparison is done with `row === selected`, even if the key is the same, the selected state will be lost as different data.

Changed to perform key comparison instead of object comparison if customRowKey is set.

### Example
https://codepen.io/hikaruworld/pen/BaKyeaM


jtommy:
Thanks

PR #2743, opened 2020-08-12
Small documentation update

jacenre:
My first contribution to Buefy!

The default value of destroy-on-hide is [true](https://github.com/buefy/buefy/blob/dev/src/components/modal/Modal.vue) (line 116), while the documentation said it was false.


jtommy:
Thanks

PR #2744, opened 2020-08-12
Improve Tabs a11y

### LINKED ISSUE #2602: Make tabs buttons NOT anchors - a11y focus issue
**Buefy** version: [0.8.20]

### Description
Tabs are not focusable.
Tabs are currently using anchor tags which is not correct.
Tabs should be buttons.

The following is a related issue, but I am creating this issue because it is more specific and hopefully is something that can be addressed quickly:
https://github.com/buefy/buefy/issues/576

### Steps to reproduce
Try to use your keyboard or a screen reader to navigate tabs - you can not.

### Expected behavior
Tabs be focusable elements. This can be fixed by using buttons instead of anchors with no href.

### Actual behavior
Buefy tabs are not focusable.

For reference here is a suggested implementation of tabs:
https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html

service-paradis:
_**Please note that it would be preferable to test this a little bit more before merging. Would love to have feedback from people since I'm not a a11y expert.**_

Fixes #2602

#### Roles and attributes (Based on https://www.w3.org/TR/wai-aria-practices/#tabpanel)

- The element that serves as the container for the set of tabs has role tablist.
- Each element that serves as a tab has role tab and is contained within the element with role tablist.
- Each element that contains the content panel for a tab has role tabpanel.
- If the tab list has a visible label, the element with role tablist has aria-labelledby set to a value that refers to the labelling element. Otherwise, the tablist element has a label provided by aria-label.
- Each element with role tab has the property aria-controls referring to its associated tabpanel element.
- The active tab element has the state aria-selected set to true and all other tab elements have it set to false.
- Each element with role tabpanel has the property aria-labelledby referring to its associated tab element.
- If a tab element has a pop-up menu, it has the property aria-haspopup set to either menu or true.
- If the tablist element is vertically oriented, it has the property aria-orientation set to vertical. The default value of aria-orientation for a tablist element is horizontal.

#### Keyboard interaction (Based on https://www.w3.org/TR/wai-aria-practices/#tabpanel)

- Tab: When focus moves into the tab list, places focus on the active tab element. When the tab list contains the focus, moves focus to the next element in the page tab sequence outside the tablist, which is typically either the first focusable element inside the tab panel or the tab panel itself.
- When focus is on a tab element in a horizontal tab list:
  - Left Arrow: moves focus to the previous tab. If focus is on the first tab, moves focus to the last tab. Optionally, activates the newly focused tab (See note below).
  - Right Arrow: Moves focus to the next tab. If focus is on the last tab element, moves focus to the first tab. Optionally, activates the newly focused tab (See note below).
- When focus is on a tab in a tablist with either horizontal or vertical orientation:
  - Space or Enter: Activates the tab if it was not activated automatically on focus.

#### Vertical tabs (Based on https://www.w3.org/TR/wai-aria-practices/#h-note-25)

- If the tabs in a tab list are arranged vertically:
  - Down Arrow performs as Right Arrow is described above.
  - Up Arrow performs as Left Arrow is described above.
- If the tab list is horizontal, it does not listen for Down Arrow or Up Arrow so those keys can provide their normal browser scrolling functions even when focus is inside the tab list.)


pez-wb:
What is the status of this PR? 
Last update was in August and it appears to be out of sync with the current head state on `buefy:dev`. 

Do we know if any of the a11y issues in Buefy are being addressed? I'm willing to assist where I can, but before I invest a bunch of time I want to make sure PRs related to a11y are going to be accepted/vetted.

service-paradis:
> What is the status of this PR?
> Last update was in August and it appears to be out of sync with the current head state on `buefy:dev`.

The branch is now up to date with the current head state. I'm still waiting for feedback for this PR. I asked on #576 and #2602. Unfortunately, I got none.

> Do we know if any of the a11y issues in Buefy are being addressed? I'm willing to assist where I can, but before I invest a bunch of time I want to make sure PRs related to a11y are going to be accepted/vetted.

a11y improvements requests are welcome. PRs to address them are also welcomed.

pezillionaire:
Excellent! Thank you!  I'll pull it down and take a look.

While I am by no means a "expert" I do have some experience in this area so I can provide feedback and perhaps even contribute some.

src/scss/components/_tabs.scss
```diff
@@ -1,8 +1,22 @@
...
+            a:focus {
+                outline: $tabs-focused-outline;
+                border-bottom-color: $tabs-link-focus-active-border-bottom-color;
+            }
+            &:not(.is-active) {
+                a:focus {
+                    border-bottom-color: $tabs-link-focus-border-bottom-color;
+                }
+            }
+
```
pezillionaire, line 19: this is good for the the default type but other types of tab should have their `:focus` match to the `:hover` state

src/components/tabs/Tabs.vue
```diff
@@ -80,6 +112,67 @@ export default {
                 }
             ]
         }
+    },
+    methods: {
+        manageTablistKeydown(event) {
+            // https://developer.mozilla.org/fr/docs/Web/API/KeyboardEvent/key/Key_Values#Navigation_keys
+            const { key } = event
+            switch (key) {
+                case this.vertical ? 'ArrowUp' : 'ArrowLeft':
```
pezillionaire, line 121: I would argue that regardless of visual orientation both up/down AND left/right should be used for tab switching.

pezillionaire (approved):
This is a DEFINITE improvement over the current itereration.

Some notes based on interactions on the `/documentation/tabs` page:

- Some sort of `focus` state should be visible on the `is-active` tab when focused via keyboard navigation (outline, colour change, etc.). This makes it easier to identify the position of the focus on the page and that the tabs are selected and ready for interaction.
- When the `type` of tab (`is-boxed` etc.) is changed the `focus` state should be the same as `hover` state in keeping with conventions.
- Custom headers (or at least the examples found on `documentation/tabs#custom-headers`) appears not to take the focusing state - perhaps an issue with the `b-slot-component`
- while the switching of left/right and up/down for horizontal and vertical tabs is clever it would make more sense if left/right and up/down both worked for tab switching regardless of visual context.
- tabs use `<a>` element as default - this should be changed to `<button>`. Use of `<a>` should be reserved solely for links. Any "on page" interactions sould use buttons by default. this should be rectified globally in Buefy but would require a larger PR to address.

Again, ANY A11y updates are welcomed at this point and nothing "breaks" the current interaction model from what I've been able to test. In that regard I'm going to approve this PR.

I would like to see the chages mentioned addressed (I can help contribute code if that would help) and Im happy to see continued updates to fix a11y issues in buefy.

src/components/tabs/Tabs.vue
service-paradis: First of all, thank you very much for your review!! I really appreciate it 😃 
According to https://www.w3.org/TR/wai-aria-practices/#h-note-25, we should keep default behavior of arrow buttons depending on tabs orientation.

pezillionaire: No prob. Yeah, my reasoning was more to avoid the context switch — but yeah if that's the recomendation that seems like the way to go. There are bigger fish to fry WRT a11y in here! 😄

service-paradis:
@pezillionaire As mentioned before, thank you very much for the reivew. Here are some comments:


> Some sort of focus state should be visible on the is-active tab when focused via keyboard navigation (outline, colour change, etc.). This makes it easier to identify the position of the focus on the page and that the tabs are selected and ready for interaction.

I agree with you, but I did not want to add any "breaking changes" to this PR. Since this is a UI library, any visual changes are, kind of, breaking. To prevent this, I added a SCSS variable called `$tabs-focused-outline`. The default is `none`, so there are no visual changes for focused active tabs. But, users that want to add an outline on focus can customize this variable.

I also added a SCSS variable for the active border-bottom color called `$tabs-link-focus-active-border-bottom-color`. Same as the above, the active tab is not styled as default to keep the same behavior, but can be customized.

For example, customizing variables to the following would add outlines and border-bottom to any focused element (even the active ones).
```scss 
$tabs-focused-outline: #000 1px auto;`
$tabs-link-focus-active-border-bottom-color: $blue;
```

In a future breaking release, we might be able to add default styles to focused elements.

> When the type of tab (is-boxed etc.) is changed the focus state should be the same as hover state in keeping with conventions.

You're totally right, I missed it. I just pushed a new commit for it. Like mentioned above, the active wont be styled on focus as default. I will add SCSS variables to be able to customize it.

> Custom headers (or at least the examples found on documentation/tabs#custom-headers) appears not to take the focusing state - perhaps an issue with the b-slot-component

Good catch!! Just pushed a fix!

> while the switching of left/right and up/down for horizontal and vertical tabs is clever it would make more sense if left/right and up/down both worked for tab switching regardless of visual context.

According to https://www.w3.org/TR/wai-aria-practices/#h-note-25, we should keep default behavior of arrow buttons depending on tabs orientation.

> tabs use <a> element as default - this should be changed to <button>. Use of <a> should be reserved solely for links. Any "on page" interactions should use buttons by default. this should be rectified globally in Buefy but would require a larger PR to address.

I agree! Unfortunately, it is not totally supported right now by Bulma. Once they made the change to support it, we will follow along for sure! There are some issues about this in their repo (ex. https://github.com/jgthms/bulma/issues/3261).

> Again, ANY A11y updates are welcomed at this point and nothing "breaks" the current interaction model from what I've been able to test. In that regard I'm going to approve this PR.

🎉

PR #2747, opened 2020-08-17
Fixed an error when unselected and customRowKey is set

ymorika:
## Proposed Changes

Sorry. Fix #2735 did not take into account `selected === undefined`.
Fixed to be false in the case of undefined.

### Example

https://codepen.io/hikaruworld/pen/WNwwVdR

(Check the console.)


ymorika:
Wrong account.

PR #2773, opened 2020-08-19
Set $input-arrow only if it is not defined

jeromelebleu:
Fixes the default value of `$select-arrow-color` to use the same arrow color for select as Bulma.

## Proposed Changes
- Set `$input-arrow` only if it is not defined

### DIFF
--- a/src/scss/utils/_variables.scss
+++ b/src/scss/utils/_variables.scss
@@ -15,5 +15,4 @@ $link-focus-border: $primary !default;
 
 $label-weight: $weight-semibold !default;
 
-$input-arrow: $primary;
-
+$input-arrow: $primary !default;


jtommy:
Are you able to customize it in your code ?

jeromelebleu:
> Are you able to customize it in your code ?

Yes, I have to define `$select-arrow-color` for that.

> Btw I don't want bulma primary but Buefy primary

You could maybe add a note in the changelog in this case. I was wondering why suddenly the arrow color was not the same as my `<select>` anymore and spend some time before understanding why!

Anyway, the problem I see is that by importing Buefy's styles in my SCSS file, `$input-arrow` is redefine in the rest of my code.

jtommy:
`$select-arrow-color` is was a my test but i didn't want to release in the 0.9.x. Btw there is a good solution to override $input-arrow with default the primary color of buefy (thanks @service-paradis )

PR #2779, opened 2020-08-19
Tooltip `append-to-body` and `always` bug fixed

### LINKED ISSUE #2778: Tooltip `append-to-body` and `always` incompatible
### Overview of the problem

**Buefy** version: [0.9.1]

### Description

Since the addition of the `append-to-body` property to the Tooltip, the `always` does not always apply anymore, since it applies a css class that is ignored by the `append-to-body` JS behaviour.

### Steps to reproduce

1. <b-tooltip append-to-body always>Hello world</b-tooltip>

### Expected behavior
Tooltip is visible.

### Actual behavior
Tooltip is not visible.

jacenre:
Fixes #2778 

## Proposed Changes

- Take into account when the `always` property is used in the  append-to-body` context


jtommy:
Thanks

PR #2799, opened 2020-08-23
Add events after-enter and after-leave to Modal

aglio:
## Proposed Changes

- Add events to be triggered after model component opens or closes animation.
- For example, this can be utilized in processing activities such as initialization or disposal of contents included in the modal after completion of the animation.
-  I required these events in my application therefore, would like to request these options to be added.

### DIFF
--- a/docs/pages/components/modal/api/modal.js
+++ b/docs/pages/components/modal/api/modal.js
@@ -137,13 +137,13 @@ export default [
                 parameters: '—'
             },
             {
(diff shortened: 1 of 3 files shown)


jtommy:
Ok, but please rename the events with the transition hook names, for example `after-opened` to `after-enter`, etc

aglio:
Thank you, I have renamed to after-enter/after-leave. Kindly check.

PR #2807, opened 2020-08-24
Fixing empty slot in autocomplete

eveenendaal:
The groups change in autocomplete broke the empty slot; since computeData now always returns an array with at least one element.


jtommy:
Thanks

PR #2813, opened 2020-08-25
Fix contextmenu event always being prevented on tooltips

adriaanmeuris:
Commit https://github.com/buefy/buefy/commit/5c76d23ddd8a69c604ed770bc4b1dfec5c4cd61d broke the context menu on elements with a tooltip as it's always prevented ([repro](https://codesandbox.io/s/buefy-template-forked-rjpuq?file=/src/App.vue), try right clicking the link to open the context menu)

## Proposed Changes
- only prevent the context menu when the `contextmenu` is not set in the `triggers` prop
- fixed a typo in the documentation + added missing `contextmenu` trigger
- this PR will again allow to right-click links with a tooltip to open them in a new tab

### DIFF
--- a/docs/pages/components/tooltip/api/tooltip.js
+++ b/docs/pages/components/tooltip/api/tooltip.js
@@ -29,7 +29,7 @@ export default [
                 name: '<code>triggers</code>',
                 description: 'Tooltip will be triggered by any events',
                 type: 'Array',
-                values: '<code>click</code>.<code>hover</code>,<code>focus</code>',
+                values: '<code>click</code>,<code>hover</code>,<code>focus</code>,<code>contextmenu</code>',
                 default: '<code>[\'hover\']</code>'
             },
             {
--- a/src/components/tooltip/Tooltip.vue
+++ b/src/components/tooltip/Tooltip.vue
@@ -16,7 +16,7 @@
             ref="trigger"
             class="tooltip-trigger"
             @click="onClick"
-            @contextmenu.prevent="onContextMenu"
+            @contextmenu="onContextMenu"
             @mouseenter="onHover"
             @focus.capture="onFocus"
             @mouseleave="close">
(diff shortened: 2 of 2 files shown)


jtommy:
@adriaanmeuris thanks, but i don't understand well your issue, it works well

adriaanmeuris:
@jtommy I want to use a tooltip on a link. This currently disables the context menu, so end-users are not able to right-click on the link and choose "open link in a new tab" from the context menu.

jtommy:
Oh you're right, but your code isn't correct. We should prevent only when contextmenu is enabled, don't you ?

adriaanmeuris:
No: if it's enabled then you explicitly indicate it **should** trigger, so only prevent when it's disabled:
```js
// 'contextmenu' is not a trigger, so we prevent
if (this.triggers.indexOf('contextmenu') < 0) return e.preventDefault()
```

service-paradis:
The issue is real. We dont want to always prevent the `contextmenu`. I think the desired behavior is the following:

- If we use `contextmenu` as a trigger: We want to open the tooltip using right click, but prevent the default `contextmenu` to be open
- If we dont use `contextmenu` as a trigger: We do not want to open the tooltip using right click (we want it to show on hover for example). In this case, we dont want to prevent the default `contextmenu` to be open.

So, I think @jtommy is right. In your repro above, you would need to remove `contextmenu` from the triggers. This way, once fixed, the right click will display the default `contextmenu` (to open in a new tab for example)

adriaanmeuris:
@service-paradis thanks, your suggestion makes sense. I'd be happy to update the PR I however don't really see the use case for triggering a tooltip on right click. Any idea why this was added?

jtommy:
As action menu like dropdown

adriaanmeuris:
Thanks guys, I have updated the PR according to https://github.com/buefy/buefy/pull/2813#issuecomment-680919075

PR #2831, opened 2020-08-29
Allow number input values smaller than step size

abroekhof:
## Proposed Changes 

Add a new prop to BNumberInput which allows for input values with fractional values smaller than `step`, without triggering a validation error. 

The previous workaround was to set `step` to the desired precision, which may be undesirable in terms of incrementing/decrementing the input value using the up/down buttons.

PR for #2832

### DIFF
--- a/docs/pages/components/numberinput/api/numberinput.js
+++ b/docs/pages/components/numberinput/api/numberinput.js
@@ -72,7 +72,14 @@ export default [
                 description: 'Incremental number step',
                 type: 'Number, String',
                 values: '—',
-                default: '—'
+                default: '1'
+            },
+            {
+                name: '<code>min-step</code>',
+                description: 'Minimum step-size allowed. Input value is validated to be integer multiple of <code>min-step</code>',
+                type: 'Number, String',
+                values: '—',
+                default: 'Defaults to value of <code>step</code>'
             },
             {
                 name: '<code>exponential</code>',
--- a/docs/pages/components/numberinput/examples/ExStep.vue
+++ b/docs/pages/components/numberinput/examples/ExStep.vue
@@ -9,5 +9,10 @@
             <b-numberinput step="0.01">
             </b-numberinput>
         </b-field>
+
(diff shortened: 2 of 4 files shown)


jonagoldman:
I like the feature but you need to be careful, javascript have some problems dealing with floating point numbers: 
```js
0.1 + 0.2 = 0.30000000000000004
```

jtommy:
It looks a breaking change to be honest

abroekhof:
@jtommy, what are you referring to specifically? If it was those .npmignore, `prepare` changes, that was a mistake on my end (used to Gerrit workflow...). Should be fixed now. If for this change in particular, I've made `minStep` optional to allow for backwards compatibility.

@jonagoldman, I see what you mean, I'll make it more robust and add some more tests.

Thanks for taking a look!

abroekhof:
@jonagoldman I've changed the unit tests to exercise float addition/subtraction--appears to be working OK. Let me know what you think!

jtommy:
@abroekhof Nevermind, my fault

jonagoldman:
@abroekhof I tested a few different common combinations and I didn't find any issues.

@jtommy I found an **unrelated** issue where the validation message is not triggered when pressing the control buttons, only when you focus in and out the input, maybe this is default behavior, not sure:

Peek 2020-08-30 22-00

abroekhof:
@jonagoldman, technical writing is definitely not one of my strong suits, but I'll try to clear things up! 

I also noticed the behavior you're seeing regarding the validation message, but haven't investigated. I suppose in some ways it's "correct", since it pertains to an explicit user input, and gives the user a chance to understand the validation constraints. But on the flip-side I would expect it to clear once the input is valid.

jonagoldman:
@abroekhof I think your original description was good:

"Allows a valid minimum fractional value smaller than `step`. Input value is validated to be integer multiple of `min-step`."

jtommy:
Why do you need it? Probably I'm missing something but you can use the current step prop.

PR #2840, opened 2020-09-01
2754-range-props-with-type-month

### LINKED ISSUE #2754: Allow Month/Range in Buefy Date Picker
### Description

Essentially, this feature would allow a user to select a number of months

```html
<b-datepicker type="month" range>
```

I think this feature would be good cause in some situations, the user doesn't want to select a number of dates, however wants to select a number of months and this would be a quicker interface.

image

ryo-gk:
Fixes #2754 

## Proposed Changes

- Enable to use `range` props with `type='month'`.

## Images
image

image

## Others
This is a draft because the I've not written unit tests yet.

### DIFF
--- a/src/components/datepicker/Datepicker.vue
+++ b/src/components/datepicker/Datepicker.vue
@@ -162,7 +162,10 @@
                             :events="events"
                             :indicators="indicators"
                             :date-creator="dateCreator"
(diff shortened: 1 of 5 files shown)


jtommy:
Good addition!! Waiting for you ;)

PR #2866, opened 2020-09-07
Fix problem using `lazy` prop for Image component

service-paradis:
## Proposed Changes

There was problem with:
- SSR
- Nuxt
- Old browsers that were not supporting lazy attribute nor Intersection Observer


Jlaird:
Anyone have an idea when this will be released?

Issue #2862, opened 2020-09-05
Autocomplete component does not trigger select event


ParadoxFilm:
### Overview of the problem

**Buefy** version: 0.9.3
**Vuejs** version: 2.6.12
**OS/Browser**: Microsoft Windows 10 _(10.0.17134_) / Chrome 85.0.4183.83

### Description

The **autocomplete** component will not trigger the `select` event if you click the **clear** button the first time (the `clearable` property is set to `true`). This is the case when the filter value is set initially. This is **IMPORTANT** if you need to set the value of the component initially. I do not know any other way how to set the value of a binded with v-model variable.

### Code to reproduce
Sorry, CodeSandbox is not working for me now. A little code snippet here:
```js
<template>
  <b-autocomplete
    v-model="itemsFilter"
    :data="items"
    field="name"
    open-on-focus
    @select="option => selectTriggered()"
    clearable
  ></b-autocomplete>
</template>

<script>
export default {
  data: () => ({
    itemsFilter: "aaa", // filter/component initial value.
    items: [
      { id: 1, name: "aaa" },
      { id: 2, name: "bbb" },
    ],
  }),
  methods: {
    selectTriggered() {
      console.log("triggred");
    },
  },
};
</script>
```

### Expected behavior
The `select` event is expected to be triggered at this point. The option value must be undefined.

### Actual behavior
The `select` event is not called when you click on the **clear** button.

PR #2867, opened 2020-09-07
fix: setSelect null when clear button is triggered

re2005:
Fixes #
https://github.com/buefy/buefy/issues/2862

## Proposed Changes

- Set the 'empty' new value


jtommy:
Thanks

PR #2875, opened 2020-09-10
Improve tabbing with autocompleted, open-on-focus taginputs

### LINKED ISSUE #2871: Taginput with Autocomplete: pass taginput’s confirm-keys prop to autocomplete
### Description

Scenario: multiple tag inputs with open-on-focus autocompletes, where each input typically will contain more than one tag

Problem: Keyboard-only navigation becomes complicated, as both enter and tab will select the highlighted option (and tab will additionally move to the next field)

Solution: new prop allowing taginput to not add tag on tab (leaving enter = select, tab = move to next field)

rubjo:
Fixes #2871

## Issue
Tab navigation with autocompleted taginputs with `openOnFocus` always selects the hovered option.

## Proposed Changes

- Add `confirmKeys` prop to Autocomplete
- Add Tab to Taginput `confirmKeys` and pass `confirmKeys` to Autocomplete

This will allow the user to move to the next field with the Tab key by using prop `confirmKeys: [',', 'Enter']` instead of the new default `[',', 'Tab', 'Enter']`.

### DIFF
--- a/docs/pages/components/autocomplete/api/autocomplete.js
+++ b/docs/pages/components/autocomplete/api/autocomplete.js
@@ -100,6 +100,13 @@ export default [
                 default: '<code>mdi</code>'
             },
             {
+                name: '<code>confirm-keys</code>',
+                description: 'Array of keys (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values) which will select an option when typing (default tab and enter)',
+                type: 'Array',
+                values: '—',
+                default: '<code>["Tab", "Enter"]</code>'
+            },
+            {
                 name: '<code>clearable</code>',
                 description: 'Add a button to clear the inputed text',
                 type: 'Boolean',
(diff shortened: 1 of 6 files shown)


rubjo:
Hmm, I think I need some help to make those tests pass, seems like keydown events aren't simulated.

service-paradis:
Maybe you will need to use something like this in your test (not tried):
```
trigger.keyPress('keydown', {'key': 'Enter'})
```

rubjo:
Tried with lowercase
```
trigger.keyPress('keydown', {'key': 'enter'})
```
to no avail. Can try some more tomorrow.

service-paradis:
Updated my comment

rubjo:
Yep, correct casing for the keys helped. :)

Locally, some of the *picker components' tests fail for me. Something to do with node / some package version, maybe, I have no idea.

Anyway, they all pass on CircleCI now, so that's dandy.

service-paradis (approved):

jtommy:
Thanks @rubjo @service-paradis

PR #2876, opened 2020-09-11
add cellClick event to b-table

dauriata:
## Proposed Changes

add a `cellClick` event triggered on a click of a `<td>` element in the main content of a b-table
with parameters : row, column, rowIndex, columnIndex 

- Use case : 
to detect click event on a cell where it is not possible to set the event listener on the content of the cell, 
ex: empty cell or content of the cell not the full size of the `<td>` element and be able to click aside of the content.

- As a side change, added colindex to b-table-column props

### DIFF
--- a/docs/pages/components/table/api/table.js
+++ b/docs/pages/components/table/api/table.js
@@ -399,7 +399,7 @@ export default [
             {
                 name: 'default',
                 description: '<strong>Required</strong>, table body and header',
-                props: '<code>row: Object</code>, <code>index: Number</code>'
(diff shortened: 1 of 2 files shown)


dauriata:
@jtommy just noticed that I also renamed `index` to `rowIndex` in the default slot props to be consistent in naming with `columnIndex`
But that might be a breaking change for those who where using that prop. 
should we consider keeping the old name of `index` and add `colindex` as in the code ?

jtommy:
where ?

dauriata:
actually I updated the docs but the modified default slot prop code is not in the commit
let me correct the PR with an updated version

dauriata:
Actually only the api doc is wrong, how can i correct this ?
should I must reopen a PR now that it's merged ?

jtommy:
New PR please

PR #2888, opened 2020-09-16
Add color to icon on buttons

CharlieBrownCharacter:
## Proposed Changes
Currently we are not able to add a type to an icon inside a button. This PR fixes this issue by applying the `icon-type` prop to a button, thus allowing users to select a type of an icon inside each button

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "buefy",
-  "version": "0.9.3",
+  "name": "foreo--buefy",
+  "version": "0.9.4",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "buefy",
-  "version": "0.9.3",
+  "name": "foreo--buefy",
+  "version": "0.9.4",
```
service-paradis, line 3: You should remove this changes from your PR

Issue #2834, opened 2020-08-31
Pull request proposal: Number input controls position left/right/center


jonagoldman:
### Description
Allow positioning the + and - controls to the left, right or center (default)

![Screenshot-20200830230147-331x273](https://user-images.githubusercontent.com/1297559/91682250-89af5b00-eb16-11ea-9acf-be9a5c70d927.png)

### Why Buefy need this feature
I think having the controls to the right is the more "standard" way.
Browsers default number input is on the right and also in most other libraries.

### Naming
I have already implemented a new `controls-alignment` prop:
```js
export default {
  // or controlsAlign?
  props: {
    controlsAlignment: {
      type: String,
      default: 'center',
      validator: (value) => {
        return [
          'center',
          'left',
          'right'
        ].indexOf(value) >= 0
      }
    }
  }
  // ...
}
```

My only issue is that there is currently a `controls-position` prop that can be set to `compact`, but I think this name is misleading because `position` generally means `left`, `right`, `top`, `bottom`, but not `compact`.

I think the "correct" way should be Boolean property called `controls-compact` or just `compact` and then use `controls-position` for the actual positioning, but this would be a breaking change.

So I guess my question is, if you are interested in this how should the prop be called? It's also possible to merge this options with `controls-position` if you think so.

jonagoldman:
Great! Should I extract the button to a new `NumberinputControl` component or just continue the same way it is right now?

jtommy:
As you prefer

PR #2895, opened 2020-09-16
Add controls-alignment prop to Numberinput

jonagoldman:
Fixes https://github.com/buefy/buefy/issues/2834

## Proposed Changes

- Add new `controls-alignment` property with possible values `left`, `right` or `center` (default). See issue.


jtommy:
Thanks!

PR #2903, opened 2020-09-18
Update Customization.vue

thefoxie:
There's been changes in bulma. If you change variable ($primary in this case) you have to provide its light and dark variants. Also, there's new option to add custom colours via $custom-colors variable that is passed as argument to mergeColorMaps function. More info in this: https://bulma.io/2019/10/15/light-dark-colors/#4-value-color-map

## Proposed Changes

- update docs on bulma variables customization


jtommy:
Thanks

PR #2923, opened 2020-09-25
Update input.js

AkmarNafi:
added minlength property to api docs

Fixes #

## Proposed Changes

- added minlength property of b-input  to api docs, which was previously missing.
-
-


kaangokdemir (changes requested):
There is not any `minlength` prop in Input component. It's just passing it's attributes to child native input or textarea component. This changes are misleading

PR #2926, opened 2020-09-26
left-label prop added for b-switch component

kaangokdemir:
## Summary
`left-label` prop added into `b-switch` component.

## Why Buefy needs this?
Currently label of switch component is displaying at right of the component. If for some reasons users need to display them at left, they need to write custom CSS instead.

## Motivation
It's coming from Quasar Framework's `toggle` component's `left-label` prop.
https://quasar.dev/vue-components/toggle#With-labels

## Examples
Ekran-Kayd-2020-09-26-200155

## Documentation
documentation

## Checklist
- [x] Documentation updated
- [x] New Example added
- [x] Unit Tests added

### DIFF
--- a/docs/pages/components/switch/examples/ExStyles.vue
+++ b/docs/pages/components/switch/examples/ExStyles.vue
@@ -30,11 +30,16 @@
                 <option value="is-large">is-large</option>
             </b-select>
         </b-field>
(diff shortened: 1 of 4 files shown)


jtommy:
Thanks

PR #2940, opened 2020-10-01
Fixes #2889 - moved trigger behind tooltip appended to body

### LINKED ISSUE #2889: Tooltip with append-to-body stutters
### Overview of the problem

**Buefy** version: [0.9.3]
**Vuejs** version: Nuxt [2.14.0]
**OS/Browser**: MacOs/ Chrome

### Description
Tooltip appended to body constantly fires `mouseenter`/`mouseleave` while mouse is moved over the trigger area crating constant show/hide for the tooltip during each move.

### Steps to reproduce

1. Create tooltip with append-to-body prop.
2. Move mouse over its trigger element.

### Expected behavior

Tooltip shows once when mouse is over the trigger element. Hides when it's off the trigger element.

### Actual behavior

Codesandbox: https://codesandbox.io/s/buefy-playground-forked-u6bfl - move mouse slowly over the "test" button.

sircodesalittle:
Fixes #2889

## Proposed Changes 
When a tooltip uses append-to-body (FYI this is not documented), make the trigger element z-index deeper than the appended element so that the hover isn't falsely triggered. This change only affects style when append-to-body is used on a tooltip.

### DIFF
--- a/src/components/tooltip/Tooltip.vue
+++ b/src/components/tooltip/Tooltip.vue
@@ -19,7 +19,8 @@
             @contextmenu="onContextMenu"
             @mouseenter="onHover"
             @focus.capture="onFocus"
-            @mouseleave="close">
+            @mouseleave="close"
+            :style="triggerStyle">
             <slot ref="slot" />
         </div>
     </span>
@@ -88,6 +89,9 @@ export default {
         return {
             isActive: false,
             style: {},
+            triggerStyle: {
+                zIndex: this.appendToBody ? '100' : 'auto'
+            },
             timer: null,
             _bodyEl: undefined // Used to append to body
         }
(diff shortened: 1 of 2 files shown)


jtommy:
I think it's better add a class instead of add a zindex on the vue component

jtommy:
I'm thinking about the value that you are using 100.. I don't think it's really correct

sircodesalittle:
In Tooltip.vue (line 147):

`javascript
                wrapper.style.zIndex = this.isActive || this.always ? '99' : '-1'
`

The z-index gets set to 99 or -1, depending on hover (when append-to-body is used), therefore making the z-index of the wrapper 100 removes the stutter by ensuring the wrapper is always on top.

jtommy:
Thanks, I added it programmatically

PR #2942, opened 2020-10-02
Added "before-sort" property to b-table

mk-startialab:
## Proposed Changes

- Added "before-sort" property to b-table.
The type of "before-sort" is "Function", and sorting procedure will be stopped if the function retuns "true".
This makes sorting procedure interruptible.

### DIFF
--- a/docs/pages/components/table/api/table.js
+++ b/docs/pages/components/table/api/table.js
@@ -45,6 +45,13 @@ export default [
                 default: '<code>is-small</code>'
             },
             {
+                name: '<code>before-sort</code>',
+                description: 'Sets the function which called when a sortable column clicked. This function will be called before sorting procedure starts. You can stop sorting procedure by returning <code>true</code> with this function',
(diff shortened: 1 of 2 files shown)


jtommy:
Why do you need it ?

mk-startialab:
@jtommy
Because I want to change sorting behavior.
e.g. Sorting specific column in descending order initially, and so on.

jtommy:
Do you want to mix default direction order on specific column ?

mk-startialab:
@jtommy 
Yes, but not only mixing default direction order.
My project specific feature implementation is related.

jtommy:
Yes but I need to understand if it's only an edge case of your project that you resolve using workaround before to include a similar feature in the core code

mk-startialab:
@jtommy 
Oh, I understand.

Please tell me the way how to mix default direction order on specific column.
As for the other cases, I manage to work out on my own.

PR #2953, opened 2020-10-07
Add table column typescript definitions

CharlieBrownCharacter:
- Added table column typescript definitions

## Proposed Changes

- Added type definitions for the column props of the table component


jtommy:
Thanks. If you want to add other types you are welcome

PR #2954, opened 2020-10-07
Adds autoFocus prop to Modal and updates trap-focus related docs.

mateuswetah:
## Proposed Changes

Adds an `auto-focus` prop, default `true`, to the modal component;
Also, during the process I noticed that all components that had the `trap-focus` feature had their documentation say that the value defaults to `false`, when its being a few releases that it is `true`.

Sorry, I forgot to create a new branch for this one!


jtommy:
Thanks

PR #2955, opened 2020-10-08
Checking if minNumber is undefined to assign it or not to newValue

### LINKED ISSUE #2945: [b-numberinput] remove fallback to minNumber if input is not a number
### Description

Inside `Numberinput.vue` file is the next code:
```javascript
if (value === '' || value === undefined || value === null) {
                    newValue = this.minNumber || null
                }
```
  
Here is veriffied if the input value is not valid. In case it is not valid, you assign `newValue` to `minNumber` or to null. These are my points:
1. Fisrt of all, I don't think it would be a good idea to fallback to the `minNumber` property. The reason si that if you want to validate the input, despite the input would be anything but a number, the validation library will accept it as a possible value because of the fallback. The next photo represents a b-numberinput which `min` property is set to 1 and even when there is not a valid input, the validation library accepts it, because of the fallback:
numberinput-0
2. In case you consider the fallback to `minNumber` should be kept, there is a problem when it is set to 0 and it is that `this.minNumber` will return as `false`. So the value will never fallback to `minNumber`. In the next image this is reflected with a b-numberinput which `min` property is set to 0:
numberinput-1

I think this feature will improve Buefy by making it more flexible. Validation is an important part of frontend development, so I think it will be better to fallback to `minNumber` in case input is not valid. By doing this, the validation library will be able to detect the incorrect user input

elC0mpa:
Fixes #2945 

## Proposed Changes

- Fixes bug when setting `min` property to 0

### DIFF
--- a/src/components/numberinput/Numberinput.vue
+++ b/src/components/numberinput/Numberinput.vue
@@ -147,9 +147,9 @@ export default {
                 let newValue = value
                 if (value === '' || value === undefined || value === null) {
                     if (this.minNumber !== undefined) {
-                      newValue = this.minNumber;
+                        newValue = this.minNumber
                     } else {
-                      newValue = null;
+                        newValue = null
                     }
                 }
                 this.newValue = newValue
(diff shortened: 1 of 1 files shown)


jtommy:
Please fix the indentation so I can merge it

elC0mpa:
Ok @jtommy

elC0mpa:
Here are the identation problems solved

PR #2963, opened 2020-10-08
Update Readme Buefy Logo for Shields.io

service-paradis:
Proposed Changes

Since Buefy icon is now available in https://simpleicons.org/?q=buefy, we can display the icon by using its name only instead of the base64 string.

Ref: https://github.com/simple-icons/simple-icons/pull/3430


jtommy:
Awesome

PR #2970, opened 2020-10-10
Typescript button props definitions

CharlieBrownCharacter:
## Proposed Changes

- add typescript button props definitons

### DIFF
--- a/types/components.d.ts
+++ b/types/components.d.ts
@@ -1,5 +1,5 @@
 import _Vue from "vue";
-import { ColorModifiers, GlobalPositions } from "./helpers";
+import {ColorModifiers, GlobalPositions, SizesModifiers} from "./helpers";
 
 // Component base definition
 export class BComponent extends _Vue {
@@ -95,7 +95,7 @@ export declare type BDialogConfig = {
     /**
      * Dialog\'s size, optional
      */
-    size?: 'is-small' | 'is-medium' | 'is-large';
+    size?: SizesModifiers;
 
     /**
      * Custom animation (transition name)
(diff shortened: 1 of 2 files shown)


CharlieBrownCharacter:
@jtommy I'm heavily using Buefy with typescript and have been adding a lot of typing definitions. With this in mind I think that the current structure will not be scalable. I was thinking of re-organizing the files in such structure:

```
- components
  - button.d.ts
  - carrousel.d.ts
  - ...
- helpers.d.ts
- index.d.ts
```

I know this comment is out of bound of this PR but I don't know how to reach you directly to ask such questions

jtommy:
Ok about reorganization, thanks

jtommy:
Looking your code I don't understand where you are defining button component. Am I missing anything ?

CharlieBrownCharacter:
@jtommy I didn't switch the components to what we spoke about (divide by folders as we spoke about)

jtommy:
Yes, but at the moment thee missing type error is not fixed

CharlieBrownCharacter:
@jtommy what do you mean?

jtommy:
Where is button component definition?

CharlieBrownCharacter:
@jtommy over here: `types/components.d.ts`. You can check the interface defined `ButtonProps`

jtommy:
But the component is Button and not ButtonProps

CharlieBrownCharacter:
Usually I give the name props because this is not the type definition for the entire component but for the props. The definition for the entire component would involve something like `ComponentOptions<any, any, ButtonProps>`

jtommy:
Ok, where is the component definition?

PR #2974, opened 2020-10-12
Added 'sticky-checkbox' option to b-table (#2635)

### LINKED ISSUE #2635: Allow checkboxes to be the sticky column in a table
It would be helpful if Buefy allowed checkable tables to keep the checkboxes as sticky columns.

When choosing to keep a column sticky in a table, most times a user will want the checkbox to remain visible so they can select a row while viewing the rest of the table's content horizontally. This avoids the need to have to scroll all the way back to the beginning of the table to be able to check the row and take an action.

Numenorian:
Fixes #2635 

## Proposed Changes

- Add a `sticky-checkbox` prop to the `<b-table>` component, which, if the table is `checkable`, will make the checkbox column sticky. 
- Add example and explanation of use to the Table documentation page.

Peek 2020-10-12 13-56

Screenshot from 2020-10-12 14-03-42

### DIFF
--- a/docs/pages/components/table/api/table.js
+++ b/docs/pages/components/table/api/table.js
@@ -101,6 +101,13 @@ export default [
                 default: '<code>left</code>'
             },
             {
+                name: '<code>sticky-checkbox</code>',
+                description: 'Make the checkbox column sticky when <code>checkable</code>',
+                type: 'Boolean',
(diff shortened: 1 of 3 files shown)


jtommy:
Good addition!!

PR #2996, opened 2020-10-22
Fix issue 2944

### LINKED ISSUE #2944: Tab items disappear when using v-model, v-for, and header slot
### Overview of the problem

**Buefy** version: [0.9.3]
**Vuejs** version: [2.6.11]
**OS/Browser**: any

### Description

When using `v-model` on a `<b-tab>` component, and creating the inner `<b-tab-item>` components with a `v-for` loop and custom headers, the tabs disappear when cycling through them.

### Steps to reproduce

1. Create `<b-tabs v-model="activeTab"></b-tabs>` component
2. Add loop inside to generate tab items, like `<b-tab-item v-for="tabName in tabArray" :key="tabName"></b-tab-item>`
3. Create custom headers using `<template #header> Custom header here </template>`

### Expected behavior

Tabs behave normally, clicking on one makes it active and shows its content without impacting the others.

### Actual behavior

When clicking on any tab, all tabs to the right of the new active tab disappear during the transition.
(This also happens if the tabs are vertical, with the tabs below the new active one disappearing.)

https://codepen.io/Acccent/pen/bGebwew?editors=1010

yuchaosydney:
Fixes #2944

## Cause
When switch between tabs, `childItem.$slots.header` in `src/components/tabs/Tabs.vue` could temporarily become falsy 
## Proposed Changes
- As per [this vue issue](https://github.com/vuejs/vue/issues/11738), if user define slot as `<template #header>`, this results in scope slot so slot should be accessed via `$scopedSlots`.
- Access named slot "header" via `$scopedSlots` rather than from `$slots` to cover this scenario
- `<template slot="header">` will still be supported though "Named Slots with the slot Attribute" will be deprecated in Vue 2.6.0+
## Reproduce in buefy tabs document page
I have made a change (adding both **v-model** & a **custom header** on **b-tabs** as per reproduce steps in #2944 ) in buefy project in a new branch that help to see the issue, please checkout branch https://github.com/yuchaosydney/buefy/tree/issue-2944, and head to http://localhost:8080/documentation/tabs, "Music tab" in first example has the same issue
### Before fix
before-fix

### After fix
after-fix

### DIFF
--- a/docs/pages/Extensions.vue
+++ b/docs/pages/Extensions.vue
@@ -1,7 +1,7 @@
 <template>
     <article class="content">
         <p>
-            Buefy has no dependencies aside from Bulma. But sometimes we need to add a very specific funcionality to our application.
+            Buefy has no dependencies aside from Bulma. But sometimes we need to add a very specific functionality to our application.
         </p>
         <p>
             These extensions are some examples of how you can extend some of the components using external libraries.
--- a/docs/pages/components/slider/api/slider.js
+++ b/docs/pages/components/slider/api/slider.js
@@ -129,6 +129,13 @@ export default [
                 type: 'Boolean',
                 values: '—',
                 default: '<code>false</code>'
+            },
+            {
+                name: '<code>tooltip-always</code>',
+                description: 'Tooltip displays always',
+                type: 'Boolean',
+                values: '—',
+                default: '<code>false</code>'
             }
         ],
         slots: [
--- a/docs/pages/components/slider/examples/ExCustomize.vue
+++ b/docs/pages/components/slider/examples/ExCustomize.vue
@@ -4,6 +4,10 @@
             <b-slider v-model="sliderValue" :tooltip-type="sliderType"></b-slider>
         </b-field>
 
+        <b-field label="Tooltip Always">
+            <b-slider tooltip-always></b-slider>
+        </b-field>
+        
         <b-field label="Hide tooltip">
             <b-slider :tooltip="false"></b-slider>
         </b-field>
--- a/docs/pages/components/tabs/examples/ExSimple.vue
+++ b/docs/pages/components/tabs/examples/ExSimple.vue
@@ -12,6 +12,10 @@
             </b-tab-item>
 
             <b-tab-item label="Music">
+                <template #header>
+        <b-icon icon="alert-circle"></b-icon>
+        <span>Music</span>
+      </template>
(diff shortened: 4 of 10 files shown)


yuchaosydney:
Dear @jtommy,

This fix is done by putting "$slots.header" existence check to a computed property in **TabbedChildMixin**, and enforce parent component **BTabs** to check against this computed property rather than directly based on **$slot** property from **BTabItem**.

My understanding is a computed property only checks "this.$slots.header" once during BTabItem component creation because $slot is not reactive (reference from this [vue issue](https://github.com/vuejs/vue/issues/10179)), therefore this computed property would never change again even if "this.$slots.header" becomes falsy in later state.

However, honestly I do not quite understand why "this.$slots.header" in BTabs could become falsy during later states when replicating the issue steps. I have been digging into it for long time but could not get my head around it. Would you be able to provide any insights & suggestions? Could this ($slots.header value randomly changes) be a vue issue?

Thank you very much for your time!

Yuchao

jtommy:
To be honest it's very strange but I don't think your code fix it because you are moving check from template to a computed prop and the template should be like call a method. Anyway did you try to move it to a method into mixin ?

yuchaosydney:
> To be honest it's very strange but I don't think your code fix it because you are moving check from template to a computed prop and the template should be like call a method. Anyway did you try to move it to a method into mixin ?

Hi @jtommy ,

Thank you very much for your reply.

I did tried your suggestion, moving it to a method into mixin results in the same issue as putting `$slots.header` in template does, which is expected.

In addition, I am in the same page with you:

- Using a computed prop hide the issue rather than exactly resolve it (we shall resolve)
- Root cause to `$slots.header` becomes `falsy` remains unclear

Perhaps leave this to me for couple of more days and I shall find its root cause.

Thank you for your time.

Greetings from Yuchao :)

yuchaosydney:
Hi @jtommy ,

I set up a very simple sample in Vue that simulates our issue and raised an issue to Vue. Eventually got the exact answer from them!

Please check [this vue issue](https://github.com/vuejs/vue/issues/11738) for more detail.

Basically we should access name slot "header" via `$scopedSlots` instead of from `$slots` in Vue 2 :)

jtommy:
It's breaking change but it might be a fix :)
Please fix the docs too

yuchaosydney:
@jtommy Thank you! Will do

PR #3001, opened 2020-10-25
add invalid event to upload

adamgreenhall:
## Proposed Changes

- add an invalid event on upload component

Currently, if you drag and drop a file into the uploader that doesn't match the types in the `accept` property (filetypes), nothing happens in the UI. This event allows you to add an error message for the user.


jtommy:
Thanks

PR #3012, opened 2020-10-29
Add rounded property to BUpload Component

major-mayer:
## Proposed Changes
I think it would be a great addition to be able to use rounded upload buttons as a Buefy component, without any additional CSS.
At first I thought about simply opening a feature request, but then I decided to have a look in the code and contribute something  myself, because i really like the framwork. And hey this is my first PR to an open source repo :partying_face: 

Soo what I did:
- Added the necessary scss class to make the input field rounded (the button in the slot must be still set to "is-rounded" manually tho)
- Added the property rounded to the Upload vue component
- Added the new property to the documentation and provided an example how to use it
- Run the linter (works :champagne:) and the tests (doesn't work, but fails on other files, that i haven't touched... :confused: )

Hope that is sufficient to make a quick merge possible, if not please tell me. I am still new to this open source development process and looking forward to get some feedback :)

### DIFF
--- a/docs/pages/components/upload/examples/ExRounded.vue
+++ b/docs/pages/components/upload/examples/ExRounded.vue
@@ -12,14 +12,14 @@
     </b-field>
 
     <b-field label="Separated filename">
-        <b-field class="file is-primary"  :class="{'has-name': !!file}">
-            <b-upload v-model="file" class="file-label" rounded>
+        <b-field class="file is-primary"  :class="{'has-name': !!file2}">
+            <b-upload v-model="file2" class="file-label" rounded>
                 <span class="file-cta">
                     <b-icon class="file-icon" icon="upload"></b-icon>
                     <span class="file-label">Click to upload</span>
                 </span>
-                <span class="file-name" v-if="file">
-                    {{ file.name }}
+                <span class="file-name" v-if="file2">
+                    {{ file2.name }}
                 </span>
             </b-upload>
         </b-field> 
@@ -32,6 +32,7 @@ export default {
   data() {
     return {
       file: {},
+      file2 : null
     };
   },
 };
--- a/docs/pages/components/upload/examples/ExRounded.vue
+++ b/docs/pages/components/upload/examples/ExRounded.vue
@@ -1,0 +1,28 @@
+<template>
+  <section>
+    <b-field class="file">
+      <b-upload v-model="file" rounded>
+        <a class="button is-primary is-rounded">
+          <b-icon icon="upload"></b-icon>
+          <span>{{ file.name || "Click to upload"}}</span>
+        </a>
+      </b-upload>
+    </b-field>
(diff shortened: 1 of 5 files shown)


service-paradis:
Congrats on your first open source PR 💯 
I will review it!

docs/pages/components/upload/examples/ExRounded.vue
```diff
@@ -0,0 +1,22 @@
+<template>
+  <section>
+    <b-field class="file">
+      <b-upload v-model="file" rounded>
+        <a class="button is-primary is-rounded">
+          <b-icon icon="upload"></b-icon>
+          <span>{{ file.name || "Click to upload"}}</span>
+        </a>
+      </b-upload>
+    </b-field>
```
service-paradis, line 10: You dont need to use a rounded button if you use this in your example:

```html
<b-field class="file is-primary" :class="{'has-name': !!file}">
        <b-upload v-model="file" class="file-label" rounded>
            <span class="file-cta">
                <b-icon class="file-icon" icon="upload"></b-icon>
                <span class="file-label">{{ file ? file.name : "Click to upload" }}</span>
            </span>
        </b-upload>
    </b-field>
```

service-paradis: It would be good to also maange the case we want to use rounded with the more traditional approach and input `type="file"`. Ex.
```html
<b-field class="file is-primary" :class="{'has-name': !!file}">
        <b-upload v-model="file" class="file-label" rounded>
            <span class="file-cta">
                <b-icon class="file-icon" icon="upload"></b-icon>
                <span class="file-label">Click to upload</span>
            </span>
            <span class="file-name" v-if="file">
                {{ file.name }}
            </span>
        </b-upload>
    </b-field>
```
This is the actual behavior:
image

This is the expected behavior:
image

service-paradis (changes requested):

docs/pages/components/upload/examples/ExRounded.vue
major-mayer: Hey thanks for the review.
It's looking now like this:
grafik

I'm not exactly sure what you mean by the `input type="file"` suggestion. I thought this tag is always created automatically and not by the user?

service-paradis: Sorry for the misleading comment about the input. I was talking about displaying the file name like you did in a separate element. In the example, you should set `file: null`. This way, `v-if="file"` in your second example will display the right part only when there is a selected file.

major-mayer: All right, now i got what you mean :smile: Thanks for your help :+1: 
I altered the examples so that the second upload button looks now like the first one, until the user actually selected a file.
I think this will be good to go now

service-paradis (approved):

PR #3015, opened 2020-11-01
Fix Issue 2990

### LINKED ISSUE #2990: B-Table multi-sort on nested data fails, working in normal sort
### Overview of the problem

**Buefy** version: Dist
**Vuejs** version: 2.Dist
**OS/Browser**: Brave

### Description

When using data more than one level deep the multi-sort fails where the normal sort works.

### Steps to reproduce

See Code Pen. In short have data more than one level deep and apply multi-sort, the columns with nested data fail to sort.

### Expected behavior

Sort to be applied

### Actual behavior

No sorting 

Codepen: https://codepen.io/RMCodes/pen/bGewgVb?editors=1010

yuchaosydney:
Fixes #2990
## Proposed Changes

- When multi-sort on nested obj, sorting fields passed to `multiColumnSort` helper function is a prop path, use existing `getValueByPath` function to get the value at path of object

### Before fix
multi-sort-before-fix
### After fix

multi-sort-after-fix

### DIFF
--- a/src/utils/helpers.js
+++ b/src/utils/helpers.js
@@ -179,7 +179,9 @@ export function multiColumnSort(inputArray, sortingPriority) {
     const fieldSorter = (fields) => (a, b) => fields.map((o) => {
         let dir = 1
         if (o[0] === '-') { dir = -1; o = o.substring(1) }
-        return a[o] > b[o] ? dir : a[o] < b[o] ? -(dir) : 0
+        const aValue = getValueByPath(a, o)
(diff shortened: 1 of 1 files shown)


jtommy:
No, i don't want to use lodash, as you can see it is a Dev dependency

yuchaosydney:
> No, i don't want to use lodash, as you can see it is a Dev dependency

Thanks for this info. Good to know and using existing `getValueByPath` is much better

jtommy:
Thanks

PR #3027, opened 2020-11-07
Fixes #2678: Override is-boxed display style in NavbarDropdown

### LINKED ISSUE #2678: tab can reach collapsed b-navbar-dropdown when :boxed="true"
navbar

In the following example, using sequential navigation with tab we always reach Home -> Info -> About -> Contact -> Documentation. Whereas when `:boxed="false"` the sequence follows Home -> Info -> Documentation, as expected.

```html
<div id="app" class="container">
    <b-navbar>
        <template slot="start">
            <b-navbar-item href="#">Home</b-navbar-item>
            <b-navbar-dropdown :boxed="true" label="Info">
                <b-navbar-item href="#">About</b-navbar-item>
                <b-navbar-item href="#">Contact</b-navbar-item>
            </b-navbar-dropdown>
            <b-navbar-item href="#">Documentation</b-navbar-item>
        </template>
    </b-navbar>
</div>
```

The problem lies in the .is-bloxed class, it is set to always `display:block` and `opacity:0` when not toggled.  
This works well to the majority of users but it can be quite confusing for the one who uses assistive technologies.

```css
.navbar-dropdown.is-boxed {
    border-radius: 6px;
    border-top: none;
    box-shadow: 0 8px 8px rgba(10, 10, 10, 0.1), 0 0 0 1px rgba(10, 10, 10, 0.1);
    display: block;
    opacity: 0;
    pointer-events: none;
    top: calc(100% + (-4px));
    transform: translateY(-5px);
    transition-duration: 86ms;
    transition-property: opacity, transform;
```

`transition-property` sadly can't work with `display`, so I can't think of a way to make it work the way it is expected.

yuchaosydney:
Fixes #2678

## Proposed Changes

- Override class `is-boxed` `display` style to `display: none` as default in NavbarDropdown
- When "box" style dropdown is opened, its div `<div class="navbar-dropdown is-boxed"></div>` will be styled as `display:block` by `.navbar-item.is-active .navbar-dropdown` as shown in following screenshot (Therefore, this issue can be resolved by neither tweaking existing transition style nor making new vue transition logic):

Screen Shot 2020-11-07 at 11 04 53 pm

### DIFF
--- a/src/components/navbar/NavbarDropdown.vue
+++ b/src/components/navbar/NavbarDropdown.vue
@@ -89,3 +89,8 @@ export default {
     }
 }
 </script>
+<style lang="scss" scoped>
+.is-boxed {
+    display: none;
+}
+</style>
--- a/src/components/navbar/NavbarDropdown.vue
+++ b/src/components/navbar/NavbarDropdown.vue
@@ -89,8 +89,3 @@ export default {
     }
 }
 </script>
-<style lang="scss" scoped>
-.is-boxed {
-    display: none;
-}
-</style>
--- a/src/scss/components/_navbar.scss
+++ b/src/scss/components/_navbar.scss
@@ -15,9 +15,7 @@
         transition-property: opacity, visibility, transform;
     }
 
-    .navbar-item.has-dropdown {
-        &.is-active .is-boxed {
-            visibility: visible;
-        }
+    .navbar-item.has-dropdown.is-active .is-boxed{
+        visibility: visible;
     }
 }
--- a/src/scss/components/_navbar.scss
+++ b/src/scss/components/_navbar.scss
@@ -4,9 +4,20 @@
             justify-content: center;
             margin-left: auto;
         }
-    
+
         .navbar-end {
             margin-left: 0;
+        }
+    }
+
+    .navbar-dropdown.is-boxed {
+        visibility: hidden;
+        transition-property: opacity, visibility, transform;
+    }
+
+    .navbar-item.has-dropdown {
+        &.is-active .is-boxed {
+            visibility: visible;
         }
     }
 }


src/components/navbar/NavbarDropdown.vue
```diff
@@ -89,3 +89,8 @@ export default {
     }
 }
 </script>
+<style lang="scss" scoped>
+.is-boxed {
+    display: none;
+}
+</style>
```
service-paradis, line 96: Thanks for this PR. I will need to make some tests.
Can you please add this code in `scss/components/_navbar.scss` instead? Buefy do not use scoped styles.

service-paradis (changes requested):
Thanks for this PR.
This solution solved the tabbing issue. Unfortunately, we lose the fading effect without using transition.

yuchaosydney:
> using

yuchaosydney:
> Thanks for this PR.
> This solution solved the tabbing issue. Unfortunately, we lose the fading effect without using transition.

Thanks for pointing out. I will make a change to use transition

src/scss/components/_navbar.scss
```diff
@@ -4,9 +4,18 @@
...
         }
-    
+
         .navbar-end {
             margin-left: 0;
         }
     }
+
+    .navbar-dropdown.is-boxed {
+        visibility: hidden;
```
yuchaosydney, line 14: Hi @service-paradis ,

There is actually a cleaner solution: although `transition-property` does not work with `display`, it works with `visibility`. And `visibility: hidden` disables tab event as well.

Therefore, I put `visibility: hidden` as default for `.navbar-dropdown.is-boxed`, and when dropdown is open `.navbar-item.has-dropdown.is-active`, set to `visibility: visible;`.

service-paradis (approved):
LGTM 💯 
Thanks for your help!

PR #3041, opened 2020-11-16
Provide an iconRightType prop

Propaganistas:
## Proposed Changes

- Add an `icon-right-type` prop to `BInput`

## Reason

I'd like to have a colored right icon inside my input. I don't think there's a proper way to do this without resorting to the input's general `type` property. However, that also colors the input border which is undesired.


jtommy:
Thanks but please fix lint error

Propaganistas:
Done!

PR #3047, opened 2020-11-18
Fix mutation of value in multiple dropdown

### LINKED ISSUE #3040: Value of multiple dropdown is directly mutated
### Overview of the problem

When using `<b-dropdown mulitple :value="myArray" />` (without `v-model`) the array value is directly mutated by the Dropdown component.

**Buefy** version: 0.9.4
**Vuejs** version: 2.6.11
**OS/Browser**: Any

### Steps to reproduce

1. Create a multiple dropdown like [this example](https://codesandbox.io/s/buefy-dropdown-bug-sit6n?file=/src/App.vue)
2. Select/deselect options

### Expected behavior

Since there is no `v-model`, the bound value (`myArray`) should not be modified.

### Actual behavior

The bound value (`myArray`) is directly modified by the component.

I think the issue comes from here: https://github.com/buefy/buefy/blob/7b1b739504076fa405dec7d4f30bdaf90a18eb11/src/components/dropdown/Dropdown.vue#L200
Instead of using `push()` and `splice()`, which change the content of the array, it should emit a new array (probably something like `[...this.selected, value]` and `this.selected.filter(v => v !== value)`).

If this is really a bug, I'll be glad to try to fix it!

VictorCazanave:
Fixes #3040

## Proposed Changes
- Create a new array instead of mutating `selected`
- ~~The 1st commit (https://github.com/buefy/buefy/pull/3047/commits/d92cf5b08d55fb54e398d6ee29299409df477743) contains a quick fix to remove the mutation of the array. However the `DropdownItem` status (`isActive`) may be desynchronized with the `value` of the `Dropdown` because `selected` is still modified.
That's why in the next commit (https://github.com/buefy/buefy/pull/3047/commits/8c9c02317c94a456d70aa9edad2002863e753850) I removed `selected` (which is actually unnecessary) and used only `value` prop.~~

Please let me know if you prefer to handle this in another way!

### DIFF
--- a/src/components/dropdown/Dropdown.spec.js
+++ b/src/components/dropdown/Dropdown.spec.js
@@ -1,9 +1,11 @@
 import { shallowMount } from '@vue/test-utils'
 import BDropdown from '@components/dropdown/Dropdown'
 
-let wrapper
-
 describe('BDropdown', () => {
+    const val1 = 'val1'
+    const val2 = 'val2'
+    let wrapper
+
     beforeEach(() => {
         wrapper = shallowMount(BDropdown, {
             slots: {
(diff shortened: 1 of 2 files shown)


jtommy:
@VictorCazanave the main idea is correct but not the implementation because it works only using two-way (`v-model`) or one-way binding (`:value`) and Buefy components should work without any binding too.

PS: please use `indexOf` because `includes` is not supported by IE

VictorCazanave:
@jtommy the issues are fixed! Let me know if you want me to change something else.

jtommy:
Thanks 👍

PR #3057, opened 2020-11-24
Fix #3056 - Reset time on carousel when clicking arrow

### LINKED ISSUE #3056: Carousel not displaying items for full length of time after using next/previous buttons or indicator
**Buefy** version: [0.9.4]
**Vuejs** version: [2.6.11]
**OS/Browser**: Windows 10 Pro build 1909 / Firefox 83.0 (64-bit)

### Description

Carousel doesn't reset timer after using the indicator or next/previous buttons.

### Expected behavior

As soon as you change the carousel item, the timer should be refreshed and the new item should be displayed for the full length of the interval (3600 ms or whatever is set).

### Actual behavior

The carousel item is displayed for the duration that was left over from the last item displayed. So if change an image in the carousel when it has 500ms left, the next image will change almost immediately.

harryJst:
Fixes #3056 

## Proposed Changes
- Add and call a `resetTimer` method to reset the timer when the carousel index is changed by clicking on the arrow

### DIFF
--- a/src/components/carousel/Carousel.vue
+++ b/src/components/carousel/Carousel.vue
@@ -265,6 +265,13 @@ export default {
                 this.timer = null
             }
         },
+        restartTimer() {
+            if (this.timer) {
+                clearInterval(this.timer)
+                this.timer = null
+            }
+            this.startTimer()
+        },
         checkPause() {
             if (this.pauseHover && this.autoplay) {
                 this.pauseTimer()
@@ -299,9 +306,11 @@ export default {
         },
         prev() {
             this.changeActive(this.activeChild - 1, -1)
+            this.restartTimer()
         },
         next() {
             this.changeActive(this.activeChild + 1, 1)
+            this.restartTimer()
         },
         // handle drag event
         dragStart(event) {
--- a/src/components/carousel/Carousel.vue
+++ b/src/components/carousel/Carousel.vue
@@ -265,7 +265,7 @@ export default {
                 this.timer = null
             }
         },
-        restartTimer(){
+        restartTimer() {
             this.pauseTimer()
             this.startTimer()
         },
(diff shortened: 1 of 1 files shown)


jtommy:
@Tofandel can you take a look?

Tofandel:
@harryJst there is some small issues in this PR, you're missing a check on the timer before clear interval, you should also use the startTimer method right after that

You also should use the changeActive method to call the reset timer and not next and prev
```
if (newIndex !== this.value) {
       this.$emit('input', newIndex)
}

Right after the emit 
```

harryJst:
Thanks for your feedback, much appreciated!
I have a two questions to clarify the change, only learning JS;
I see what you mean, but wouldn't expected behaviour be that the carousel would continue playing after a click not pause?
If I put it in change active won't it have unintended consequences in the other parts of the code where `changeActive` is called?

Tofandel:
@harryJst It would continue, the pause actually doesn't work as you would expect, it is actually stopping the timer and when unstopping resetting it, so effectively restarting the timer is pause + restart, if you look you will see the pause has the exact same code at the exception of the `this.pause = true` but you'll also see that when you restart the timer you have `this.pause = false`

> If I put it in change active won't it have unintended consequences in the other parts of the code where changeActive is called?

Correct, that's exactly the point since next and prev will change the value, so will a v-model change but you'd want the behavior to be the same, that is the timer to be reset

Tofandel (changes requested):
You misunderstood my comment re the pause timer, you shouldn't delete the restartTimer method rather just make it
```
restartTimer() {
  this.pauseTimer();
  this.startTimer();
} 
```

harryJst:
Hey @Tofandel, thanks for you time to review this. Seems like you've spent too much time on such a tiny PR, so sorry about that!

Tofandel (approved):
All good now

PR #3062, opened 2020-11-27
Document update: Change slot to v-slot

yuwu9145:
## Proposed Changes

- According to [Deprecated `slot` Attribute](https://vuejs.org/v2/guide/components-slots.html#Deprecated-Syntax), `slot` & `slot-scope` attributes should not be further encouraged from buefy doc
- Using `slot` & `slot-scope` could lead to difficult issue, e.g. issue [2944](https://github.com/buefy/buefy/issues/2944)
- Replace `slot` & `slot-scope` with `v-slot` or  its [shorthand](https://vuejs.org/v2/guide/components-slots.html#Named-Slots-Shorthand) in buefy document


jtommy:
Thanks

PR #3070, opened 2020-12-01
Fix b-carousel variable documentation 

cetooley:
The variable documentation for b-carousel includes a duplicate item and incorrect values when you look at the defaults in `/src/scss/components/_carousel.scss` - specifically:
- `$carousel-indicator-color` was repeated (one with an incorrect value and description) and,
- `$carousel-indicator-border` had the wrong default.

Very small changes, but one I noticed when looking at the documentation, and something I figured would be a quick fix! Thank you for buefy!


jtommy:
Thanks!

PR #3088, opened 2020-12-13
NavbarDropdown - fix of collapsible prop

### LINKED ISSUE #3086: Navbar Dropdown collapsible and hoverable combination
### Overview of the problem

**Buefy** version: 0.9.4
**Vuejs** version: 2.6.12
**OS/Browser**: MacOS/Chrome

### Description

If collapsible and hoverable properties are set on navbar dropdown, hovering is not working in desktop view anymore.
The documentation says the collapsible property should only affect the mobile view

### Expected behavior

In desktop view, navbar dropdown triggers by hover. In mobile view navbar dropdown is collapsed.

### Actual behavior
When hovering the navbar dropdown in desktop view, the dropdown is not opening.

https://codesandbox.io/s/buefy-navbar-dropdown-q6k3g

darotar:
Fixes #3086 

## Proposed Changes

- Added mobile view condition for 'collapsible' prop to not conflict with 'hoverable

### DIFF
--- a/src/components/navbar/NavbarDropdown.vue
+++ b/src/components/navbar/NavbarDropdown.vue
@@ -22,7 +22,7 @@
             <slot v-else name="label" />
         </a>
         <div
-            v-show="!collapsible || (collapsible && newActive)"
+            v-show="isShowDropDownItems"
             class="navbar-dropdown"
             :class="{
                 'is-right': right,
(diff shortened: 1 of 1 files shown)


jtommy:
It's not really correct because you are checking mobile device and not mobile resolution. Navbar CSS checks the with my media query

src/components/navbar/NavbarDropdown.vue
```diff
@@ -56,18 +58,31 @@ export default {
...
     data() {
         return {
             newActive: this.active,
             isHoverable: this.hoverable,
             _isNavbarDropdown: true // Used internally by NavbarItem
         }
     },
+
+    computed: {
+        isShowDropDownItems() {
```
jtommy, line 71: It's not really correct because you are checking mobile device and not mobile resolution. Navbar CSS checks the with my media query

jtommy (changes requested):

PR #3092, opened 2020-12-16
Add option to show all years at once

wkasem:
right now i am working on a project where user can enter his birthday and many other dates 

i am using date picker and i want to show all years user can select at once 
the current behavior right now  is : the list of years is getting updated each time the user select a year based on years range config
for the user this will be bad UX as he will think he can not select any of years not in the list which is not the case 

## Proposed Changes

- add an option to show all years in date picker at once not some of it


jtommy:
You can use years-range prop, did you try it ?

PR #3106, opened 2020-12-21
Reverting showX change when using hasModalCard prop

GerryWilko:
## Proposed Changes

Could we revert this change to hide the 'X' in the top right when we use the `has-modal-card` prop?

I'm not sure why this change was added (presumably as it was assumed that all modal cards would have their own close button?), but it now means we cannot use modal cards in our UI as we were using the 'X' button to close it.

You can still hide the 'X' using the `cancelOptions`.

### DIFF
--- a/src/components/modal/Modal.vue
+++ b/src/components/modal/Modal.vue
@@ -149,7 +149,7 @@ export default {
                 : this.canCancel
         },
         showX() {
-            return this.cancelOptions.indexOf('x') >= 0 && !this.hasModalCard
+            return this.cancelOptions.indexOf('x') >= 0
         },
         customStyle() {
             if (!this.fullScreen) {


jtommy:
You are right but it's a breaking change

GerryWilko:
@jtommy Yep agreed but it was an undocumented breaking change with v0.9.0 :)

jtommy:
I didn't remember it, when ?

GerryWilko:
https://github.com/buefy/buefy/commit/6f80982cdaf178c57ccfe231a303e93c2234c6f0

This commit for 0.9.0 added that extra isModalCard check.

GerryWilko:
@jtommy any chance this can get merged?

GerryWilko:
@jtommy Thanks! Any time frame for a release?

PR #3114, opened 2020-12-28
Append-to-body dropdown overlays the page when inactive

### LINKED ISSUE #3113: Append to body dropdown has positive z-index when inactive
### Overview of the problem

**Buefy** version: 0.9.4
**Vuejs** version: 2.6.12
**OS/Browser**: Chrome (but doesn't matter)

### Description

When using a dropdown with `append-to-body="true"`, the appended div has a positive z-index, making a part of the page unclickable as it is hidden behind this appended div.

### Steps to reproduce

1. Create a dropdown
2. Add the `append-to-body` prop to it
3. Open your page and you will see the div has no specified z-index.

### Expected behavior

The expectation would be that the div has no influence on the page while inactive.

### Actual behavior

The div has an unspecified z-index, letting it capture clicks while in an inactive state.

Dragoncall:
Fixes #3113 

## Proposed Changes

- The fix is very simple, as the body element is already updated by the `isActive` watcher. We only have to do a simple switch on `isActive`. If it is true, use `z-index: 0` else `z-index: -1`.

### DIFF
--- a/src/components/dropdown/Dropdown.vue
+++ b/src/components/dropdown/Dropdown.vue
@@ -336,6 +336,7 @@ export default {
                     parentNode.classList.forEach((item) => {
                         parent.classList.add(item)
                     })
+                    parentNode.style.zIndex = this.isActive ? '-1' : '0'
                 }
                 const rect = trigger.getBoundingClientRect()
                 let top = rect.top + window.scrollY


service-paradis:
I will revert this change in #3257
@Dragoncall Can you create a minimal codepen to reproduce the original issue so we can try to find another way to fix it?

jazzzz:
> I will revert this change in #3257 @Dragoncall Can you create a minimal codepen to reproduce the original issue so we can try to find another way to fix it?

I made a minimal Codepen there: https://codepen.io/jazzzz/pen/BaxobPp

As #3113 is still not fixed, maybe it should be reopened.

PR #3127, opened 2021-01-06
reset padding-left padding-right

Valar103769:
I used this component based on bulma, but the actual rendering is inconsistent and has been scolded by ui. Can I modify the style of bulma? Thank you, brother
In bulma, this value is 0.75, so continue to use it,?

Fixes #

## Proposed Changes

-
-
-

### DIFF
--- a/src/scss/components/_pagination.scss
+++ b/src/scss/components/_pagination.scss
@@ -1,8 +1,8 @@
 .pagination {
     .pagination-next,
     .pagination-previous {
-        padding-left: 0.25em;
-        padding-right: 0.25em;
+        padding-left: 0.75em;
+        padding-right: 0.75em;
         &.is-disabled {
             pointer-events: none;
             cursor: not-allowed;


service-paradis (approved):
You're right, it is now `0.75em` in Bulma. I agree that we should change it.
Maybe we could add a SCSS variable for it defaulting to `0.75em`? This way, it would be easy to customize if needed.

Valar103769:
> You're right, it is now `0.75em` in Bulma. I agree that we should change it.
> Maybe we could add a SCSS variable for it defaulting to `0.75em`? This way, it would be easy to customize if needed.

Valar103769:
> You're right, it is now `0.75em` in Bulma. I agree that we should change it.
> Maybe we could add a SCSS variable for it defaulting to `0.75em`? This way, it would be easy to customize if needed.

I think there are three situations. 
1. The person who finds the problem will use css to modify it to be consistent with bulma. 
2. You need to have your own custom style, which overrides the buefy style. 
3, people who still use buefy style (do you understand that they don't have high requirements for ui details? 
). 
For the first and second cases, even if the original value is modified directly, it will not have any impact on their project. 
In addition, the addition of variables, although it will not change the original thing, but it does not make much sense. 
Probably no one will use it. 
If possible, treat it as bug and modify the original value. 
If bulma and buefy are used together, it will result in two page inconsistencies. 
Very weird.

service-paradis:
> > You're right, it is now `0.75em` in Bulma. I agree that we should change it.
> > Maybe we could add a SCSS variable for it defaulting to `0.75em`? This way, it would be easy to customize if needed.
> 
> I think there are three situations.
> 
> 1. The person who finds the problem will use css to modify it to be consistent with bulma.
Having a SCSS variables will gave exactly the same results as your PR but would give the extra flexibility if people want to customize it. The variables will use the value same value as Bulma, so people wont have to modify anything to be consistent.

> 2. You need to have your own custom style, which overrides the buefy style.
I dont get your point here.

>    3, people who still use buefy style (do you understand that they don't have high requirements for ui details?
>    ).
I also dont get your point here.

>    For the first and second cases, even if the original value is modified directly, it will not have any impact on their project.
>    In addition, the addition of variables, although it will not change the original thing, but it does not make much sense.
>    Probably no one will use it.
>    If possible, treat it as bug and modify the original value.
>    If bulma and buefy are used together, it will result in two page inconsistencies.
>    Very weird.
There wont be any inconsistencies since the variable will use the same value as Bulma... Buefy already uses SCSS variables in other component to let people customize certain behavior.

I think you dont really understood the review. I'm ok with the change as-is, but I wanted to give people the flexibility to customize it (in fact, ideally, it would be a variable in Bulma).

Let see what @jtommy thinks of this.

jtommy:
I agree with @service-paradis, please open a PR to Bulma before and then we'll decide about it

service-paradis:
> I agree with @service-paradis, please open a PR to Bulma before and then we'll decide about it

I opened a PR (https://github.com/jgthms/bulma/pull/3238). If it is merged, we'll be able to use the Bulma variable instead of the hardcoded one.
In the mean time, do you want to merge this PR as-is since Bulma really uses `0.75em` now?

service-paradis:
PR (https://github.com/jgthms/bulma/pull/3238#event-4207406837) have been merged. Will wait the official release and will update this component then.

jtommy:
@service-paradis oh I merged it. Can you revert it ?

service-paradis:
It is ok for now! It solves the issue. 
I will open a new PRs that uses the variable once available.

PR #3128, opened 2021-01-06
Improve modal a11y

crowdcow-shae:
Adds aria-label prop to the modal component.

Currently the modal fails [jest-axe](https://github.com/nickcolley/jest-axe) accessibility tests because the rendered html does not have an aria-label attribute.

image


jtommy:
Thanks!

Issue #3135, opened 2021-01-08
Multiple NavBar fixed prop behavior


WTDuck:
### Overview of the problem

**Buefy** version: [0.9.4]
**Vuejs** version: [2.6.12]
**OS/Browser**: Google Chrome

### Description

In a multiple NavBar scenario on a single page.
If one of those is fixed (top or bottom), body classes like `has-navbar-fixed-top` are removed by other navbars when created since the bodyClass logic is included inside an immediate watcher.

I've prepared a PR to fix the behavior.

PR #3136, opened 2021-01-08
Multiple NavBar fixed prop behavior

WTDuck:
Fixes #
PR to fix bug reported on issue #3135

## Proposed Changes

- BodyClasses for Navbar are added on beforeMount or fixedTop/fixedBottom update
- BodyClasses for Navbar are only removed on fixedTop/fixedBottom update


jtommy:
Thanks!

PR #3148, opened 2021-01-15
data binds an `object[]`, there is a TypeError

Valar103769:
When data binds an array of objects, the value thrown by the select event should be an object, not an error caused by null, null when used, such as
```vue
<b-autocomplete @select="handleSelectChange"></b-autocomplete>
 handleSelectChange(option) {
          console.log('option',option) // null 
            this.url = option.url // Error in v-on handler: "TypeError: Cannot read property 'url' of null"
        }
```

Fixes #

## Proposed Changes

-
-
-

### DIFF
--- a/src/components/autocomplete/Autocomplete.vue
+++ b/src/components/autocomplete/Autocomplete.vue
@@ -351,7 +351,7 @@ export default {
         setSelected(option, closeDropdown = true, event = undefined) {
             if (option === undefined) return
 
-            this.selected = option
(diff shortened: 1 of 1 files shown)


jtommy:
Why?

Valar103769:
> Why? It's no a good way in my opinion because you have to check if it contains a specific field too

yes, but it's exist a bug actually

jtommy:
What bug?

PR #3170, opened 2021-01-30
Cast number type value to number when emit input event in BInput

### LINKED ISSUE #3109: b-numberinput incorrect handling of negative numbers between 0 and -1
### Overview of the problem

**Buefy** version: 0.9.4
**Vuejs** version: 2.6.11
**OS/Browser**: OSx / Firefox

### Description
It's impossible to enter negative numbers between 0 and -1 in b-numberinput.

### Steps to reproduce
1. Go to https://buefy.org/documentation/numberinput
2. Try entering -0.3, as soon as you hit the 0 key the negative sign will be removed and the number will be made positive

### Expected behavior
Starting a number with - should not mutate that number to positive.

### Actual behavior
Number becomes positive while it should be negative as the user started entering the number with a - sign

yuwu9145:
Fixes #3109

## Proposed Changes

- This issue is caused by this [vue behaviour](https://github.com/vuejs/vue/issues/11897)
- As per the answer in [vue behaviour](https://github.com/vuejs/vue/issues/11897), `.number` only work on native `v-model`, not on custom components. So `.number` is removed from `v-model` in `b-input`, instead `NumberInput` will be responsible to do number casting when it emits `@input`.

### DIFF
--- a/src/components/numberinput/Numberinput.vue
+++ b/src/components/numberinput/Numberinput.vue
@@ -28,7 +28,7 @@
         <b-input
             type="number"
             ref="input"
-            v-model.number="computedValue"
+            v-model="computedValue"
             v-bind="$attrs"
             :step="minStepNumber"
             :max="max"
@@ -153,7 +153,9 @@ export default {
                     }
                 }
                 this.newValue = newValue
-                this.$emit('input', newValue)
+                if (!isNaN(newValue) && newValue !== null && newValue !== '-0') {
+                    this.$emit('input', Number(newValue))
+                }
                 !this.isValid && this.$refs.input.checkHtml5Validity()
             }
         },
--- a/src/components/input/Input.vue
+++ b/src/components/input/Input.vue
@@ -114,11 +114,7 @@ export default {
             },
             set(value) {
                 this.newValue = value
-                if (isNaN(value)) {
-                    this.$emit('input', value)
-                } else {
-                    this.$emit('input', Number(value))
-                }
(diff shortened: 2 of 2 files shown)


src/components/numberinput/Numberinput.vue
```diff
@@ -153,7 +153,9 @@ export default {
                     }
                 }
                 this.newValue = newValue
-                this.$emit('input', newValue)
+                if (!isNaN(newValue) && newValue !== null && newValue !== '-0') {
```
yuwu9145, line 156: Purpose is only emit when `newValue` can be cast to a valid number, also `-0` will be cast to `0` without negative sign which will cause input becomes 0, so this case is ignored here.

yuwu9145:
Hi @jtommy , sorry for bothering. Just want to reach out to see if there is an ETA to review this pull request? :)

jtommy:
@yuwu9145 i don't understand -0 to be honest

yuwu9145:
> @yuwu9145 i don't understand -0 to be honest

When it comes to `v-model`, if `value` part is `-0`, it is a legitimate number `Number(-1) === -1`. However, Vue will set convert any number to string before update the `input` dom using `String(value)`. Then `String(-0) === '0'`. So `-0` should be specially treated

yuwu9145:
@jtommy Here is a [demo link](https://codepen.io/yuchaosydney/pen/poNREOE) that simulates this `-0` issue with Vue only

PR #3181, opened 2021-02-04
Split ConfigComponent into a dedicated entry

### LINKED ISSUE #3124: Adding ConfigProgrammatic reverses bundle size gains
### Overview of the problem

**Buefy** version: 0.9.4
**Vuejs** version: 2.6.11

### Description

Changing my project to include individual components instead of the full bundle leads to the vendors-js bundle size being reduced from 463.9 KiB down to 258.8 KiB. 

But this is without registering the font-awesome icon pack. It appears that using individual components, I must use `ConfigProgrammatic` to do that (see [Constructor options](https://buefy.org/documentation/constructor-options)). The problem is that when I set options using `ConfigProgrammatic` it seems to undo the bundle size gains even when specifying an empty options object:

```
ConfigProgrammatic.setOptions({ });
Vue.use(Table);
Vue.use(Input);
```

It brings the vendors-js bundle size back up to 467.1 KiB, which is actually a bit worse than including the full bundle!

### Steps to reproduce

1. Create a new Vue 2 app using Vue CLI.
1. Import the buefy package: `npm install buefy --save`
1. Import individual components as on the ["Getting Started" page](https://buefy.org/documentation/start). My main.ts looks as follows:
```
import Vue from "vue";
import App from "./App.vue";
import router from "./router";
import { Table, Input } from "buefy";
import "buefy/dist/buefy.css";

Vue.config.productionTip = false;

Vue.use(Table);
Vue.use(Input);

new Vue({
  router,
  render: h => h(App)
}).$mount("#app");
```
1. Run a build: `npm run build`
1. The reported vendors-js bundle size is relatively small
1. Add `ConfigProgrammatic.setOptions({ });` before the `Vue.use()` statements.
1. Run a build: `npm run build`

### Expected behavior

The vendors-js bundle size must not increase significantly after adding `ConfigProgrammatic`

### Actual behavior

The vendors-js bundle size explodes after adding `ConfigProgrammatic`

MartinodF:
Previously, ConfigComponent would be inlined into index.js, causing it to capture every component imported above. This meant that importing ConfigComponent/ConfigProgrammatic would break tree shaking.

Fixes #3124 

## Proposed Changes

- Add a new entry to rollup.config.js
- Restore tree shaking 🙌

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -30,6 +30,7 @@ const components = fs
 const entries = {
     'index': './src/index.js',
     'helpers': './src/utils/helpers.js',
+    'config': './src/utils/ConfigComponent.js',
     ...components.reduce((obj, name) => {
         obj[name] = (baseFolder + componentsFolder + name)
         return obj


jtommy:
Thanks!!

PR #3183, opened 2021-02-05
Improve usage example of Carousel List with Card

509dave16:
## Issue
When using a Carousel List with a custom template that fills the `item` slot, we were running into rendering issues where the product title on our card did not correspond with the image.

## Proposed Changes
Wrap contents of `template` with `b-carousel-list-item` component.


509dave16:
@jtommy We need to revert this documentation change. Only works in `development` builds. Does not work in `production` builds. My deepest apologies for the error. Rendering error was solved another way.

PR #3188, opened 2021-02-10
Fix tooltip blinking in table with sticky header

vasilyz:
## Proposed Changes

- Set `z-index: 200` for: ` .b-table .table-wrapper.has-sticky-header tr:first-child th`

This removes tooltip blinking when using b-table with sticky-header prop.


jtommy:
It's an issue of tooltip using append-to-body, we need to fix it and not the table

PR #3203, opened 2021-02-18
Improve taginput usage with maxtags == 1

apauly:
## Proposed Changes

This two little changes will improve the behaviour of the Taginput component in cases where `maxtags` is set to `1`:

- Show the input so the user knows that the field is editable
- Replace current tag when a user selects a tag. UX is comparable to a good old classic ´select´ where users don't have to remove their choice before they can select something else.

I didn't take a look at the tests yet, but I'd be happy to do so if you think this feature makes sense at all.

### DIFF
--- a/src/components/taginput/Taginput.vue
+++ b/src/components/taginput/Taginput.vue
@@ -231,7 +231,7 @@ export default {
          * Show the input field if a maxtags hasn't been set or reached.
          */
         hasInput() {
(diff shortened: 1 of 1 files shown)


jtommy:
Your improvement is correct but I don't know the reason to use it only for one tags

apauly:
For a single tag, the users intention is quite clear: replace that single tag. 
For multiple tags, we'd have to assume which tag to replace.

PR #3212, opened 2021-02-22
Fixing tooltip problem when both values 'inside' and 'outside' are included in autoClose property

### LINKED ISSUE #3195: Problem with autoClose property of tooltips
### Overview of the problem
**nuxt-buefy** version: 0.4.4
**Vuejs** version: [2.6.12]
**OS/Browser**:
Windows/Chrome  88.0.4324.150 (latest)

### Description
I am having a problem when trying to use the autoClose property of the tooltips. 
```html 
<b-tooltip
            type="is-white"
            :append-to-body="false"
            :triggers="['click']"
            :auto-close="['outside', 'escape', 'inside']"
            position="is-left"
          >
           <b-button icon-left="menu" type="is-white" />
</b-tooltip>```
<!--Description of the bug-->
In my app I need the tootip to be closed both when the user click outside or inside it. 
When I specify both options, it only closes when I click outside. 
If I remove the outside value from the array, it closes when I click inside.   
Basically the problem is that when I specify both options, it only takes the outisde value

### Steps to reproduce

1. Create a tooltip and add both value to the autoClose property
2. Try to close it, you will see it won't close when clicking inside
3. Remove outside property and it will close inside now

### Expected behavior
I expect it to work when specifying both properties
<!--What you expected to happen-->

elC0mpa:
Fixes #3195 

## Proposed Changes

- Now tooltip will work as expected when both values 'inside' and 'outside' are included in autoClose property


jtommy:
Thanks

PR #3214, opened 2021-02-28
Remove FOE-Tools from expo

Heziode:
I removed FOE-Tools since I (creator of FOE-Tools) do not use Bulma and Buefy anymore.


jtommy:
Oh ok.. i'm sorry to know it :(

Heziode:
@jtommy 

There are many reason to this change.

First I completely adhered to the utility class philosophy.

Then, I encounter some problems using Bulma/Buefy:
-  the customization can create side effect
- Having a light/dark theme is very heavy. I almost had around 2mo of CSS. After I discovered PurgeCSS I always reached more than 600ko. Now with tailwind, I have less than 100ko of CSS.

PR #3217, opened 2021-03-03
Update content slot and default slot ordering

schtr4jh:
HTML doesn't show if you put the `default` slot first. Should be also fixed in CodePen: https://buefy.org/documentation/tooltip/#custom-content

### DIFF
--- a/docs/pages/components/tooltip/examples/ExCustom.vue
+++ b/docs/pages/components/tooltip/examples/ExCustom.vue
@@ -1,10 +1,10 @@
 <template>
     <section class="b-tooltips">
         <b-tooltip position="is-bottom" multilined>
-            <b-button label="Html Content" type="is-primary" />
             <template v-slot:content>
                 <b>Lorem ipsum dolor sit amet</b>, consectetur warning elit. <i>Fusce id fermentum quam</i>.
             </template>
+            <b-button label="Html Content" type="is-primary" />
         </b-tooltip>
 
         <b-tooltip type="is-light" :triggers="['click']" :auto-close="['outside', 'escape']">


jtommy:
The problem is the auto close tag of button component they doesn't work on codepen. You can fix it adding `<b-button>`

schtr4jh:
Here's an example with tag closed: [CodePen](https://codepen.io/schtr4jh/pen/vYyrVxV?editors=1010) - first popup doesn't show the content while second does.

jtommy:
You are closing </button>

schtr4jh:
Updated with 4 [https://codepen.io/schtr4jh/pen/vYyrVxV?editors=1010](examples):
- 1. - default-slot + closed button - doesn't work
- 2. - content-slot + closed button - works
- 3. - default-slot + auto-closed button - doesn't work
- 4. - content-slot + auto-closed button - works

jtommy:
You are wrong, please check your code.

`<b-button label="Html Content" type="is-primary"></button>`

service-paradis:
In fact, you are both (kind of) right.

You are right @schtr4jh. There are some issues when opening Codepen examples from the documentation.
But, the issue is not related to the order of elements. As @jtommy said, we would need to update the examples without using self closing tags in order to make them works as expected on Codepen.

So, in this specific case, the correct fix to make it works would be the following code:
```
<b-tooltip position="is-bottom" multilined>
  <b-button label="Html Content" type="is-primary"></b-button>
  <template v-slot:content>
    <b>Lorem ipsum dolor sit amet</b>, consectetur warning elit. <i>Fusce id fermentum quam</i>.
  </template>
</b-tooltip>
```

It looks like the first example in https://github.com/buefy/buefy/pull/3217#issuecomment-790115251. You mentioned it was not working while, in fact, it is.

PR #3223, opened 2021-03-08
Suggestion: row groups

c0bra:
# Suggestion: Row groups

Ability to group rows by a certain column and provide visual "group headers".  Here's a screenshot:

image

Our business has a case where we need row groups and we'd like to stick with Buefy components if possible. This PR just represents a basic test to see if it could work, and it seems to. There would need to be more work to make sure it performs as expected with sorting, filtering, responsiveness, etc. I'm just opening this for feedback.

Thanks!


jtommy:
It's a great feature!! @c0bra

c0bra:
- [ ] Allow deep property bindings
- [ ] Extract row groups from regular sorting
- [ ] Fix colspan
- [ ] Make sure row groups work with filtering

service-paradis:
Nice work @c0bra 💯 
Will mark this as a draft while you're still working on it!

wesdevpro:
@c0bra what is the status of this PR?

c0bra:
@wesdevpro our business pivoted away from Vue so I was not able to pursue working on it.

wesdevpro:
> @wesdevpro our business pivoted away from Vue so I was not able to pursue working on it.

Could you fill me in on what this draft was left at and what the next steps were to complete this PR?

PR #3231, opened 2021-03-14
fix imports of bulma files

jojomak13:
I face this problem and it fixed after update those paths

Fixes #

## Proposed Changes

- fix **bulma**  `functions` path
- fix **bulma** `initial-variables`  path
- fix **bulma** main path


service-paradis:
Hum, the problem you faced were probably caused by another configuration issue or something. Double `/` and adding `.sass` unnecessarily is surely not a standard way to includes those files.
If you have some problem with the actual release, you can open a new issue with details of what is happening so we can try to understand and help you with it!

jojomak13:
@service-paradis could you check the **Issue** #3242

PR #3240, opened 2021-03-18
Modify DialogProgrammatic to be able to recieve Number.

YukiKAJIHARA:
When I pass number as argument to DialogProgrammatic, it shows nothing.
It can confuse someone else.
So I change DialogProgrammatic to be able to recieve Number.
```
import { DialogProgrammatic as Dialog } from 'buefy';

Dialog.alert(1); // Nothing is shown.
Dialog.alert("1"); // 1 is shown
Dialog.alert((1).toString()); // 1 is shown
```

Fixes #

## Proposed Changes

-
-
-

### DIFF
--- a/src/components/dialog/index.js
+++ b/src/components/dialog/index.js
@@ -38,6 +38,10 @@ const DialogProgrammatic = {
             params = {
                 message: params
             }
+        } else if (typeof params === 'number') {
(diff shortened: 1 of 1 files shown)


jtommy:
It accepts only a string, you should pass it

PR #3245, opened 2021-03-25
docs: add block for displaying empty table content

callahanrts:
## Proposed Changes

The [Sandbox with custom template](https://buefy.org/documentation/table/#sandbox-with-custom-template) example is missing the empty template. This adds a div that displays "no records" in the table when empty is checked.


service-paradis (approved):
Thank you @callahanrts!

PR #3267, opened 2021-04-06
Add mouse click event to b-table row click

tadams10:
## Proposed Changes
-Simply add mouse click event to the b-table row click. This will allow users to create more powerful extensions such as multi select via ctrl or shift click.


tadams10:
This would be a breaking change for anyone currently using the click event. Not sure what your preferred approach for that would be

jtommy:
@tadams10 it's a breaking change, we cannot accept it at the moment

wesdevpro:
https://github.com/buefy/buefy/discussions/3973

PR #3285, opened 2021-04-16
Feature 3269 transition appear

### LINKED ISSUE #3269: Support transition `appear`
### Support transition `appear` in components that use TabbedChildMixin

Add a propery `appear` or alternatively `animateInitially` to TabbedChildMixin to support the [same behavior as vue](https://vuejs.org/v2/guide/transitions.html#Transitions-on-Initial-Render).

Components that use TabbedChildMixin (e.g TabItem and StepItem) can be animated and are animated by default.  However, there is no way to control the initial animation behavior with the current API.  Also many people use Vue SSR and/or nuxt.js and because of the way Vue transition works, these components do not render on initial load.  See this question on SO: https://stackoverflow.com/questions/59627195/vue-js-transition-doesnt-apply-on-page-first-load

jermel:
Fixes #3269

## Proposed Changes

- Add `appear` property to TabbedMixin to support Vue's [appear - ie. transition on initial render](https://vuejs.org/v2/guide/transitions.html#Transitions-on-Initial-Render) feature
- Allows these components to render initially when SSR
- Backwards compatible

### DIFF
--- a/docs/pages/components/steps/api/steps.js
+++ b/docs/pages/components/steps/api/steps.js
@@ -17,6 +17,13 @@ export default [
                 default: '<code>true</code>'
             },
             {
+                name: '<code>appear</code>',
+                description: 'Apply animation on the initial render',
+                type: 'Boolean',
+                values: '—',
+                default: '<code>undefined</code>'
+            },
+            {
                 name: '<code>animation</code>',
                 description: 'Custom animation (transition name)',
                 type: 'String',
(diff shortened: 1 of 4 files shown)


jtommy:
Ok but better use a generic name as prop and no default value (only by global config)

jermel:
> Ok but better use a generic name as prop and no default value (only by global config)

Removed default and updated name to `animateInitially` 👍

Issue #2529, opened 2020-05-21
b-table's b-checkbox 100% broken without css imported


dgattey:
### Overview of the problem

**Buefy** version: 0.18.18 and 0.8.19
**Vuejs** version: 2.5.22 and 2.6.11 (both tested)
**OS/Browser**: Chrome, Firefox on macOS

### Description

Unless you import buefy's css file (`buefy/dist/buefy.css`), a `b-table` with `checkable=true` doesn't reactively emit `@check` except when clicking on the checkbox in the table header. In other words, checkboxes are broken on tables unless you import the CSS. This is because these lines have the effect of 1) *hiding* the native input and putting a different element there instead that acts as the click target.

Buefy should always work without importing styles. Totally breaks checkboxes on tables here.

```scss
.b-checkbox.checkbox {
  outline: none;
  display: inline-flex;
  align-items: center;
}
```
and
```scss
.b-checkbox.checkbox input[type=checkbox] {
  position: absolute;
  left: 0;
  opacity: 0;
  outline: none;
  z-index: -1;
}

.b-checkbox.checkbox input[type=checkbox] + .check {
  width: 1.25em;
  height: 1.25em;
  flex-shrink: 0;
  border-radius: 4px;
  border: 2px solid #7a7a7a;
  transition: background 150ms ease-out;
  background: transparent;
}
```

### Steps to reproduce

1. Import and use any b-table with any of the supported methods
2. Set `checkable=true` and `checked-rows=checkedRowsArray` and `@check=someMethod`
3. Make sure Buefy's CSS isn't imported
4. Try clicking on any of the checkboxes in any of the rows. While the checkbox itself shows a check or not, `checkedRowsArray` never gets updated, and `someMethod` never gets called.

### Expected behavior

1. `@check` should trigger reactively without styles imported
2. `checked-rows` should update without styles imported (both with and without the `sync` modifier)

### Actual behavior

Neither one of those things happened - checking the checkboxes only changed the UI, not the backing data.

Thanks for taking a look!

service-paradis:
Buefy is a UI component library. Styles are needed for most of its components. Why it should work without it?

markovejnovic:
Yes, I seem to have broken the tests in PR#3330 -- I am sorry for that. Has it been fixed? If not, I can dedicate time to look into it.

PR #3330, opened 2021-05-13
Resolved issue 3329 by disabling autocompletion on b-table checkboxes

### LINKED ISSUE #3329: BTable Checked State is cleared on Back navigation but the Checkbox is visually checked.
### Overview of the problem

**Buefy** version: [0.9.7]
**Vuejs** version: [2.5.17]
**OS/Browser**: Arch Linux / Chromium 90.0.4430.85 (Official Build) Arch Linux (64-bit)

### Description

In an enterprise application that I am working on I have noticed that Checkboxes on the Table component have erroneous output. Upon selecting some rows using the Checkboxes, navigating to a link and then navigating back, the Checkboxes that were previously selected keep their state, however, the `checked` state is empty.

[This article](https://stackoverflow.com/questions/299811/why-does-the-checkbox-stay-checked-when-reloading-the-page) shows the issue is not of Buefy but autocompletion.

### Steps to reproduce

1. Select a couple items.
After selecting the checkboxes:
image

This is as expected.

2. Visit a link on the same page and navigate back to the page with the table.

After navigating to a link and coming back to the same page, the state is cleared (`checkedRows.size === 0`), but the checkboxes remain visible.

image

### Expected behavior

Either the state is updated on Back navigation or the checkboxes are not ticked (which I believe is easier to implement).

I will be making a PR implementing the latter solution.

markovejnovic:
Fixes #3329

## Proposed Changes

- I propose `Checkbox.vue` has an autocomplete prop which is, by default, [what it is in HTML -- on](https://www.w3schools.com/tags/att_input_autocomplete.asp), but can be manually overriden to `off`, if necessary.
- I propose `Table.vue` uses `autocomplete=off` so that back-navigation can operate as expected.

Thank you for considering!

### DIFF
--- a/src/components/checkbox/Checkbox.vue
+++ b/src/components/checkbox/Checkbox.vue
@@ -12,6 +12,7 @@
             type="checkbox"
             ref="input"
             @click.stop
+            :autocomplete="autocomplete"
             :disabled="disabled"
             :required="required"
             :name="name"
@@ -38,6 +39,10 @@ export default {
         falseValue: {
             type: [String, Number, Boolean, Function, Object, Array],
             default: false
+        },
+        autocomplete: {
+            type: String,
+            default: 'on'
         }
     }
 }
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -59,6 +59,7 @@
                             v-if="checkable && checkboxPosition === 'left'">
                             <template v-if="headerCheckable">
                                 <b-checkbox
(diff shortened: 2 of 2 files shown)


Raljuk:
Hello! You broke tests with your commit =) Fix it please

markovejnovic:
> Hello! You broke tests with your commit =) Fix it please

Yes, I am sorry -- I just saw your message. Apparently issue #2529 references this exact problem.

PR #3349, opened 2021-05-28
Dialog Loading and Disable Options

### LINKED ISSUE #3348: Dialog Loading/Disabled Option
### Description

Currently I did not find any option to control the `dialog` button status. Specifically, I did not find a way to set the `input` field and the buttons of a prompt dialog to either `loading` or `disabled`.

I think this feature is quite useful for `dialog`s that ask for user input which eventually gets processed by a server. It would be, in my opinion, a nice quality-of-life feature for users to clearly know that their data is being processed (and also it would prevent them from accidentally making too many requests that they may not want to do).

### PR

I will be making a PR implementing this feature.

Thank you for considering! :)

markovejnovic:
Fixes #3348 

## Proposed Changes
I have implemented the following changes:
- I've used `b-button` instead of `button` in `Dialog` as this enables easy and convenient usage of the `disabled` and `loading` props. The `disabled` attribute is present in the HTML `button`, but I found no good reason for re-implementing the `loading` feature.
- I've added a `promptStatus` datum which delegates what the input status should be `loading`, `disabled` or any other (in which case the component is treated as enabled). I've added a `computed` value for `inputDisabled` which disables the form input based if `promptStatus === 'disabled' || promptStatus === 'loading'`.
- I've set the `confirm` button to the appropriate value based off `promptStatus`.
- I've added a prop to `Dialog` named `initialPromptStatus` which sets the initial value of the `promptStatus`, by default to an enabled value.
- I've added a `setStatus` method which appropriately sets `promptStatus`.

## Discussion

There are a couple of design decisions I've made that might be relevant:
- I did not disable the `Cancel` button even if `promptStatus` is enabled. I think it's unreasonable to not let the user cancel their selection even after they've made the request. If I did, implementing undo functionality, for example, would be impossible. This way, the user can still click cancel and the `onCancel` handler can appropriately correct their error. I believe this is reasonable, but if there are differing opinions, I'm all ears.

## Bookkeeping

Sorry for the lots of commits, but from what I've gathered you squash PR commits into one so it shouldn't be too big of an issue. If it is, please notify me I will handle them manually.

### DIFF
--- a/docs/pages/components/carousel/api/carousel.js
+++ b/docs/pages/components/carousel/api/carousel.js
@@ -298,7 +298,7 @@ export default [
                 description: 'count of items to showed per view (support a decimal).',
                 type: 'Number',
                 values: '—',
-                default: '<code>true</code>'
+                default: '<code>4</code>'
             },
             {
                 name: '<code>items-to-list</code>',
--- a/package-lock.json
+++ b/package-lock.json
@@ -2790,7 +2790,8 @@
     "ansi-regex": {
       "version": "2.1.1",
       "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
-      "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8="
+      "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
+      "dev": true
     },
     "ansi-styles": {
       "version": "3.2.1",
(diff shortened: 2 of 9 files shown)


jtommy:
I understand your need but it looks a forced way to show loading button.
In my opinion you should create a own modal and the dialog; anyway I'll think a better way to meet your needs

markovejnovic:
> I understand your need but it looks a forced way to show loading button.
> In my opinion you should create a own modal and the dialog; anyway I'll think a better way to meet your needs

Yep, makes sense. I'll be closing the PR.

PR #3362, opened 2021-06-08
Fix for "Table component passing $attrs (such as ID) to pagination, instead of table div" (issue #2946)

### LINKED ISSUE #2946: Table component passing $attrs (such as ID) to pagination, instead of table div
### Overview of the problem

**Buefy** version:0.9.3
**Vuejs** version: 2.6.12
**OS/Browser**: MacOs Catalina/ All browsers

### Description

When passing an attribute, such as `id` to b-table, it gets  tacked onto pagination, instead of the table div created by the component. I believe the issue lies here: 
``` <b-table-pagination
                    **v-bind="$attrs"**
                    :per-page="perPage"
                    :paginated="paginated"
                    :icon-pack="iconPack"
                    :total="newDataTotal"
                    :current-page.sync="newCurrentPage"
                    @page-change="(event) => $emit('page-change', event)"
                >
```

### Steps to reproduce

1 - add an ID or other attribute to b-table
2 - inspect the table with dev tools and notice the ID added is not on `div class='b-table` anymore, but instead in div class='top-level'` 

### Expected behavior

Similarly to how it was before the 0.9.0 upgrade, the ID passed to the b-table should be on the div surrounding the table, not the pagination div.

### Actual behavior

Id is on pagination div.

Numenorian:
Fixes #2946
## Proposed Changes

- in `src/components/table/Table.vue`, I removed the `v-bind="$attrs"` statement from both `<b-table-pagination>` components and added it to the main `<div class="b-table">`
- I'm not sure why the `v-bind="$attrs"` statements were on the pagination components in the first place, so I'm not sure if this breaks any other new features or intended functionality. If so, let me know and I can take another look at it.
- also a minor typo fix in `README.md`


jtommy:
I don't know if there are missing prop about pagination but it would be a breaking change

Numenorian:
I'm sorry, I'm not sure I follow. Have there been new things added in Buefy 0.9.0+ that depend on HTML attributes from `<b-table>` being passed on to the `<b-table-pagination>` component(s)?

jtommy:
At the moment v-bind is on pagination component so moving it to table it might create problems or breaking changes

wesdevpro:
This PR has been migrated to a discussion for later reference https://github.com/buefy/buefy/discussions/3974

PR #3364, opened 2021-06-09
These elements are being flagged by the accessibility scanner SortSite.

eveenendaal:
Fixes #
These elements are being flagged by the accessibility scanner SortSite.

Proposed Changes

Adding button role and tab index to autocomplete's dropdown item to make it keyboard accessible
Adding tab index to the dropdown trigger. The div already has a role of a button and the tab index will allow for it to be keyboard accessible.


jtommy:
@eveenendaal thanks, can you run test in order to update snapshots?

PR #3367, opened 2021-06-15
Added a delay prop to sidebar to expand with delay

holtolee:
## Proposed Changes

- To avoid the user to open the sidebar when he doesn't want to, i added a `delay` prop to the sidebar component.
Like the tooltip delay prop put a value in ms, it works only on reduced mode and doesn't need to enable "Expand on hover"

Preview: 
delaySidebar


jtommy:
Thanks

PR #3369, opened 2021-06-15
3305 b-carousel Drag Event is not working

### LINKED ISSUE #3305: b-carousel Drag Event is not working
### Overview of the problem

`b-carousel`'s drag event to update slides either to next or previous does not seems to be changing the slides.

**Buefy** version: [0.9.7]
**Vuejs** version: [2.6.1]
**OS/Browser**: Google Chrome , Google Chrome mobile

### Description

`hasDrag` prop should be `true` by default. Any draggable device should be able to update the active `carousel-item` except `hasDrag` is being set to `false`. 
In the Carousel [documentation](https://buefy.org/documentation/carousel), the carousels in Default, Custom(with Drag Event enabled), Arrow, Progress and Indicator sections can't seems to be dragged. Weird is that, the Carousels after that ([Custom indicator](https://buefy.org/documentation/carousel/#custom) onwards) does not have this issue. 

My apologies, I can't seems to find the pattern but only can tell which are the working ones, which are not.

### Steps to reproduce

1. Navigate to https://buefy.org/documentation/carousel
2. Attempt any drag event on the default Carousel (first Carousel)

### Expected behavior

Items in the default Carousel should be changed, either to the previous or next.

### Actual behavior

Carousel does not update the active carousel item, either to the previous or next depending on thee drag direction.

holtolee:
Fixes #3305

## Proposed Changes

- Removed `!event.target.draggable` condition, it fixed the issue but i didn't find out why there was this condition (maybe deprecated ?)

Edit: in my project it was working (without this fix) inside a `b-modal` with just the `:autoplay="false"` 🤔


src/components/carousel/Carousel.vue
```diff
@@ -310,8 +310,7 @@ export default {
         },
         // handle drag event
         dragStart(event) {
-            if (!this.hasDrag ||
-                !event.target.draggable) return
```
service-paradis, line 314: Without this condition, form elements (input, select, textarea, ...) does not work as expected inside the carousel.

holtolee: so maybe we need to play with `pointer-events`  of childs elements, i will check this later (had similar problem with child elements of custom drag n drop component).

holtolee:
@service-paradis I have updated my code and I think it's working fine now

- Drag only work on the background or first child in the slot (checked with classlist of `parentElement` of the drag target)
- Drag is disabled on every child element so form elements work as intended
- Ci Test failed but i don't really know how to adapt it

BraedenKilburn:
Can we find a fix to this? I'm still having this issues more than a year later.

jtommy:
@BraedenKilburn what's carousel-body element?

BraedenKilburn:
> @BraedenKilburn what's carousel-body element?

That was an example. It would be whatever the item is in that carousel, you can add the "draggable" attribute to the element you're displaying and it will allow for dragging through your carousel.

wesdevpro:
Closing as `stale` and found a work around! See #3305

Issue #3351, opened 2021-05-28
Various Sass Deprecation Warning - When Setting Custom Themes


SomethingNew71:
### Overview of the problem

**Buefy** version: 0.9.7
**Vuejs** version: 2.6.12
**OS/Browser**: N/A

### Description
With the use of the current versions of sass and importing buefy for custom theming as described in the docs various deprecation warnings are now present during application compile. According to Sass docs removal of deprecated sass functions is required to resolve. (which I believe will require Buefy to update their Sass version too).

Details about breaking change - https://sass-lang.com/documentation/breaking-changes/slash-div

### Steps to reproduce
1. Import vuefy styles as follows:
<img width="766" alt="Screen Shot 2021-05-28 at 4 09 57 PM" src="https://user-images.githubusercontent.com/4572799/120037267-59c2b480-bfcf-11eb-83dc-d4407230a9fe.png">

2. Build application

### Expected behavior

No deprecation warnings when using the current version of sass.

### Actual behavior
Deprecation Error:
<img width="917" alt="Screen Shot 2021-05-28 at 4 08 14 PM" src="https://user-images.githubusercontent.com/4572799/120037304-6c3cee00-bfcf-11eb-9019-55b5a642404d.png">

service-paradis:
Buefy will need to change a few things for the next version, but we need a new version of Bulma that removes slash divisions. You can follow the discussion at https://github.com/jgthms/bulma/issues/3333

SomethingNew71:
@service-paradis thanks for the update. It didnt even dawn on me that this might be further upstream than buefy.

SomethingNew71:
Hey this isn't fully solved yet. While the ones caused by bulma are addressed, Buefy has some / division warnings in the functions util.

https://github.com/buefy/buefy/blob/dev/src/scss/utils/_functions.scss#L17

I am gonna toss up a PR later today to replace that.

PR #3375, opened 2021-06-19
Project Level `math.div` removal

### LINKED ISSUE #3374: DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.
$loops: ceil(str-length($svg)/$slice);

SomethingNew71:
Fixes  #3374 and #3351 

## Proposed Changes

- Update to new version of Sass
- Replace use of slash divide in the _functions.scss file
- Removes the unused legacy font-awesome package

## Notes:
There are still some upstream slash div issues from font awesome but dart sass has released an update that allows you to silence them. Unfortunately, you need the newest version of `sass-loader` in order to use it which isn't a quick upgrade for the buefy webpack build. 

https://github.com/webpack-contrib/sass-loader/issues/954


jtommy:
Thanks, I already fixed it (in an other way also)

SomethingNew71:
> Thanks, I already fixed it (in an other way also)

Ah awesome! Thank you. One thing I do think they are deprecating `divide()` in the upcoming versions too. So something to be on the look out for.

PR #3401, opened 2021-07-02
Avoid error when trying to read property of undefined

apauly:
## Proposed Fix
If the `dateSelected` is not set for a Datepicker with range selection, trying to select any date will cause a JS error (`this.dateSelected.length`). This PR fixes that and `focusedDateData` will not get changed when `dateSelected` is not present.


jtommy:
@apauly thanks

Issue #3185, opened 2021-02-07
B-autocomplete first value Erase/retype won't trigger the select event


4l1c30zz:
### Overview of the problem

**Buefy** version: [0.9.4]
**Vuejs** version: [2.6.12]
**OS/Browser**: Ubuntu 20.04 Firefox/Chrome

### Description

Hello , We are using the b-autocomplete with mostly predefined value,
On first interaction with the element (erasing the value with backspace or select all than type) Won't call the select event. 
From the first clear with button or selecting from the list and on everything works as expected.
### Steps to reproduce
1.Please Open the pen : https://codepen.io/f1na3/pen/BaQKGzE (I copied the scripts and styles from the template you provided +  added the minimal needed to reproduce the case)
2.Erase the value for the first time (with keyboard and not with erase button),
Or select all and type something else
### Expected behavior

I'm expecting that select event will be called from the first manual clear/ value retype
### Actual behavior

the ID won't change. and the select event is not called
After clear button press or after first select from list and on everything works as expected.(till page refresh, one time bug)

4l1c30zz:
@jtommy Thank you!
The solution provided solved my issue:)

PR #3403, opened 2021-07-03
Autocomplete enhancements and fixes for accessibility

### LINKED ISSUE #3357: Autocomplete with
### Overview of the problem

**Buefy** version: 0.9.7
**Vuejs** version: 2.6.13
**OS/Browser**: Windows 10, Firefox 89

### Description

An Autocomplete with `keep-first` will trigger `@select` for the currently selected element when clicking outside.

### Steps to reproduce

```vue
<template>
  <div id="app">
    <b-autocomplete
      :data="statuses"
      field="name"
      keep-first
      :open-on-focus="true"
      @select="selected = $event"
    >
      <template slot-scope="props">
        {{ props.option.name }}
      </template>
    </b-autocomplete>

    <div>
      Selected: {{selected}}
    </div>
  </div>
</template>

<script>
export default {
  data() {
    return {
      selected: null,
      statuses: [
        {
          status: "one",
          name: "One"
        },
        {
          status: "two",
          name: "Two"
        },
        {
          status: "three",
          name: "Three"
        },
        {
          status: "four",
          name: "Four"
        },
        {
          status: "five",
          name: "Five"
        }
      ]
    };
  }
};
</script>
```

Click on the Autocomplete, then click outside to close it.

### Expected behavior

`selected` stays `null`

### Actual behavior

`selected` becomes the last selected item.

elisamuelps:
# Autocomplete
## Bug Fixes
- Fixed Bug when pressing esc which briefly toggled/expanded the suggestion list
- Moved esc input event from key up to key down
- Use of keycodes to detect esc, tab and enter keys
## Accesibility
- Add aria-autocomplete to the input element as "list" or "both" depending if keep-first is enabled or not
- Add native-search prop which changes the input element's type to "search" or "text"

# Documentation
## Autocomplete
- Added example of Keep First and select on clicking outside

# Fixes
Fixes #3357 

## Proposed Changes
It is all on my changelog above

### DIFF
--- a/src/components/autocomplete/Autocomplete.vue
+++ b/src/components/autocomplete/Autocomplete.vue
@@ -387,19 +387,13 @@ export default {
         },
 
         keydown(event) {
-            const { key, which } = event // cannot destructure preventDefault (https://stackoverflow.com/a/49616808/2774496)
-
-            // Use Keycodes to detect input as it can cause adverse effects
-            // with other Keyboard layouts
-            // see https://css-tricks.com/snippets/javascript/javascript-keycodes/
-            // see also https://unixpapa.com/js/key.html
-            // test keycodes with https://www.w3.org/2002/09/tests/keys.html
+            const { key } = event // cannot destructure preventDefault (https://stackoverflow.com/a/49616808/2774496)
 
             // prevent emit submit event
-            if (which === 13 || key === 'Enter') event.preventDefault()
+            if (key === 'Enter') event.preventDefault()
 
             // Close dropdown on Tab & no hovered
-            if (which === 9 || which === 27 || key === 'Escape' || key === 'Tab') {
+            if (key === 'Escape' || key === 'Tab') {
                 event.preventDefault()
                 this.isActive = false
             }
@@ -409,7 +403,7 @@ export default {
                 if (key === ',') event.preventDefault()
 
                 // Close dropdown on select by Tab
-                const closeDropdown = !this.keepOpen || which === 9
+                const closeDropdown = !this.keepOpen || key === 'Tab'
                 this.setSelected(this.hovered, closeDropdown, event)
             }
         },
--- a/src/components/autocomplete/Autocomplete.vue
+++ b/src/components/autocomplete/Autocomplete.vue
@@ -396,11 +396,12 @@ export default {
             // test keycodes with https://www.w3.org/2002/09/tests/keys.html
 
             // prevent emit submit event
-            if (which === 13) event.preventDefault()
+            if (which === 13 || key === 'Enter') event.preventDefault()
+
             // Close dropdown on Tab & no hovered
-            if (which === 9 || which === 27) {
-                this.isActive = false
+            if (which === 9 || which === 27 || key === 'Escape' || key === 'Tab') {
                 event.preventDefault()
+                this.isActive = false
             }
             if (this.hovered === null) return
             if (this.confirmKeys.indexOf(key) >= 0) {
--- a/src/components/autocomplete/Autocomplete.vue
+++ b/src/components/autocomplete/Autocomplete.vue
@@ -3,7 +3,7 @@
         <b-input
             v-model="newValue"
             ref="input"
-            :type="newType"
+            :type="type"
             :size="size"
             :loading="loading"
             :rounded="rounded"
@@ -115,7 +115,6 @@ export default {
             type: String,
             default: 'value'
         },
-        nativeSearch: Boolean,
         keepFirst: Boolean,
         clearOnSelect: Boolean,
         openOnFocus: Boolean,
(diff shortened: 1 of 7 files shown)


elisamuelps:
Please check in the new documentation, I added Buefy 0.9.9 but I'm not sure if you will release it under 0.9.8.1 or 0.9.9

Glad to Help!

elisamuelps:
Just updated the jest snapshot for the autocomplete element since adding the accessibility functionality adds an extra aria element and also the esc key was not being triggered when using keycodes, so I made the element to detect both keycodes and key names

jtommy:
@elisamuelps 
why did you convert key to which ?

elisamuelps:
I did a mistake :( though that by using which it would help with different keyboard layouts, javascript is weird, but I'm reading that is being deprecated, I will solve it now. Hold on.

elisamuelps:
@jtommy itd all sorted, removed which from the pr, I tested the implementation and it performs as expected.

elisamuelps:
I have to amend the changelog above, going to type it below:

# Autocomplete
## Bug Fixes
- Fixed Bug when pressing Escape/Tab which briefly toggled/expanded the suggestion list
- Moved Escape input event from key up to key down
## Accessibility
- Add aria-autocomplete to the input element as "list" or "both" depending if keep-first is enabled or not
- Add native-search prop which changes the input element's type to "search" or "text"

# Documentation
## Autocomplete
- Added example of Keep First and select on clicking outside

jtommy:
@elisamuelps now it's ok, thanks.
What's the advantage to use search as input type ?

elisamuelps:
It is more about accessibility: line breaks are also automatically disabled when search is used instead of text and also it helps screen readers to identify the object. In fact, the autocomplete component is more of a search box than a text box. Although both look similar.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search

elisamuelps:
Thing is, because Vue forms use client side rendering, a 200 get request with a search query is not commonly performed which adds autocompletion of recent searches and I fear whether or not they clash with the list, if so, I can put the native html search form to use inside of the component. It will help people to be W3C compliant.

jtommy:
@elisamuelps We might add search type as default, what do you think?

elisamuelps:
That would be the best solution, we can get rid of one prop and treat the element as a native html search input which is more compliant with the W3C. If there isn't any reason to keep using text I'd say is good to make the change. I can make another commit later to reflect on the change.

jtommy:
It could be a breaking change.. but I don't like the new prop if search type is a good approach

elisamuelps:
Yeah me neither, I went originally with the prop idea because I didn't wanted to change the existing functionality, but if there is the possibility to switch the autocomplete input to search I will gladly add the implementation.

elisamuelps:
Hey @jtommy Been researching a bit, there is more into autocomplete elements than previously though, I additionally checked Google's implementation of their native input and to my wonder they use text instead of search. 

I found that we are going to be fine as long as we use the right aria elements, because there are more cons than pros of using the <datalist> and <options> elements. One of them is not being able to style those components and the other is not being able to provide a prop which is sad because you cannot add icons or any content inside of the list. In a resume: it is not worth doing datalist or input type search at all. 

The only thing we miss by not using the search input type is that the browser automatically saves your most recent searches first but the later can be done without any change to the component at all, that's what the slots are there for.

It is kind of sad because I wanted to improve the element and I ended up rolling back most changes, however, I did fixed #3185 and the library will benefit from it

I left also the type for those who still want to play around with the native input type element.

Hit me up if you need anything else, I have tested the changes and it is all solid now.

src/components/autocomplete/Autocomplete.vue
```diff
@@ -308,6 +310,12 @@ export default {
             this.$emit('input', value)
             // Check if selected is invalid
             const currentValue = this.getValue(this.selected)
+
+            // Fixes #3185
+            if (!currentValue) {
```
jtommy, line 315: Better a typeof undefined check and an else if ;)

elisamuelps: Hi @jtommy I have been busy with the release of a project :( but I will have a look tomorrow and submit a commit.

By the way, I found that strict type checking has some bugs with specific minifier settings, for example google closure compiler and terser with the option of booleans as integers will trigger errors with Buefy because of the strict type checking. Everything else runs as expected.

dauriata:
to fix one of the bug mentionned in #3469 (option still selected even if there is no match), I propose to modify the `selectFirstOption` like this in `Autocomplete.vue`: 

```
        /**
         * Select first option
         */
        selectFirstOption(computedData) {
            this.$nextTick(() => {
                if (computedData.some((element) => element.items && element.items.length)) {
                    // If has visible data or open on focus, keep updating the hovered
                    const option = computedData[0].items[0]
                    if (this.openOnFocus || (this.newValue !== '' && this.hovered !== option)) {
                        this.setHovered(option)
                    }
                } else {
                    this.setHovered(null)
                }
            })
        },
```

I am not opening a new PR cause I think it could fit well in this one. 
the previous test `if(element.length)` should mimic the `isEmpty` function, I guess it has not been updated when the Groups feature was introduced.

jtommy:
@dauriata open a PR please

PR #3421, opened 2021-07-15
add aria-labelledby prop to checkbox component

milad-mirkarimi:
Fixes #

## Proposed Changes

Adding aria-labelledby property to Checkbox component for accessibility


jtommy:
Thanks

Issue #3378, opened 2021-06-21
Not emit value when clear data from input number


MinhPhuPham:
### Overview of the problem

**Buefy** version: [0.9.4]
**Vuejs** version: [2.6.12]
**OS/Browser**: Chrome

### Description

- When i clear input => it not emit value empty? So how can i check it? 

### Steps to reproduce

- See on Codepen: https://codepen.io/minh-ph/pen/poeBOdQ?editors=1011

### Expected behavior

- I want it emit value back.

jtommy:
You are right it doesn't emit when value is empty

Issue #3059, opened 2020-11-25
b-numberinput doesn't recover from invalid value even after value is valid


navmed:
### Overview of the problem
Validation message doesn't go away even after the value is corrected using buttons

**Buefy** version: 0.9.4
**Vuejs** version: 2.6.12
**OS/Browser**: Windows/Chrome

### Description

My b-numberinput has a range of -3 to 3. If I enter -5 manually and blur, it shows me an error as expected. But, if I increase the value using the + button to -3 which is valid or any valid value, the validation message doesn't go away. The only way to make it go away it so focus on the text input and 

### Steps to reproduce

1. Enter -5 in the text input
2. Click outside the text input - validation message is shown
3. Click the + button 3 times to make the number -2

### Expected behavior

Validation message should disappear

### Actual behavior

Validation message remains unless the text input is focused and blurred

https://codepen.io/navmed/pen/poyxZgo?editors=1010

karoletrych:
Got similiar issue, when the input exceeds range specified by min max, the validation doesn't disappear.
https://codepen.io/morphix0/pen/pYYBgG?editors=1010

In my case it seems caused by additional element: <span class="centered-text">Number </span>, if i remove it, it works.

mateussilva92:
This probably caused by `b-field` when using `b-numberinput` along side other element inside that b-field.
Here is another codepen.

https://codepen.io/cpn20/pen/VwbbPYY

fabiofdsantos:
> This probably caused by `b-field` when using `b-numberinput` along side other element inside that b-field.
> Here is another codepen.
> 
> https://codepen.io/cpn20/pen/VwbbPYY

@jtommy @service-paradis Any idea? Thanks!

milad-mirkarimi:
> > This probably caused by `b-field` when using `b-numberinput` along side other element inside that b-field.
> > Here is another codepen.
> > https://codepen.io/cpn20/pen/VwbbPYY
> 
> @jtommy @service-paradis Any idea? Thanks!

Have you tried updating to the latest version (0.9.8)?

mateussilva92:
@milad-mirkarimi the last codepen provided is using the latest version of (0.9.8) and the issue is still replicable.

PR #3430, opened 2021-07-25
feat: improve number input validation

fabiofdsantos:
Fixes https://github.com/buefy/buefy/issues/3378 
Fixes https://github.com/buefy/buefy/issues/3059

## Proposed Changes

- Emit an input event if the model is `null`
- Force the `checkHtml5Validity()`


jtommy:
I don't think it's correct to force emit input event, there are internal validations to check

fabiofdsantos:
@jtommy I think now it's okay

PR #3440, opened 2021-08-01
feat: enable click event in notification

franciscoh017:
- Enable click event on notification and notificationProgramatic

Fixes #

## Proposed Changes

- As requested in [discord](https://discord.com/channels/370174661159419908/488689777743364096/860089083987361812) by `IceBlizz6`. Click event handler for the notification. For the notificationProgrammatic should do as with the `close` event, using `notif.$on('click', () => ... )`

### DIFF
--- a/src/scss/components/_notices.scss
+++ b/src/scss/components/_notices.scss
@@ -95,6 +95,7 @@ $snackbar-box-shadow: $notices-box-shadow !default;
         }
     }
     .notification {
+        pointer-events: auto;
         max-width: 600px;
     }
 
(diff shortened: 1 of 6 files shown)


src/scss/components/_notices.scss
```diff
@@ -95,6 +95,7 @@ $snackbar-box-shadow: $notices-box-shadow !default;
         }
     }
     .notification {
+        pointer-events: auto;
```
milad-mirkarimi, line 98: was the close button the default of target of pointer event?

franciscoh017: pointer events were not being triggered since .notification was inheriting pointer-events: none from .notice.

PR #3462, opened 2021-08-16
Fix PR#3454  with a name correction

dlamblin:
## Proposed Changes

- The example of horizontal input forms needed to be correctly named.
- I authored the incorrect name in PR#3454 and as @mouadTaoussi pointed out, it was a mistake.


jtommy:
Thanks

PR #3470, opened 2021-08-24
Fix autocomplete with keepFirst

### LINKED ISSUE #3469: Autocomplete: make option for keep-first selects hovered option on blur
### Overview of the problem

I was about to post an issue about autocomplete with keep-first that select hovered option on blur but I found out that it was intended in issue  #3118 and then commit ff7d574   ( original commit f268fd5 )

**Buefy** version: [0.9.8]
**Vuejs** version: [2.6.14]

### Description

In my opinion, when clicking outside or blur it is common UX behavior to cancel the current action. 
here is a common scenario:  if I look for a name in an autocomplete field and the dropdown makes me propositions but I don t find what I was looking for, I click outside or blur and nothing should be selected.
If I find what I was looking for, `keep-first` make it easy to just it "Enter" or "Tab" to select.

The doc for keep-first has not been updated with commit ff7d574  :
`keep-first : The first option will always be pre-selected (easier to just hit enter or tab)`

### Steps to reproduce

https://codepen.io/zouloufr/pen/bGRbMPj

### Expected behavior

I would like this to be a configurable option. I am open to discussion but in my opinion behavior before commit ff7d574 should be the default

### Actual behavior

keep-first selects hovered option on blur

dauriata:
Fixes #3469

## Proposed Changes

Fix some autocomplete behavior with the `keepFirst` option

- item preselected even if there is no match with search text
- item preselected even if the dropdown does not show ( no `openonfocus` and empty search text )

this PR would also works if the filter function return empty group arrays like : 
````
     [
                {
                    type: 'Fruit',
                    items: []
                },
                {
                    type: 'Vegetables',
                    items: ['Carrot', 'Broccoli', 'Cucumber', 'Onion']
                }
       ]
```

### DIFF
--- a/src/components/autocomplete/Autocomplete.vue
+++ b/src/components/autocomplete/Autocomplete.vue
@@ -388,11 +388,8 @@ export default {
                     (element) => element.items && element.items.length
                 )
                 if (nonEmptyElements.length) {
-                    // If has visible data or open on focus, keep updating the hovered
                     const option = nonEmptyElements[0].items[0]
-                    if (this.openOnFocus || (this.newValue !== '' && this.hovered !== option)) {
-                        this.setHovered(option)
-                    }
+                    this.setHovered(option)
                 } else {
                     this.setHovered(null)
                 }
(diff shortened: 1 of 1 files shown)


jtommy:
@fabiofdsantos can you review it?

PR #3472, opened 2021-08-24
Check that a detail row is visible before trying to close it

### LINKED ISSUE #3461: b-table closeDetailRow(row) closes the detail area of another row.
### Overwiew of the problem

buefy@0.9.8
vue@2.6.14
OS/Browser: any

### Description

Calling the closeDetailRow(obj) method closes an unrelated row, if obj represents a row that is NOT currently "expanded".

### Steps to reproduce

1. Have a working <b-table ref="mytable" detailed detail-key="id" ...
1. Open the detailed view of one or more rows.
1. From the parent component, call this.$refs.mytable.closeDetailRow(obj) on a row that is currently collapsed.

### Expected behaviour

Nothing should happen.

### Actual behaviour

The last opened row from step 2 above gets closed.

### Workaround

First, call the undocumented API method isVisibleDetailRow(obj) to check if the target row is really expanded. If false, do not call closeDetailRow(obj).

### Proposed fix

In the error situation, const i becomes -1 because indexOf() cannot find the index in the visibleDetailRows array
https://github.com/buefy/buefy/blob/7dbe191761bcc169e895ac42817a403a2c038f6a/src/components/table/Table.vue#L1139
The following line then removes the -1th (the last) element from the visibleDetailRows array.
https://github.com/buefy/buefy/blob/7dbe191761bcc169e895ac42817a403a2c038f6a/src/components/table/Table.vue#L1140
To prevent this, wrap the line with: if (i >= 0) { ... }

GMartigny:
Fixes #3461

## Proposed Changes

- Check before tying to close detail row

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -1137,13 +1137,14 @@ export default {
         closeDetailRow(obj) {
             const index = this.handleDetailKey(obj)
             const i = this.visibleDetailRows.indexOf(index)
-            this.visibleDetailRows.splice(i, 1)
+            if (i >= 0) {
+                this.visibleDetailRows.splice(i, 1)
+            }
         },
 
         isVisibleDetailRow(obj) {
             const index = this.handleDetailKey(obj)
-            const result = this.visibleDetailRows.indexOf(index) >= 0
-            return result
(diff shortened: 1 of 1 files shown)


jtommy:
Thanks

PR #3476, opened 2021-08-31
emit the number of table rows in the filtered data set when newDataTo…

ccondry:
emit the number of table rows in the filtered data set when newDataTotal changes

## Proposed Changes

- emit the number of rows in the filtered data set for a table when filtering is applied, either with backend filtering or built-in filtering. The use case I am interested in is displaying the current number of filtered results in the table footer.


jtommy:
Why do you need it?

ccondry:
The use case I am interested in is displaying the current number of filtered results in the table footer.

PR #3477, opened 2021-09-01
Update modal example to use :active instead of v-model

dylanlott:
The current Modal example uses `v-model` which doesn't work anymore. `:active` should be used instead.

## Proposed Changes

- Use `:active` instead of `v-model` in modal example


jtommy:
Thanks

PR #3482, opened 2021-09-06
[Fixing bug]: Jumpy collapse slide Fixed 

### LINKED ISSUE #3471: Slide transition is jumpy
### Overview of the problem

**Buefy** version: 0.9.8
**Vuejs** version: 2.6.14
**OS/Browser**: Chrome 92

### Description

When using Collapse component and using the slide transition, the animation is very jumpy. It slide on `100px` and then jump to completion.

### Steps to reproduce

https://codepen.io/GMartigny/pen/KKqKBjm

### Expected behavior

The animation should be smooth and cover the whole height of the component.

### Actual behavior

Animation is only on `100px` because this value is hardcoded on [`_animation.scss`](https://github.com/buefy/buefy/blob/dev/src/scss/utils/_animations.scss#L177).

mouadTaoussi:
Fixes #3471
Since slide animation is jumpy from 100px as described in this issue #3471, The issue might be bad for longer collapses. because it cannot cover all the collapse by the animation.

## Proposed Changes
- Updated `max-height` from `100px` to `999px`, so it gonna cover all the collapse without any other **issues**.


jtommy:
And when the content is higher than 999px ?

mouadTaoussi:
Raise it down ? it just to prevent the jumpy collapse.

jtommy:
@GMartigny what do you think?

GMartigny:
As per https://github.com/buefy/buefy/issues/3471#issuecomment-904656941, making the animation start from `9999px` will not be ideal.

The browser will do the transition from 0 to 9999px in the set amount of time ([150ms](https://github.com/buefy/buefy/blob/dev/src/scss/utils/_variables.scss#L1)) therefore making it very very quickly. Also, when retracting, the animation will go from 9999px to 0 and will do nothing from 9999px to Hpx (H being the height of the animated component).

mouadTaoussi:
@GMartigny Which solution you think is good for your needs?

GMartigny:
> I think that solved the jumpy slide which was a problem for longer collapses.

It solves the issue but with the caveats explained above.

> Which solution you think is good for your needs?

Sadly I don't know. From my own research, it's not solvable with only CSS. In order to have a smooth animation, the height of the component should be first computed.

PR #3492, opened 2021-09-11
Autocomplete keyboard navigation enhancement

### LINKED ISSUE #3437: Keyboard Navigation for Header and Footer Slots of Autocompletes
### Overview of the problem

**Buefy** version: [0.9.8]
**Vuejs** version: [2.6.12]
**OS/Browser**: Chome and Edge

### Description
The Header and Footer slots of autocompletes can not be navigated to with the arrow keys. Currently, it can only be selected by a mouse. I believe this is incorrect and the option to use arrow keys should exist.

no-keyboard-navigation-header

### Steps to reproduce

1.Open an autocomplete with a header or footer slot.
2.Attempt to navigate to the header or footer slot.

### Expected behavior

I believe that the header and footer slots should be able to be navigated to with arrow keys. When the Keep-First option is selected the first result should still be the one hovered. However, the user should be able to hit the up arrow to navigate to the header item.

### Actual behavior
The ability to navigate to the header and footer slots is not present.

JDurman:
Fixes #3437

## Proposed Changes

- Header and footer slots for autocompletes can now be navigated to with a keyboard.
- Added a select-header and select-footer event. This can be used to call a function when the appropriate slot is selected.
- I also added the keep-first and open-on-focus to the header example in the documentation so that people can see how the behavior interacts with the new slot behavior.

### DIFF
--- a/src/components/autocomplete/Autocomplete.vue
+++ b/src/components/autocomplete/Autocomplete.vue
@@ -44,7 +44,7 @@
                         role="button"
                         tabindex="0"
                         :class="{ 'is-hovered': headerHovered }"
-                        @click="selectableHeadersetSelected('header', undefined, $event)"
+                        @click="(event) => checkIfHeaderOrFooterSelected(event, true)"
                     >
                         <slot name="header" />
                     </div>
@@ -91,7 +91,7 @@
                         role="button"
                         tabindex="0"
                         :class="{ 'is-hovered': footerHovered }"
-                        @click="setSelected('footer', undefined, $event)"
+                        @click="(event) => checkIfHeaderOrFooterSelected(event, true)"
                     >
                         <slot name="footer" />
                     </div>
@@ -378,15 +378,6 @@ export default {
          */
         setSelected(option, closeDropdown = true, event = undefined) {
             if (option === undefined) return
-            if (this.selectableHeader && option === 'header') {
-                this.$emit('select-header')
-                return
-            }
-            if (this.selectableFooter && option === 'footer') {
-                this.$emit('select-footer')
-                return
-            }
-
             this.selected = option
             this.$emit('select', this.selected, event)
             if (this.selected !== null) {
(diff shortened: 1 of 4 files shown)


src/components/autocomplete/Autocomplete.vue
```diff
@@ -38,9 +38,16 @@
                     class="dropdown-content"
                     v-show="isActive"
                     :style="contentStyle">
-                    <div v-if="hasHeaderSlot" class="dropdown-item">
+                    <a
+                        v-if="hasHeaderSlot"
+                        class="dropdown-item"
+                        role="button"
+                        tabindex="0"
+                        :class="{ 'is-hovered': 'header' === hovered }"
```
jtommy, line 46: What happen if an option value is "header" or "footer" ?

src/components/autocomplete/Autocomplete.vue
```diff
@@ -38,9 +38,16 @@
...
+                        @click="setSelected('header', undefined, $event)"
```
jtommy, line 47: Why don't you emit 'select-header' instead of call `setSelected` ?

jtommy (changes requested):
I didn't try but reading the code there are a few things to change, please check my review.
Also it adds breaking changes, for example the replacing from `div` to `a` and the header and footer are now always selectable (probably better to add two props that enable that behaviour)

src/components/autocomplete/Autocomplete.vue
```diff
@@ -78,9 +85,16 @@
                         class="dropdown-item is-disabled">
                         <slot name="empty" />
                     </div>
-                    <div v-if="hasFooterSlot" class="dropdown-item">
+                    <a
+                        v-if="hasFooterSlot"
+                        class="dropdown-item"
+                        role="button"
+                        tabindex="0"
+                        :class="{ 'is-hovered': 'footer' === hovered }"
```
jtommy, line 93: I am not sure about it.. is it a good to show the item with with hovered class? I should try it before..

jtommy:
@JDurman did you read my review ?

JDurman:
@jtommy Yes, I did. I am currently working on implementing your suggestions. I added two props for making it selectable or not. The last thing I am working on is fixing the select event. I should finish it today.

src/components/autocomplete/Autocomplete.vue
JDurman: Yes, this would be an issue. I am resolving this now by using an internal variable for managing the state of hover for the header/footer and leaving the hovered variable for just items in the list.

src/components/autocomplete/Autocomplete.vue
JDurman: I think it needs the hover effect otherwise it's very hard to tell when you have the header or footer selected.

src/components/autocomplete/Autocomplete.vue
JDurman: Your question made me realize that setSelected was probably not the best function to be handling this event since it won't actually be setting a value. I went ahead and made a function for it instead that handles both the confirm keys and click event.

src/components/autocomplete/Autocomplete.vue
JDurman: To clarify I made it so that it only shows the hover when using the arrow keys. It won't show when hovered by a mouse.

jtommy (approved):

jtommy:
@JDurman thanks 👍

PR #3496, opened 2021-09-12
Update: redundant css class removal

egoomni:
Issue: Provokes an error if using an external a non-default icon pack.  
Fix: Removing undesired "minus" and "plus" classes from the <p class="control XXX ",

### DIFF
--- a/src/components/numberinput/Numberinput.vue
+++ b/src/components/numberinput/Numberinput.vue
@@ -3,7 +3,7 @@
         <p
             v-for="control in controlsLeft"
             :key="control"
-            :class="['control', control]"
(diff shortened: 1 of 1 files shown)


jtommy:
@egoomni why do you want to remove it ? It might help the customization of buttons

egoomni:
> @egoomni why do you want to remove it ? It might help the customization of buttons

Just figured it out, in my case I was using an icon font that had no prefix in its class name (ex: default "mdi-minus"; mine: "minus" ) so the class names were coinciding resulting in a broken design. Sorry for an ignorant pull request!

PR #3502, opened 2021-09-16
Don't attach attrs to NavbarDropdown

### LINKED ISSUE #3475: 0.9.9 introduce duplication of attributes
### Overview of the problem

**Buefy** version: [0.9.9]
**Vuejs** version: [2.6.14]
**OS/Browser**:
Windows10/Vivaldi
### Description

On some conditions buefy duplicate an attribute.
my case is with `data-test-id` attribute

### Steps to reproduce

1 - create a template like the below:
```
<template>
		<b-navbar-dropdown
		data-test-id="localeSwitch"
    class="locale_switch"
    :label="$i18n.locale"
		collapsible
		hoverable
    >
		<b-navbar-item
		@click="switchLanguage"
		href="#"
		>
		en
		</b-navbar-item>
		<b-navbar-item
		data-test-id="localeHe"
		@click="switchLanguage"
		>
		he
		</b-navbar-item>
	</b-navbar-dropdown>
</template>
```

After building the bundle and serving there are two elements with `data-test-id="localeSwitch"`
The outer `div` under the `navbar-end` and the chosen `a` element under this `div`

image

### Expected behavior
There is only one element with the `data-test-id="localeSwitch"` attribute and it is clickable

### Actual behavior

the attribute is duplicated and there is no one and only one clickable element with this attribute

GMartigny:
Fix #3475

## Proposed Changes

- Add `inheritAttrs: false,` on `NavbarDropdown` component

### DIFF
--- a/src/components/navbar/NavbarDropdown.vue
+++ b/src/components/navbar/NavbarDropdown.vue
@@ -44,6 +44,7 @@ export default {
     directives: {
         clickOutside
     },
+    inheritAttrs: false,
     props: {
         label: String,
         hoverable: Boolean,


GMartigny:
While checking that no other `inheritAttrs: false` was missing, I found out that [`paginationButton` has a useless `v-bind="$attrs"`](https://github.com/buefy/buefy/blob/dev/src/components/pagination/PaginationButton.vue#L9) because it's at the root level.
Removing it doesn't seams to change anything, but I might be missing something.

jtommy:
In the pagination button is ok with v-bind

PR #3503, opened 2021-09-17
fix bug on datetimepicker when use granularity

### LINKED ISSUE #3500: datetimepicker - min-datetime not working correctly with non-default granularity
### Overview of the problem
**Buefy** version: v0.9.10
**Vuejs** version: v2.6.14
**OS/Browser**: Chrome Version 92.0.4515.159 (Officiel version) (64-bit)

### Description

When configuring the datetimepicker to use both granularity (`incrementMinutes: 15` and `min-datetime: new Date()`, selecting the current date in the picker causes the picker to break. It breaks as a result of the selected datetime being invalid. 

Here's Codepen to illustrate the issue:
https://codepen.io/Impero/pen/GREypEo

### Steps to reproduce

1. Configure the datetimepicker as seen in the above codepen
2. Click the input field and select the current date
3. Observe that the timepicker shows empty values and attempting to pick a different date doesn't work

### Expected behavior

I expect the selected date to be rounded to a date closest to a valid date according to the selected granularity. 

### Actual behavior

image

geeksilva97:
Fixes #3500

## Proposed Changes

- Check if date is valid


src/components/datetimepicker/Datetimepicker.vue
```diff
@@ -166,8 +166,11 @@ export default {
             get() {
                 return this.newValue
             },
+
             set(value) {
                 if (value) {
+                    // check invalid date
+                    value = isNaN(value.getTime()) ? this.newValue : value
```
GMartigny, line 173: This would only solves the issue on DateTimePicker, but I guess the problem exists with every components depending on TimePicker. The `TimepickerMixin.js` should be improve instead.

GMartigny (changes requested):
Please revert changes made to `packag-lock.json`

src/components/datetimepicker/Datetimepicker.vue
geeksilva97: Ok. I will solve it in the root of problem.

geeksilva97: I have tested the granularity on TimePicker and it works fine. Apparently the problem only occurs in the Datetimepicker. I noticed that when I click to select a date without time the problem is occurring. 

Here is after the component opened
print

Here is after select a date (in this point the `input` is being emitted with `Invalid Date`).
print2

geeksilva97: Note the timepicker. It was cleaned, or something, and after that I can't select time anymore.

geeksilva97: I'll debug to get more details about what is happening.

geeksilva97:
Hi @GMartigny.
You was right, I moved the check if date is valid to `TimePickerMixin.js` and it solved problem too. 

When a date is selected in DateTimePicker the method `updateDateSelected` is fired with values of `hours`,  `minutes`, `seconds` and `meridiens`. The first `time`, the hours arrive as `NaN` and this causes the `Invalid Date`.

GMartigny (approved):

PR #3505, opened 2021-09-17
Fix bug on loading when using fixed navbar

### LINKED ISSUE #3335: B-Table loading overlays nav-bar when is fixed-top
### Overview of the problem
Table loading overlays nav-bar when is fixed-top

**Buefy** version: [0.9.7]
**Vuejs** version: [2.6.12]
**OS/Browser**:

### Description

B-Table loading overlays nav-bar when is fixed-top

image

### Steps to reproduce
https://codepen.io/mrtnzagustin/pen/VwpKERZ

geeksilva97:
Fixes #3335

## Proposed Changes

- Set `z-index` of `.loading-overlay` to `999` only when full page.  Otherwise, keep it smaller than the `z-index` of the fixed navigation bar.

### DIFF
--- a/src/scss/components/_loading.scss
+++ b/src/scss/components/_loading.scss
@@ -9,7 +9,7 @@ $loading-full-page-icon-size: 5em !default;
     display: none;
     justify-content: center;
     overflow: hidden;
-    z-index: 999;
+    z-index: 29;
     &.is-active {
         display: flex
     }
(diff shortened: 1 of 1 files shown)


jtommy:
Why do you need it?

geeksilva97:
> Why do you need it?

Are two main reasons. 

The first is because the behavior sounds weird. If the loader is only in a `table`, for example, I think the loader should not stay over a certain element, unless if the `table` is over that element.

The second is to recover behavior. As mentioned in the ISSUE this was working the way I said in the first reason.

PR #3506, opened 2021-09-19
feat: 🎸 Adding window mixin

ajomuch92:
Adding window size mixin to watch for changes programmaticall with JS according with Bulma breakpoints.


jtommy:
How do you think to use it? Why should add it in the core code?

ajomuch92:
@jtommy I thinks is a good feature to show or not (using v-if) some HTML elements or components not only using CSS. Also to use or toggle some classes based on the device size. Another Vue libraries has methods to determine the window size on flight, I think Buefy could have one too using this mixin.

ajomuch92:
@jtommy what do you think? I could be awesome if you merge this PR, besides the project I'm currently working needs this one

jtommy:
I don't know, it is an helper that internally I wouldn't use.

PR #3509, opened 2021-09-20
TagInput: Add capability to listen Autocomplete events

### LINKED ISSUE #3504: Impossible to listen to Autocomplete events in TagInput component
### Overview of the problem

**Buefy** version: 0.9.9
**Vuejs** version: 2.6.14
**OS/Browser**: Chrome

### Description

The `TagInput` component is built upon the `Autocomplete` component, but it's not passing down the event listeners.

### Steps to reproduce

Use `Taginput` component and try to listen to `Autocomplete`'s events.

### Expected behavior

It's possible to listen to `icon-right-click` or `select-header` event for example.

### Actual behavior

Nothing happens.

https://codepen.io/GMartigny/pen/jOwZOKG?editors=1010

### Possible fix

Adding a `v-on="$listeners"` to the [`<b-autocomplete>` of `TagInput.vue`](https://github.com/buefy/buefy/blob/dev/src/components/taginput/Taginput.vue#L34)

geeksilva97:
Fixes #3504

## Proposed Changes

- Emit the rightIconClick event

### DIFF
--- a/src/components/taginput/Taginput.spec.js
+++ b/src/components/taginput/Taginput.spec.js
@@ -1,4 +1,4 @@
-import { shallowMount } from '@vue/test-utils'
+import { shallowMount, mount } from '@vue/test-utils'
 import BTaginput from '@components/taginput/Taginput'
 import BTag from '@components/tag/Tag'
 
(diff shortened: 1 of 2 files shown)


jtommy:
Maybe better $listeners?

geeksilva97:
> Maybe better $listeners?

Yes! I'm working on it

geeksilva97:
Well, I added `v-on`, but not directly from `$listeners`. I added a computed entry and pass all listeners except for `input`. The `input` event was crashing `TagInput` in some ways:

1.  definition of props, since they were trying to set the value to String, but only Array is acceptable (https://github.com/buefy/buefy/blob/dev/src/components/taginput/Taginput.vue#L115)
2. we can change the definition to accept String and Array, but it crashes exhibition where each typed word is being split as tags (unexpected behavior)

In my tests, I've found that ignoring the `input` event doesn't affect TagInput at all. What you think?

jtommy:
I should investigate in the deep, at the moment if is missing only right click please add only it as your first commit

geeksilva97:
Actually, except for `input`, all events of `Autocomplete` can be listened.

geeksilva97:
I added a test to listen for autocomplete events.

PR #3513, opened 2021-09-22
Fix disappearing message bug

### LINKED ISSUE #3429: b-field with has-addons : message disappear when input is blured
### Overview of the problem

**Buefy** version: [0.9.8]
**Vuejs** version: [2.6.14]
**OS/Browser**: Edge, Chrome (lastest)

### Description

When a b-field has addons and a message, the message disappear when an input field is blured.

### Steps to reproduce

Here is a codepen: https://codepen.io/CRAKERSS/pen/bGWYOYW?editors=1000

geeksilva97:
Fixes #3429

As described in ISSUE, the help message is disappearing when focus/blur on a b-field with add-ons. This is happening because an undefined value is given to `parent.newMessage` (set from watched newMessage property).

## Proposed Changes

- Remove `message` property only instead of `newMessage`.

### DIFF
--- a/src/components/field/Field.vue
+++ b/src/components/field/Field.vue
@@ -25,7 +25,7 @@
         </template>
         <b-field-body
             v-if="horizontal"
-            :message="newMessage ? formattedMessage : ''"
+            :message="message ? formattedMessage : ''"
             :type="newType">
             <slot/>
         </b-field-body>
@@ -99,7 +99,6 @@ export default {
     data() {
         return {
             newType: this.type,
-            newMessage: this.message,
             fieldLabelSize: null,
             _isField: true // Used internally by Input and Select
         }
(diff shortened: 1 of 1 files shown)


jtommy:
@geeksilva97 what about reset message? In this way it's no possible

geeksilva97:
It should work with empty string.

geeksilva97:
If this is not enough I can investigate more.

jtommy:
Users might use undefined or null at the moment

geeksilva97:
> Users might use undefined or null at the moment

This make sense. I found the root and I'm trying to fix.

geeksilva97:
@jtommy I think this solves the problem at the root. Can you review?

src/components/field/Field.vue
```diff
@@ -218,18 +217,11 @@ export default {
...
-        /**
-        * Set parent message if we use Field in Field.
-        */
-        newMessage(value) {
             if (this.parent && this.parent.hasInnerField) {
                 if (!this.parent.type) {
                     this.parent.newType = this.newType
                 }
-                this.parent.newMessage = value
+                this.parent.message = value
```
jtommy, line 224: Are you sure? You can't mutate a prop

jtommy (changes requested):

src/components/field/Field.vue
geeksilva97: I ran tests for `Field.vue` and all passes. Also, I tested `BField` element nesting and it worked. Is there something wrong with my tests?

jtommy: Really strange that Vue doesn't show you warnings or similar

geeksilva97:
This is so weird. I will close this PR and test more. Thanks for helping @jtommy

PR #3524, opened 2021-09-25
Added icon on the left and click-event to tag

Timoms:
## Reason

Tags are a beautiful way to show a specific type. With an icon on the left you can create different categories like "car" and "motorcycle" using different icons. That way the user can exactly see what type of category this tag applies to.

Click events on the other hand are useful when implementing e.g. taglists where the user can interact with different tags to emit specific tasks (like apply a search filter based on the tag clicked).

## Proposed Changes

- Added icon on the left using:

`<tag icon="cloud-tags" closable attached>Tag with icon on the left!</tag>`

- Added click event to tag content using

`<tag icon="cloud-tags" @click="testClick">Tag with icon on the left!</tag>`

### DIFF
--- a/src/components/tag/Tag.vue
+++ b/src/components/tag/Tag.vue
@@ -3,7 +3,8 @@
         <span
             class="tag"
             :class="[type, size, { 'is-rounded': rounded }]">
-            <span :class="{ 'has-ellipsis': ellipsis }">
+            <b-icon custom-class="" v-if="icon" :icon="icon" :size="size" :type="iconType" :pack="iconPack" />
+            <span :class="{ 'has-ellipsis': ellipsis }" @click="click">
                 <slot/>
             </span>
         </span>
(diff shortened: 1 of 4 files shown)


src/components/tag/Tag.vue
```diff
@@ -33,7 +34,8 @@
         v-else
         class="tag"
         :class="[type, size, { 'is-rounded': rounded }]">
-        <span :class="{ 'has-ellipsis': ellipsis }">
+        <b-icon custom-class="" v-if="icon" :icon="icon" :size="size" :type="iconType" :pack="iconPack" />
```
jtommy, line 37: Please remove `custom-class`

jtommy (commented):
Please update API docs, tag.js file and regen snapshot running tests

src/components/tag/Tag.vue
Timoms: Thanks for letting me know, going to do the changes in the upcoming days.

mouadTaoussi:
The snapshot is no longer comparable to the newest version of the component, I think you should rewrite it, it occurs lot of failing tests.

PR #3526, opened 2021-09-26
[Feature Request]: Added Progress rounded property

### LINKED ISSUE #3525: [Feature Request]: Adding a rounded property in progress component
### Description
When I want to create remaining time in Message; Notifications component, I found no way to make the progress not rounded! some  cases needs to be not rounded, so I made it manually. Why not creating a rounded property to make it simple?

Some cases like above needs a not rounded progress.

mouadTaoussi:
Fixes #3525

## Proposed Changes
- Added **rounded** property, default to **true**
- Added **is-not-rounded**, class
- Added conditions in classes in **progress** and wrapper elements
- Added Example in docs
- Made each attribute in docs of progress Examples in new line for readability

### DIFF
--- a/src/scss/components/_progress.scss
+++ b/src/scss/components/_progress.scss
@@ -63,12 +63,10 @@
                 }
             }
         }
-        &.is-not-rounded {
-            border-radius: 0!important
+
+        &.is-squared {
+            border-radius: 0;
         }
-    }
-    &.is-not-rounded {
-        border-radius: 0!important
     }
 
     &.is-not-native {
@@ -99,5 +97,9 @@
                 }
             }
         }
+    }
+    
+    &.is-squared {
+        border-radius: 0;
     }
 }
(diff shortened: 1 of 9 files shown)


src/scss/components/_progress.scss
```diff
@@ -63,6 +63,12 @@
                 }
             }
         }
+        &.is-not-rounded {
```
jtommy, line 66: I prefer `is-squared` and please don't use !important

jtommy (changes requested):

src/scss/components/_progress.scss
mouadTaoussi: it cannot set without !important keyword

mouadTaoussi: Done!

jtommy (approved):

PR #3531, opened 2021-09-28
Image custom class

vikiival:
## Rationale

Buefy is an awesome UI lib but there are cases that we need to use `figure` + `img` instead of `b-image`.
This could be easily flexed by adding custom css class into the img tag
-
-
-

### DIFF
--- a/docs/pages/components/image/api/image.js
+++ b/docs/pages/components/image/api/image.js
@@ -85,6 +85,13 @@ export default [
                 values: '—',
                 default: '<code>false</code>'
             },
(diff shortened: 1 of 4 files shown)


mouadTaoussi:
The snapshot is no longer comparable to the newest version of the component, I think you should rewrite it, it occurs lot of failing tests.

vikiival:
> I think you should rewrite it

'It' sounds too ambiguous in the context and I am not sure if I am following. 
Could you be more descriptive please ?

PR #3539, opened 2021-09-29
Fix notice queue

### LINKED ISSUE #3424: Toast breaks when Indefinite Toast opens and closes while another toast is active.
### Overview of the problem
We encountered this issue on our app, and found out that it's also happening on the example on the website. This issue revolves around a regular Toast and an Indefinite Toast occurring almost at the same time. 

**Buefy** version: Latest
**Vuejs** version: [2.3.0]
**OS/Browser**: Windows/Edge

### Description
When a regular toast with some timeout (3000ms for example) is being displayed, and we display an Indefinite Toast and close it while the regular Toast is still active, it will break and won't allow us to create or open any new toast.

**This is reproducible with the current example on the website**

### Steps to reproduce
image
1. Navigate to the toast example https://buefy.org/documentation/toast/#from-inside-vue-instance
2. Hit the green `Launch Toast (custom)` button
3. While the toast is active, hit `Launch Toast (indefinite)` and then `Close Toast (indefinite)`, while the toast from step 2 is still active.
4. Observe that toast will stop working.

### Expected behavior
Toast to continue function as normal.

### Actual behavior
Toast stopped functioning. Cannot open new toast messages.

geeksilva97:
Fixes #3424

Following steps to reproduce it is possible to see that no other Toast can be opened, but the real and big problem was happening in recursion. Since the toast is closed (with` .close() `call) before added to DOM the component is destroyed, but nothing happens with the element that will added to DOM. It will be added to DOM anyway,  it is not visible, but is in the DOM.

In this scenario the `shouldQueue()` always return `true` because of the following condition.

```js
return (
    this.parentTop.childElementCount > 0 ||
    this.parentBottom.childElementCount > 0
)
```

In `showNotice()` this ends the execution and call `showNotice()` again, and it will **`happen FOREVER`**.

## Proposed Changes

- Check if component is destroyed

### DIFF
--- a/src/utils/NoticeMixin.js
+++ b/src/utils/NoticeMixin.js
@@ -125,6 +125,7 @@ export default {
         },
 
         showNotice() {
+            if (this._isDestroyed) return
             if (this.shouldQueue()) {
                 // Call recursively if should queue
                 setTimeout(() => this.showNotice(), 250)
--- a/src/utils/NoticeMixin.spec.js
+++ b/src/utils/NoticeMixin.spec.js
@@ -48,4 +48,28 @@ describe('NoticeMixin', () => {
         wrapper.vm.close()
         expect(wrapper.vm.isActive).toBeFalsy()
     })
+
+    it('should not insert element when component is destroyed', () => {
+        jest.useFakeTimers()
+        HTMLElement.prototype.insertAdjacentElement = jest.fn()
+        let queueCounter = 0
+
+        const component = {
+            template: '<div class="b-component"></div>'
+        }
+        const wrapper = shallowMount(component, {
+            attachToDocument: true,
+            mixins: [NoticeMixin],
+            methods: {
+                shouldQueue() {
+                    if (queueCounter++ >= 5) wrapper.destroy()
+                    return true
+                }
+            }
+        })
+
+        jest.advanceTimersByTime(1500)
+
+        expect(HTMLElement.prototype.insertAdjacentElement).toHaveBeenCalledTimes(0)
+    })
 })


jtommy:
Why don't you remove it from DOM?

geeksilva97:
> Why don't you remove it from DOM?

In the current flow, the `.remove()` is being called, but before the element is in the DOM so the element doesn't have to be removed from. It is similar to the following code.

```js
const myDiv = document.createElement('div');
myDiv.remove()
```

From the code above, myDiv has no place to be removed.

Once the component is destroyed, we can check it before inserting it into the DOM. Can it be done by removing the element in the next row? Yes, but the element is already closed and destroyed. It makes sense that this element should never be present in the DOM.

Another point is the queue. If we allow an element from a destroyed component to be added to the DOM and then remove it, we will need two recursive calls, the first to add and the second to remove (the rest of the queue will wait for a won element to appear ). In both cases, we need to check if the component is already destroyed, otherwise the recursion will run forever (as it is currently happening).

In short: I think preventing a destroyed element from being added to the DOM is more understandable and keeping the queue free for others.

jtommy:
Ok, I understand but I would prefer a refractor instead of that trick

geeksilva97:
> Ok, I understand but I would prefer a refractor instead of that trick

Do you have any idea what this refactoring might look like?

jtommy:
I don't know at the moment, I should investigate in the code

geeksilva97:
> I don't know at the moment, I should investigate in the code

I understand. I can do some experiments too.

PR #3557, opened 2021-10-19
[Feature Request]: Added Backtotop Component

### LINKED ISSUE #3548: BackTop Component
### Description

Backtop can be used when page content is very long and a shortcut is needed to go back to the top fluently. Usually placed at the right bottom of the page.

image

mouadTaoussi:
Fixes #3548

## Proposed Changes
- Added new component for **BackToTop** 
- Added new component example to docs

Web capture_19-10-2021_23530_localhost
Web capture_19-10-2021_23450_localhost

## Why buefy needs this feature?
Making a function that scrolls back to top with animation would be hard and wasting time.


src/components/backtotop/Backtotop.vue
```diff
@@ -0,0 +1,92 @@
...
+                if (window.window.scrollY <= 500) {
+                    // hide
+                    this.isVisible = false
+                } else if (window.window.scrollY >= 500) {
+                    // show
+                    this.isVisible = true
+                }
+            }
+        })
+    }
```
Mars073, line 90: If you put a `v-if` depending on the route, the component will rebind window when the component becomes  visible and there will be events that make references that do not have to be. you should put your function in `method.checkScrollY` and remove the listener in the `beforeDestroy` method.

src/components/backtotop/Backtotop.vue
```diff
@@ -0,0 +1,92 @@
+<template>
+    <div
+        class="backtotop"
+        :class="[wrapperClasses, { 'is-visible' : isVisible, 'is-hidden' : !isVisible }]"
+    >
+        <b-button
+            @click="backToTop"
+            v-bind="$attrs"
+            icon-left="arrow-up"
```
Mars073, line 9: Does `arrow-up` exist in MDI and FA icons?

PR #3573, opened 2021-11-05
add github release action & replace circle ci test jobs.

kgrubb:
__CHANGES:__

* use github actions instead of circle ci for automated testing (lint, unit tests, codecov).
  * add matrix for multiple versions of node when testing (currently 14 & 16, since they're the active LTS releases).
  * tests run on pull requests and pushes to important branches(`dev`, `master`, `next`).
  * Update readme to show github actions test results in shields.io badge instead of circle ci badge

* automate release generation via tag pushes. A maintainer who has access to pushing git tags can create a new release for github and npm via running `npm version <newversion>`. For example, `npm version v0.9.12`.

  Running `npm version <newversion>` as a maintainer will:
    1. Bump the version in the `package.json`, commit the version changes in the `package.json` and `package-lock.json`.
    2. Push the new version to the current branch and push the related tag.
    3. Draft and publish a github release from the pushed tag. The release will also contain the changelog contents of the matching version.
    4. Publish the module to the npm registry (after building the dist dir).

__NOTES:__

* To use the new "release" github action, a maintainer's npm token will need to be added as a secret to the project in github. To add the secret, you can go to the project, select "Settings" -> "Secrets" -> "New repository secret". If you need to create an access token, npm has some good documentation on [how to create an access token](https://docs.npmjs.com/creating-and-viewing-access-tokens). When adding the secret, it should look something like this:

    ```
    name: NPM_TOKEN
    Value: mysupercoolnpmtoken
    ```

* The new release generation flow does not include any `CHANGELOG.md` generation. The changelog will need to be filled out before running the versioning command. The reason I kept changelog generation out of this pr is to allow human interaction for changelog creation. Currently buefy doesn't follow any kind of semantic commit format, so it's probably better to have a person create the changelogs for now.

* Currently, the `npm version` run-script pushes the new version to whatever branch you have checked out. If there is a different desired behavior, let me know! I'd be glad to change that behavior to whatever is more convenient for maintainers.


jtommy:
It looks ok but a question.. running it the email sent by GitHub on release doesn't contain any changelog, right?

PR #3584, opened 2021-11-10
Emit an 'active-change' event from the navbar dropdown

arusahni:
Thanks for Buefy!

## Proposed Changes

- Emit an `active-change` event from the navbar dropdown when it is shown or hidden

This change brings the navbar dropdown into closer API symmetry with the dropdown component.  My use case is I have dynamic content inside the dropdown that I wish to reset when the dropdown is hidden.  By listening for the event, I can handle it in my enclosing component.


jtommy:
Thanks

PR #3594, opened 2021-11-21
Refact NoticeMixin

### LINKED ISSUE #3424: Toast breaks when Indefinite Toast opens and closes while another toast is active.

geeksilva97:
This PR updates the `NoticeMixin` and Fixes #3424

## Proposed Changes

- Remove the recursive call when showing Notices

## Snackbar behavior after refactoring
ezgif com-gif-maker

## Toast behavior after refactoring
ezgif com-gif-maker (1)

### DIFF
--- a/src/components/snackbar/index.js
+++ b/src/components/snackbar/index.js
@@ -17,7 +17,8 @@ const SnackbarProgrammatic = {
 
         const defaultParam = {
             type: 'is-success',
-            position: config.defaultSnackbarPosition || 'is-bottom-right'
+            position: config.defaultSnackbarPosition || 'is-bottom-right',
+            queue: true
         }
         if (params.parent) {
             parent = params.parent
@@ -29,7 +30,6 @@ const SnackbarProgrammatic = {
             delete params.message
         }
         const propsData = merge(defaultParam, params)
-        propsData.queue = true
         const vm = typeof window !== 'undefined' && window.Vue ? window.Vue : localVueInstance || VueInstance
         const SnackbarComponent = vm.extend(Snackbar)
         const component = new SnackbarComponent({
--- a/src/components/snackbar/index.js
+++ b/src/components/snackbar/index.js
@@ -29,7 +29,7 @@ const SnackbarProgrammatic = {
             delete params.message
         }
         const propsData = merge(defaultParam, params)
-        propsData.replace = true
+        propsData.queue = true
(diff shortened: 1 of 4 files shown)


jtommy:
PR is ok but changing the prop name it becomes a breaking change.
I prefer to fix/improve the code without introducing changes to dev's code

geeksilva97:
> PR is ok but changing the prop name it becomes a breaking change. I prefer to fix/improve the code without introducing changes to dev's code

Hi, @jtommy. We can keep the variable names, I think that is OK. I changed it because I thought it could be clearer.

src/components/snackbar/index.js
```diff
@@ -29,7 +29,7 @@ const SnackbarProgrammatic = {
             delete params.message
         }
         const propsData = merge(defaultParam, params)
-        propsData.replace = true
+        propsData.queue = true
         const vm = typeof window !== 'undefined' && window.Vue ? window.Vue : localVueInstance || VueInstance
```
jtommy, line 33: Are you forcing it to be true??

jtommy (changes requested):

src/components/snackbar/index.js
geeksilva97: It was a mistake I made when  I tried to set `queue: true` as default in snackbars. I fixed it moving the `queue` property to the `defaultParams` object.

Issue #3578, opened 2021-11-07
Normalize string for search in tables (remove accents)


radex02:
### Description

Ignore accents when searching in a table. So, for exemple, "e" matches "é".

Something like https://stackoverflow.com/a/37511463 could be used.

### Why Buefy need this feature

There are numerous languages in which accents are used and its normal to expect searche forms to ignore them.

geeksilva97:
Can this ISSUE be closed, @jtommy ?

PR #3606, opened 2021-11-30
feat(Table): normalize string for search in tables

matheushahnn:
Feature requested https://github.com/buefy/buefy/issues/3578

## Proposed Changes

- Normalize string for search in tables

Example using filter in table to find the last name with accent
image

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -372,7 +372,7 @@
 </template>
 
 <script>
-import { getValueByPath, indexOf, multiColumnSort, escapeRegExpChars, toCssWidth } from '../../utils/helpers'
+import { getValueByPath, indexOf, multiColumnSort, escapeRegExpChars, toCssWidth, removeDiacriticsFromString } from '../../utils/helpers'
 import debounce from '../../utils/debounce'
 import { VueInstance } from '../../utils/config'
 import Checkbox from '../checkbox/Checkbox'
@@ -1169,7 +1169,8 @@ export default {
                         if (value !== Number(input)) return false
                     } else {
                         const re = new RegExp(escapeRegExpChars(input), 'i')
-                        if (!re.test(value)) return false
+                        const valueWithoutDiacritics = removeDiacriticsFromString(value)
+                        if (!re.test(valueWithoutDiacritics) && !re.test(value)) return false
                     }
                 }
             }
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -1170,7 +1170,7 @@ export default {
                     } else {
                         const re = new RegExp(escapeRegExpChars(input), 'i')
                         const valueWithoutDiacritics = removeDiacriticsFromString(value)
-                        if (!re.test(valueWithoutDiacritics) && !re.test(value)) return false
+                        return re.test(valueWithoutDiacritics) || re.test(value)
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


src/utils/helpers.js
```diff
@@ -174,6 +174,15 @@ export function escapeRegExpChars(value) {
...
     return value.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&')
 }
+/**
+ * Remove accents/diacritics in a string in JavaScript
+ * https://stackoverflow.com/a/37511463
+ */
+export function removeDiacriticsFromString(value) {
+    if (!value) return value
+
+    return value.normalize('NFD').replace(/[\u0300-\u036f]/g, '')
```
matheushahnn, line 184: I'd say the best approach would be using `value.normalize("NFD").replace(/\p{Diacritic}/gu, "")` but looks like babel has problems with this regex https://github.com/babel/babel/issues/8951.

src/components/table/Table.vue
```diff
@@ -1169,7 +1169,8 @@ export default {
                         if (value !== Number(input)) return false
                     } else {
                         const re = new RegExp(escapeRegExpChars(input), 'i')
-                        if (!re.test(value)) return false
+                        const valueWithoutDiacritics = removeDiacriticsFromString(value)
+                        if (!re.test(valueWithoutDiacritics) && !re.test(value)) return false
```
matheushahnn, line 1173: It keeps filtering if I search using accents or not. 
I thought to create a new param in the column to enable the filter with accents, but not sure if it is necessary. 
What do you think about it?

geeksilva97: `OR` could be enough here?

```suggestion
        return !re.test(valueWithoutDiacritics) || !re.test(value)
```

matheushahnn: Nice catch!
I just need to remove the `!` in the condition and it worked.

antoniopetlove: Cool!

jtommy:
@matheushahnn thanks 👍

MariusRedseed:
Hey team, this breaks searching in tables when the value is an array of strings.

jtommy:
@MariusRedseed array?

MariusRedseed:
@jtommy yup. This assumes the value in the column is a string. The pull request below attempted to allow the value of a column to be an array, but still be searchable:
https://github.com/buefy/buefy/pull/2723

jtommy:
Mmm you are right but it's no a good solution.. we need a field prop to filter the correct text

MariusRedseed:
Sure thing, I used a custom search for now, for a column containing an array of strings:
```
:custom-search="
    ({column_with_array_values}, input) => {
        const re = new RegExp(input, 'i')
        return column_with_array_values.some(role => re.test(role));
    }"
```

jtommy:
@MariusRedseed just restored

AN-WM:
This addition breaks the ability to search numbers with decimals. I fixed it for myself by adjusting the removeDiacriticsFromString() function to be:
`export function removeDiacriticsFromString(value) {
    if (!value) return value

    //Only normalize when value type === string
    if (typeof value === 'string' || value instanceof String) {
        return value.normalize('NFD').replace(/[\u0300-\u036f]/g, '');
    } else {
        return value;
    }
}`

Screenshot buefy search decimal
Screenshot error buefy search decimal

Issue #3605, opened 2021-11-29
b-upload HTML5 validation not working consistently


l1bra2013:
### Overview of the problem

**Buefy** version: 0.9.13
**Vuejs** version: 2.x

### Description

When using b-upload with the accept field, if an invalid file is selected, the input field is sometimes marked as valid despite being empty.

### Steps to reproduce

1. Setup b-upload component using parameters `required` and `accept`
2. Select an invalid file - doesn't work - despite file not being present, the form is still considered valid
3. Select a valid file - works properly - form can be submitted
4. Select an invalid file - works properly - form cannot be submitted

### Expected behavior

Invalid files don't allow form submission

### Actual behavior

Form is submitted despite having no file selected. Might be related to the value watcher (https://github.com/buefy/buefy/blob/dev/src/components/upload/Upload.vue#L82). Since the value doesn't change when the file is initially unset, the native input has a file set despite the Buefy input having no file set due to validation failure. Simple fix may be to always set value and ref input values to null here (https://github.com/buefy/buefy/blob/dev/src/components/upload/Upload.vue#L108).

jtommy:
@l1bra2013 please make a PR 😉

l1bra2013:
PR pushed - https://github.com/buefy/buefy/pull/3608

PR #3608, opened 2021-11-30
fix upload validation bugs

### LINKED ISSUE #3604: b-upload accept doesn't work with multi-dot file extensions
### Overview of the problem

**Buefy** version: 0.9.13
**Vuejs** version: 2.x

### Description

Buefy validation fails when using a multi-dot file extension in the b-upload accept field

### Steps to reproduce

1. Use b-upload with accept using multi-dot value such as `accept=".tar.gz"`

### Expected behavior

Validation should succeed

### Actual behavior

Validation always fails. Validation code (https://github.com/buefy/buefy/blob/dev/src/components/upload/Upload.vue#L152) uses index of last dot in file name, although this fails when multiple dots are present such as with `.tar.gz`. Might be better to just check if the file name ends with the `type` if the type starts with a `.`

l1bra2013:
Fixes #3604 and #3605

## Proposed Changes

- When validating file extensions, use extension length instead of looking for only single dot at end
- If invalid file is set and file wasn't previously selected, force validation
- Add example with validation to documentation

No tests added, although if there is a good mechanism for testing, I'd be happy to do so.

### DIFF
--- a/docs/pages/components/upload/Upload.vue
+++ b/docs/pages/components/upload/Upload.vue
@@ -7,6 +7,8 @@
         <Example :component="ExExpanded" :code="ExExpandedCode" title="Expanded" vertical/>
         
         <Example :component="ExRounded" :code="ExRoundedCode" title="Rounded" vertical/>
+        
+        <Example :component="ExValidation" :code="ExValidationCode" title="Validation" vertical/>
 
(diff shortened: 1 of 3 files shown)


jtommy:
@l1bra2013 thanks!

PR #3628, opened 2022-01-03
Fix typo with isNil helper method

### LINKED ISSUE #3618: Null values are always sorted to the bottom in the table
### Overview of the problem

**Buefy** version: [0.9.13]
**Vuejs** version: [2]
**OS/Browser**: Chrome

### Description

When there are null values under a sortable column in a table, sorting the column always results in the null values remaining in their same position at the bottom of the table. The expected behavior would be that these values would be at the top when sorting in descending order and at the bottom when sorting in ascending order.

Ryan-Case-SingleStone:
Fixes #3618

## Proposed Changes

- There was a typo when adding the new `isNil` helper method in #3619 to resolve the null value table sorting behavior
- This PR fixes this typo so that this method correctly checks for null or undefined values

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -372,7 +372,7 @@
 </template>
 
 <script>
-import { getValueByPath, indexOf, multiColumnSort, escapeRegExpChars, toCssWidth, removeDiacriticsFromString } from '../../utils/helpers'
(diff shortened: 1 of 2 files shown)


jtommy:
Oh you are right!

PR #3639, opened 2022-01-16
add page input to pagination component

gabrielchl:
Fixes #3428 #3274

## Proposed Changes

- add number input to pagination component to directly type in target page
- update table component to have options to toggle the said input

PR #3640, opened 2022-01-16
Added the page input to the b-pagination and the b-table that allows …

chadlafontaine:
…the user to enter a specific page. It also allows the developer to specify where the page input should be and specify a denounce time.

Fixes #3639

## Proposed Changes
Added page input to b-pagination and b-table
Allows developer to specify where the page input should be located in the pagination.
Allows developer to specify denounce in milliseconds.
Allows developer to specify page list location in b-table similar to b-pagination.
-
-
-

### DIFF
--- a/src/components/pagination/Pagination.vue
+++ b/src/components/pagination/Pagination.vue
@@ -3,60 +3,76 @@
         <slot
             v-if="$scopedSlots.previous"
             name="previous"
-            :page="getPage(current - 1, {
-                disabled: !hasPrev,
-                class: 'pagination-previous',
-                'aria-label': ariaPreviousLabel
-        })">
+            :page="
+                getPage(current - 1, {
+                    disabled: !hasPrev,
+                    class: 'pagination-previous',
+                    'aria-label': ariaPreviousLabel,
+                })
+            "
+        >
             <b-icon
                 :icon="iconPrev"
                 :pack="iconPack"
                 both
-                aria-hidden="true"/>
+                aria-hidden="true" />
         </slot>
         <BPaginationButton
             v-else
             class="pagination-previous"
             :disabled="!hasPrev"
             :page="getPage(current - 1)"
-            :aria-label="ariaPreviousLabel">
+            :aria-label="ariaPreviousLabel"
+        >
             <b-icon
                 :icon="iconPrev"
                 :pack="iconPack"
                 both
-                aria-hidden="true"/>
+                aria-hidden="true" />
         </BPaginationButton>
         <slot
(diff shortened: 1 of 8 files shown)


src/components/pagination/Pagination.vue
```diff
@@ -3,98 +3,103 @@
...
         </BPaginationButton>
 
+        <b-numberinput
+            v-if="pageInput"
+            class="pagination-input"
+            min="1"
+            :max="pageCount"
+            :value="inputValue"
+            :controls="false"
+            @input.native="onInput"
```
jtommy, line 71: Why native? Input already emits a number

src/components/pagination/Pagination.vue
```diff
@@ -154,16 +162,29 @@ export default {
...
         ariaPreviousLabel: String,
         ariaPageLabel: String,
-        ariaCurrentLabel: String
+        ariaCurrentLabel: String,
+        pageInput: {
+            type: Boolean,
+            default: false
+        },
+        pageInputPosition: String,
+        debouncePageInput: Number
```
jtommy, line 171: What's the need about debounce?

chadlafontaine: This is to give the developer an option to put in their own denounce time depending on their needs that way a query, for example backend pagination, is not sent for every number input if the user is still inputting a large number.

src/components/pagination/Pagination.vue
jtommy: Why not simply input?

chadlafontaine: I can. and should just use <input>. I will change that to just an input, since I am just getting the event and then getting the value from the event.

chadlafontaine: Sure no problem. I am changing it now.

jtommy (changes requested):
The PR looks ok but two things to change:
- force the input `type` to `number`
- the input is too wide in my opinion, you might set the size attribute by max page number.
- automatic clear value when is 0 or > max page

src/components/pagination/Pagination.vue
```diff
@@ -177,66 +198,66 @@ export default {
         },
 
         /**
-        * Total page size (count).
-        */
+     * Total page size (count).
```
gabrielchl, line 201: just nitpicking, you might want to re-align the comment blocks

gabrielchl (commented):
(also, you might want to remove the .vs directory from git :) )

jtommy (approved):

PR #3645, opened 2022-01-24
Fix: solve two bugs in multiple sorting table

### LINKED ISSUE #3638: Sort Arrow still persist even after sorting is removed
Overview of the problem

Buefy version: [0.9.14]
Nuxtjs version: [2.15]
OS/Browser: Chrome

Description

https://buefy.org/documentation/table/#sorting-multiple

When i use this feature with multiple sorting and backend sorting enabled. During the removal of the primary and secondary
sorting, the arrow icon is still persist even after removing all sorting.

image

Steps to reproduce

Enable Multisort and Backend Sorting

1. Click on first column header,
2. Click on second column header
3. Click on third column header
4.  Now remove third column sorting
5. After removing third column sorting arrow icon is still persisting which should not 

Expected behaviour

Arrow icon should not persist after removing sorting from the header icon.

nimaebra:
Fixes #3638

### First Problem (Issue #3638)

When multiple sorting and backend sorting are enabled, even if all sorts are removed, the arrow icon persists.

https://user-images.githubusercontent.com/78249749/150859704-0866ed07-7e20-48d3-adb7-d63bcf3e2aff.mov

I fix this issue in 27246c1ed192044877e3d8ce4e17cee36221a5c7.

### Second Problem

In multiple sorting, the table should be reset when all sorts are removed:

https://user-images.githubusercontent.com/78249749/150859656-2dd676de-003d-4c99-9d6c-51d15e7033b6.mov

I fix this issue in 27246c1ed192044877e3d8ce4e17cee36221a5c7 by adding a simple condition in `removeSortingPriority` method and call `resetMultiSorting` method if the length of `formattedSortingPriority` is equal to zero.

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -880,7 +880,12 @@ export default {
                 let formattedSortingPriority = this.sortMultipleDataLocal.map((i) => {
                     return (i.order && i.order === 'desc' ? '-' : '') + i.field
                 })
-                this.newData = multiColumnSort(this.newData, formattedSortingPriority)
+
+                if (formattedSortingPriority.length === 0) {
+                    this.resetMultiSorting()
+                } else {
+                    this.newData = multiColumnSort(this.newData, formattedSortingPriority)
+                }
             }
         },
         resetMultiSorting() {
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -988,8 +988,8 @@ export default {
                 }
                 if (!this.backendSorting) {
                     this.doSortSingleColumn(column)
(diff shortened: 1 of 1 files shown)


dev-avenger:
@nimaebra what if formattedSorting is not equal to zero? You want two sorting priorities but remove the third one but after removing third one, the arrow will still persist.

nimaebra:
@dev-avenger  I think you are relating the first and second problems, but they are completely independent. Check `formattedSortingPriority` length is only needed to reset the table and will not affect the issue with persisting arrow icons.

PR #3372, opened 2021-06-18
Close #3371: Update to Bulma 0.9.3

service-paradis:
Fixes #3371

## Proposed Changes

- Update to Bulma 0.9.3
- Remove SCSS division to prevent Dart SASS warning

PR #3670, opened 2022-02-16
Fixes slash div warnings

### LINKED ISSUE #3669: Math Div Warnings in 0.9.15
### Overview of the problem

**Buefy** version: 0.9.15

### Description

Math div introduced in the newest version of Buefy which is deprecated. 

Screen Shot 2022-02-16 at 9 20 02 AM

SomethingNew71:
Signed-off-by: Cole Gentry <peapod2007@gmail.com>

Fixes #3669 

## Proposed Changes

- migrate away from / division

### DIFF
--- a/src/scss/components/_colorpicker.scss
+++ b/src/scss/components/_colorpicker.scss
@@ -1,5 +1,3 @@
-@use "sass:math";
-
 $colorpicker-radius: $dropdown-content-radius !default;
 $colorpicker-hue-selected-stroke: $primary !default;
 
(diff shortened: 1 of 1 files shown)


jtommy:
I prefer to avoid use math dependencies as you can see in the other components

SomethingNew71:
I just used the `sass-migrator` offered by dart to update them. The slash div stuff is deprecated so it throws downstream errors for anyone using newer versions of sass. Would you prefer it be `calc(12px / 4px);` as that is going to be supported according to the sass docs?

jtommy:
As you can see https://github.com/buefy/buefy/pull/3372/files we avoid to introduce math package and we transformed as multiply

SomethingNew71:
> As you can see https://github.com/buefy/buefy/pull/3372/files we avoid to introduce math package and we transformed as multiply

Updated! 😄

PR #3706, opened 2022-03-12
Changed the checkbox to listen for a change event instead of a just a click event

### LINKED ISSUE #3432: Accessibility glitch: Keyboard does not trigger b-table checkbox change (click)
### Overview of the problem

**Buefy** version: 0.9.8
**Vuejs** version: 2.6.14
**OS/Browser**: macOS Big Sur 11.2.2 / Safari 14.0.3

### Description

When checking a checkbox in a Buefy Table with Checkboxes __using the keyboard or a screen reader__ such as VoiceOver, the visual and "spoken" states of the checkbox update properly, but the __checkedRows__ property of the table does not change, and (or possibly because) the __click__ event is not fired.

### Steps to reproduce

1. Setup any Buefy Table example with checkboxes with which you can watch the checkedRows state somewhere;
2. Check/uncheck rows with the mouse. Works properly.
3. Focus a checkbox with the keyboard. Use space bar to toggle. checkedRows property does not update.
4. Turn on VoiceOver and navigate to a checkbox. Toggle the checkbox. It toggles, but checkedRows property does not update.

### Expected behavior

checkedRows property should update no matter the method of input. Would be good to have outline focus on keyboard focus on the checkboxes as well, but this can be added in CSS.

### Actual behavior

checkedRows not updating on keyboard/screen reader input.

### Workaround

I added the following to my updated() lifecycle hook, which has the onchange handler fire a click to its parent for each checkbox input on DOM update (__this assumes <b-table ref="table" ...__):

```js
updated() {
    const that = this
    const changeListener = (e) => {
                const event = new Event('click')
                e.target.parentNode.dispatchEvent(event)
            }
    this.$nextTick(() => {
        that.$refs.table.$el.querySelectorAll('input[type="checkbox"]').forEach(el => {
            if (el.onchange == null)
                el.onchange = changeListener
        })
    })
}
```

Essentially, screen readers and keyboards trigger the "input" and "change" events of input type="checkbox" elements, and ignore the span and label elements. To get around this, a bunch of ARIA roles etc. would need to be added to the visual elements. Would be better to just listen for changes on the input and allow them to trigger/fire a click event, probably better yet, to just call the component method to update state directly rather than having to fire another event, which I don't know if it would have weird side effects anywhere.

chadlafontaine:
Changed the checkbox to listen for a change event instead of a just a
click event so that that using the keyboard spacebar will also trigger the checkbox change.

Fixes #3432 

## Proposed Changes
Fix b-table checkbox check event.
-
-
-

### DIFF
--- a/src/components/checkbox/Checkbox.vue
+++ b/src/components/checkbox/Checkbox.vue
@@ -5,7 +5,11 @@
         ref="label"
         :disabled="disabled"
         @click="focus"
-        @keydown.prevent.enter="$refs.label.click()">
+        @keydown.prevent.enter="$refs.label.click()"
+        @keydown.prevent.space="$refs.label.click()">
+        <!-- Checkbox needs to listen for a space event instead of a just a
+             click and enter event so that that using the keyboard spacebar will also
+             trigger the checkbox change in the b-table -->
         <input
             v-model="computedValue"
             :indeterminate.prop="indeterminate"
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -258,8 +258,11 @@
                                     autocomplete="off"
                                     :disabled="!isRowCheckable(row)"
                                     :value="isRowChecked(row)"
-                                    @click.native.prevent.stop="checkRow(row, index, $event)"
+                                    @change.native.prevent.stop="checkRow(row, index, $event)"
                                 />
+                                <!-- Checkbox needs to listen for a change event instead of a just a
+                                click event so that that using the keyboard spacebar will also
+                                trigger the checkbox change. -->
                             </td>
 
                             <template v-for="(column, colindex) in visibleColumns">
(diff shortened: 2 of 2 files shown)


jtommy:
To be honest the checkbox component (individually) supports keyboard and then we should only listen input event to listen the change. There is the problem to don't propogate the event in order to avoid to invoke row click event

chadlafontaine:
@jtommy The only solution that I could find was to add `@keydown.prevent.space="$refs.label.click()"` to the checkbox component, because I noticed pressing the enter key on b-table would also trigger the checkbox and `@keydown.prevent.enter="$refs.label.click()"` was already on the checkbox component. This seemed to solve my problem of using the spacebar to select a checkbox on b-table.

chadlafontaine:
@jtommy Any updates or feedback on this Pull Request?

jtommy:
@chadlafontaine I don't understand what your change should fix. The checkbox already works with space or enter (space is a native behaviour)

chadlafontaine:
@jtommy Yes that behavior does work correctly when the checkbox is by itself, but it does not work like that when the checkbox is used on the b-table. That is what this pull request will fix.

jtommy:
@chadlafontaine just tested and it works without your PR

chadlafontaine:
@jtommy I just tried it on this table https://buefy.org/documentation/table#checkable and when I tab to a checkbox and press the spacebar, the checkbox shows as selected but the value is not actually being selected. That is shown in the total checked value.

chadlafontaine:
@jtommy Thanks. Sorry for my lack of explanation in the beginning.

PR #3711, opened 2022-03-17
Pass click event to datepicker icon-right-click

punkt2:
To call, e.g. stopPropagation on the icon click event, pass the event to the icon-right-click


akicho8:
Thank you for the correction.
I was in trouble with just the same problem.
I hope this correction will be released quickly.

punkt2:
@akicho8 The fix is already available in the latest buefy version v0.9.20

akicho8:
@punkt2 Thank you for teaching me.
I misunderstood because there was no number (3711) in the release information
https://github.com/buefy/buefy/releases?q=%233711&expanded=true

PR #3712, opened 2022-03-18
Fix validation on input focus (#3705)

### LINKED ISSUE #3705: b-input shows validation error on focus instead of blur
### Overview of the problem

**Buefy** version: [0.9.19]
**Vuejs** version: [2.6.14]
**OS/Browser**: Chrome

### Description

On focus of any input with "required" attribute, it immediately shows an error.
Even on the docs it happens: https://buefy.org/documentation/input#validation
On the field called "user handle".

### Steps to reproduce

Create a `<b-input required />` and focus it.

### Expected behavior

The message is suppose to show on blur, not on focus as stated in the docs.

### Actual behavior

https://jsfiddle.net/qfLtb70n/

lucraraujo:
Fixes #3705

Remove HTML validation when onFocus event is triggered.

### DIFF
--- a/src/utils/FormElementMixin.js
+++ b/src/utils/FormElementMixin.js
@@ -114,7 +114,6 @@ export default {
         onFocus($event) {
             this.isFocused = true
             this.$emit('focus', $event)
-            this.checkHtml5Validity()
         },
 
         getElement() {


jtommy:
You are right, it doesn't make sense

Issue #3260, opened 2021-04-01
Numberinput: e.toFixed is not a function


fabiofdsantos:
### Overview of the problem

**Buefy** version: 0.9.5
**OS/Browser**: Facebook 301.0.0 (in-app browser), iOS 2.5.1

### Description

```
TypeError: e.toFixed is not a function. (In 'e.toFixed(this.stepDecimals)', 'e.toFixed' is undefined)
  at increment(/_nuxt/da83da5.js:1:336404)
  at increment([native code])
  at longPressTick(/_nuxt/da83da5.js:1:336580)
  at longPressTick([native code])
  at onStartLongPress(/_nuxt/da83da5.js:1:336849)
  at onStartLongPress([native code])
  at te(/_nuxt/0298e78.js:1:11753)
  at c(/_nuxt/da83da5.js:1:33111)
```

### Steps to reproduce

1. Input `0e2`
2. Press button to increment

ronnyvdweyer:
Same problem over here, i update to version 0.9.13
but this did not help.  
When you click first the - and then + it works. when you start with the + gives an error.

```
TypeError: e.toFixed is not a function
    at a.increment (buefy.min.js)
    at a.longPressTick (buefy.min.js)
    at a.onStartLongPress (buefy.min.js)
 ...
```

PR #3714, opened 2022-03-21
Numberinput: Parse float value before attempting toFixed

morganarnel:
This issue is potentially linked to #3260 - however I am unable to say for sure as the returned errors are slightly different. They do appear to be linked to the same increment method though.

I haven't created an issue for this as it's something I came across whilst deep in the zone at work - however when trying to increment a decimal value bound to a b-numberinput component I received the following console error:

```javascript
vue.common.dev.js:1902 TypeError: value.toFixed is not a function
    at VueComponent.increment (numberinput.js:202:1)
    at VueComponent.longPressTick (numberinput.js:213:1)
    at VueComponent.onStartLongPress (numberinput.js:221:1)
    at mousedown (numberinput.js:236:2155)
    at invokeWithErrorHandling (vue.common.dev.js:1868:1)
    at HTMLButtonElement.invoker (vue.common.dev.js:2193:1)
    at HTMLButtonElement.original._wrapper (vue.common.dev.js:7587:1)
```

Decrementing works fine, just incrementing it seems to fail.
I was a few versions behind but after updating the issue seems to persist. 

Below is my component just in case I'm being daft and missing something obvious. If so, then sorry for wasting your time!

```html
<b-numberinput
	:exponential=".5"
	:size="size"
	:controls="controls"
	:controls-alignment="controlsAlignment"
	:controls-position="controlsPosition"
	min="0"
	step=".01"
	@input="emitValue()"
	v-model="amount"
	:expanded="isExpanded">
</b-numberinput>
```

## Proposed Changes

- Attempt `parseFloat` before attempting a `.toFixed(...)`

### DIFF
--- a/src/components/numberinput/Numberinput.vue
+++ b/src/components/numberinput/Numberinput.vue
@@ -270,7 +270,7 @@ export default {
             }
             if (typeof this.minNumber === 'undefined' || (this.computedValue - this.stepNumber) >= this.minNumber) {
                 const value = this.computedValue - this.stepNumber
-                this.computedValue = parseFloat(value.toFixed(this.stepDecimals))
+                this.computedValue = parseFloat(value).toFixed(this.stepDecimals)
             }
         },
         increment() {
@@ -283,7 +283,7 @@ export default {
             }
             if (typeof this.maxNumber === 'undefined' || (this.computedValue + this.stepNumber) <= this.maxNumber) {
                 const value = this.computedValue + this.stepNumber
-                this.computedValue = parseFloat(value.toFixed(this.stepDecimals))
+                this.computedValue = parseFloat(value).toFixed(this.stepDecimals)
             }
         },
         onControlClick(event, inc) {


morganarnel:
i had a dumb moment, nevermind
sorry xoxox

wabit:
@morganarnel I seem to be having a simlar dumb moment and i cant for the life of me see whats wrong.

How did you fix your issue?

morganarnel:
> @morganarnel I seem to be having a simlar dumb moment and i cant for the life of me see whats wrong.
> 
> How did you fix your issue?

I honestly have no idea, sorry!

I originally put it down to some Docker Compose mounting trickery as after attempting a second rebuild without cache seemed to solve it.

Although if your code is failing at the same point within Buefy itself, we can safely assume that we don't have a float at the above highlighted line of code. Perhaps working backwards from there might highlight something on your end.

If not though and you think this PR may be beneficial I'm happy to reopen it. As I didn't have the issue anymore, it didn't seem right to suggest additions that might be unnecessary and just add another (granted minute) layer of processing.

PR #3738, opened 2022-05-10
[Feature Request]: Tooltip close delay

ievhen-soloviov:
## Proposed Changes

The tooltip component has a `delay` prop which is the time before the tooltip appears. Why not add the opposite: the time before the tooltip disappears? This way we have explicit `openDelay` and `closeDelay` props.


jtommy:
It's a breaking change, I cannot merge it

ievhen-soloviov:
@jtommy ah yes, of course, sorry. I brought back the original `delay`.

PR #3776, opened 2022-08-29
Add messages prop to message slot in BField

GMartigny:
## Proposed Changes

 - In the BField component, the slot `#message` now receive a `:messages` prop with computedMessage
  This allow users to easily wrap help messages.

```vue
<BField>
  <BInput />
  <template #message="{ message }">
    <strong class="custom-wrapper">{{  messages }}</strong>
  </template>
</BField>
```

### DIFF
--- a/docs/pages/components/field/api/field.js
+++ b/docs/pages/components/field/api/field.js
@@ -91,7 +91,7 @@ export default [
             {
                 name: '<code>message</code>',
                 description: 'Custom message',
-                props: '-'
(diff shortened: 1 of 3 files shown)


jtommy:
It's ok, but I don't understand the need.
Probably only when message comes from internal validation

GMartigny:
> It's ok, but I don't understand the need. Probably only when message comes from internal validation

Exactly ! I had the issue with `required` input.

PR #3777, opened 2022-08-31
Fix input type not updated when type prop is changed

GMartigny:
Fixes #

## Proposed Changes

- Adding a listener for changes in `type` property


GMartigny:
[Bug reproduction codepen](https://codepen.io/GMartigny/pen/JjvPdJP?editors=1010)

PR #3778, opened 2022-08-31
Modify docs

knot22:
An additional improvement to this page would be to include the reason why `import 'buefy/dist/buefy.css'` can be omitted (step 3 Import and use Buefy).  I do not know the reason why so cannot make that change.


jtommy:
It's not correct. Buefy css is required when you are not using scss

PR #3782, opened 2022-09-06
Add prop to disable Numberinput long press behavior

SylvainMarty:
## Proposed Changes
Our needs: In my company, we have some cases where we don't want the long press button behavior and we had no clean way to disable it.

- New property `disableLongPress` added to disable the long press logic
- Unit tests updated
- New property added to the API doc

I'd be happy to know if something's missing. :-)


jtommy:
It's ok, but you simply call it `long-press` (default value true)

SylvainMarty:
@jtommy Fixed ! 😄

PR #3787, opened 2022-09-13
Fixed the calculation of the checkIfReachedTheEndOfScroll when footer…

### LINKED ISSUE #3786: Infinite Scroll For Autocomplete doesn't work with using only arrow keys
### Overview of the problem

**Buefy** version: [0.9.22]
**Vuejs** version: [2.6.11]

### Description

When scrolling down the autocomplete list using just the arrow keys the infinite scroll check will not fire the event. It works using the mouse however.

### Steps to reproduce

1. focus the autcomplete
2. hit the down arrow till you hit the bottom of the list.

### Expected behavior

Should grab the next set of results.

### Actual behavior

Does nothing.

JDurman:
… is present and using arrow keys to navigate down the list.

Fixes #

Fixes #3786 
## Proposed Changes
Changed the calculation for infinite scroll to take into account the footer spacing as well.
-
-
-

### DIFF
--- a/src/components/autocomplete/Autocomplete.vue
+++ b/src/components/autocomplete/Autocomplete.vue
@@ -501,8 +501,9 @@ export default {
          * reached it's end.
          */
         checkIfReachedTheEndOfScroll(list) {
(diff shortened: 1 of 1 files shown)


jtommy:
You are right, it doesn't check the footer slot

PR #3788, opened 2022-09-15
Draft: Use the color-parser props for colorPicker value

GMartigny:
I made this PR to open conversation about the ColorPicker component:

1. When setting the `color-parser` prop in a ColorPicker, it's used only at the component definition (mounting). It feels like a bug, but I don't know if this is intentionnal.

2. Also I would like to point out that [throwing error on color parsing](https://github.com/buefy/buefy/blob/dev/src/utils/color.js#L332) is kinda annoying. This force the user to deal with uncaught exception bubbling.

3. One last thing, the `Color` class is not exported by Buefy. User needs to `import 'buefy/src/utils/color.js?module'` to access it, which is not trivial to guess.


jtommy:
About Color class you can add it in the index.js so the user can import using `import { Color } from 'buefy'`

PR #3790, opened 2022-09-21
Move sort and expand class to cell

GMartigny:
## What I try to improve

When using `b-table-column` with a custom header users have to manually add sort icon :
```vue
<b-table>
  <b-table-column>
    <template #header>
      // ...
      <b-icon icon="carret-down" class="sort-icon" />
    </template>
  </b-table-column>
</b-table>
```
At least, with this PR, user don't to handle sort direction.

## Proposed Changes

- Move the sort class `is-desc` and the expand class `is-expanded` to the cell. This allow user to rely on it even when using templates.

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -77,6 +77,9 @@
                             v-bind="column.thAttrs(column)"
                             :class="[column.thClasses, {
                                 'is-current-sort': !sortMultiple && currentSortColumn === column,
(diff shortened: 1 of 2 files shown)


jtommy:
It looks like a breaking change

GMartigny:
Yeah, I agree. Might be hard to merge

jtommy:
We can discuss about on https://github.com/oruga-ui/oruga repository 😉

PR #3815, opened 2022-12-18
Vue breakpoint mixin

ajomuch92:
## Proposed Changes

- Adding a new extension for breakpoing mixin


jtommy:
What do you think to ad an example?

ajomuch92:
@jtommy done, an example added

PR #3820, opened 2022-12-30
extending the future data to match past date

### LINKED ISSUE #3817: Buefy Date picker Max year range not following the max-datetime range given
**Buefy** version: [0.9.0
**Vuejs** version: [2.6.10]
**OS/Browser**: windows 11

### Description

When set a max-datetime duration something above 10 years component does not show year range above 10 years there is a limitation in the components to cut down the years in future max upto 10 years. 

This bug shows in your example codes as well. I have attached screenshot below. 

### Expected behavior

default past years-range is 100 from current year and default future year-range is 10 this should be 100 as well. Other wise component behaviour is not same as past and future. 
once set the max-datetime to 50 years user suppose to see all those years not limited to 10 years from the current year by default. 

as per my knowledge setting up years-range:[-100, 100] by default this error could be easily handled. 
Component needs similar behavior when picking dates in past and future. 

2022-12-20 10_48_50-Datepicker _ Buefy

rahulcool1910:
Fixes #3817

- Extending the limitation to only show 10 years from current year to 100 years.

### DIFF
--- a/src/utils/config.js
+++ b/src/utils/config.js
@@ -38,7 +38,7 @@ let config = {
     defaultUseHtml5Validation: true,
     defaultDropdownMobileModal: true,
     defaultFieldLabelPosition: null,
-    defaultDatepickerYearsRange: [-100, 10],
+    defaultDatepickerYearsRange: [-100, 100],
     defaultDatepickerNearbyMonthDays: true,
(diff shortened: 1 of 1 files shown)


wesdevpro:
Why did you close this PR @rahulcool1910?

PR #3822, opened 2023-01-10
fix: add .vue extension to component imports

daniil4udo:
This PR add `.vue` extension to the component imports.

## Reasons

- Vue CLI will stop supporting extension-less Vue imports in the next major ([see Evan's comment](https://github.com/vitejs/vite/issues/178#issuecomment-630138450))
- Vite doesnt work without extension already


daniil4udo:
@jtommy could you plz check & merge, trying to use it with vite & autoimports

PR #3830, opened 2023-01-27
Dedicated prop for keep-open on taginput

CobraClutch:
Allow the dropdown to be closed when selecting a tag. Default to true to preserve existing functionality.

### DIFF
--- a/src/components/taginput/Taginput.vue
+++ b/src/components/taginput/Taginput.vue
@@ -43,7 +43,7 @@
                 :loading="loading"
                 :autocomplete="nativeAutocomplete"
                 :open-on-focus="openOnFocus"
-                :keep-open="openOnFocus"
+                :keep-open="keepOpen"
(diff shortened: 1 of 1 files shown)


jtommy:
@CobraClutch to be honest the default value is false at the moment

CobraClutch:
Right, but the proposed change allows setting the prop to “true” instead. I think true should actually be the default, but I left it false so it wouldn’t effect any existing functionality for other users.   
  
  
  
  
  
  
  

  
  
>

jtommy:
Sure but it's a breaking change

CobraClutch:
I don’t believe so. By setting the default to false on the root component, existing projects won’t be affected if they update Buefy. Any taginput   instances will continue to function as before, with the dropdown staying open on select.   
  
  
  
  
  
  
  

  
  
>

PR #3832, opened 2023-02-08
Feature/new prop checkbox

ajomuch92:
Adding a new prop to set the input id on checkbox component

### DIFF
--- a/src/components/checkbox/Checkbox.vue
+++ b/src/components/checkbox/Checkbox.vue
@@ -12,7 +12,7 @@
              trigger the checkbox change in the b-table -->
         <input
             v-model="computedValue"
-            :id="innerInputId"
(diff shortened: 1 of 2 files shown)


src/components/checkbox/Checkbox.vue
```diff
@@ -12,6 +12,7 @@
              trigger the checkbox change in the b-table -->
         <input
             v-model="computedValue"
+            :id="innerInputId"
```
jtommy, line 15: Just `input-id`

jtommy (changes requested):

src/components/checkbox/Checkbox.vue
ajomuch92: @jtommy done

PR #3840, opened 2023-03-06
Added mobileModal to Clockpicker Component

julzELO:
No Issue found. Direct PR
## Proposed Changes

- This PR added the feature to directly set the prop mobile-modal on the Clockpicker and Timepicker Component. It's created parallel to the datepicker mobile-modal logic.
- updated docs clockpicker, datepicker and timepicker.

### DIFF
--- a/src/utils/config.js
+++ b/src/utils/config.js
@@ -32,6 +32,7 @@ let config = {
     defaultModalScroll: null,
     defaultDatepickerMobileNative: true,
     defaultTimepickerMobileNative: true,
+    defaultTimepickerMobileModal: true,
     defaultNoticeQueue: true,
     defaultInputHasCounter: true,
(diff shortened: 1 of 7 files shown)


src/utils/config.js
```diff
@@ -32,6 +32,7 @@ let config = {
     defaultModalScroll: null,
     defaultDatepickerMobileNative: true,
     defaultTimepickerMobileNative: true,
+    defaultTimepickerMobileModal: true,
```
jtommy, line 35: It's a double

jtommy (changes requested):

src/utils/config.js
julzELO: Hi @jtommy,

The constants are created parallel to the datepicker constants, same prefix and structure.

PR #3859, opened 2023-06-12
Fix: b-button nativeType for tag 'input'

mistersaal:
We cannot use nativeType for b-button with tag 'input' and type 'submit' after pull request with disabling nativeType for tag 'a' (https://github.com/buefy/buefy/pull/3733).
In this pull request I fixed it.


kikuomax (approved):
I have confirmed the fix with the following code:
```vue
<template>
  <div>
    <b-button tag="input" native-type="submit">Submit</b-button>
  </div>
</template>
```

PR #3863, opened 2023-07-09
feature(Autocomplete): add @active event whenever dropdown is active

daniil4udo:
### Proposed changes

Add emit for isActive

### DIFF
--- a/src/components/autocomplete/Autocomplete.vue
+++ b/src/components/autocomplete/Autocomplete.vue
@@ -316,6 +316,10 @@ export default {
                     }, 100)
                 }
             }
+
+            this.$nextTick(() => {
+                this.$emit('active', active)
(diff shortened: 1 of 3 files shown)


src/components/autocomplete/Autocomplete.vue
```diff
@@ -316,6 +316,10 @@ export default {
                     }, 100)
                 }
             }
+
+            this.$nextTick(() => {
```
kikuomax, line 320: @daniil4udo Do you have any specific reason for deferring emit to the next tick?

kikuomax: It works anyway.

kikuomax (approved):
Looks good to me!

PR #3865, opened 2023-07-19
fix(taginput): fix dropdown position when autocompleted and appended …

francisfontoura:
…to body

When Autocomplete is used as Taginput child, appended to body Dropdown position should be calculated from Taginput element as trigger.

Fixes #

## Proposed Changes

-
-
-

### DIFF
--- a/src/components/autocomplete/Autocomplete.vue
+++ b/src/components/autocomplete/Autocomplete.vue
@@ -664,7 +664,7 @@ export default {
         },
         updateAppendToBody() {
             const dropdownMenu = this.$refs.dropdown
-            const trigger = this.$refs.input.$el
+            const trigger = this.$parent.$data._isTaginput ? this.$parent.$el : this.$refs.input.$el
             if (dropdownMenu && trigger) {
                 // update wrapper dropdown
                 const root = this.$data._bodyEl


src/components/autocomplete/Autocomplete.vue
```diff
@@ -664,7 +664,7 @@ export default {
         },
         updateAppendToBody() {
             const dropdownMenu = this.$refs.dropdown
-            const trigger = this.$refs.input.$el
+            const trigger = this.$parent.$data._isTaginput ? this.$parent.$el : this.$refs.input.$el
```
wesdevpro, line 667: How does this fix the positioning problem?

francisfontoura: The appended to body dropdown is absolutely positioned and its bounds are calculated from `const trigger` element bounds. In standalone Autocomplete, align left dropdown with input makes sense, because input left bound remains fixed. But, when inside Taginput, after adding tags, the input left bound changes, causing the bug (I think this is, at least, a consistency bug, considering left position of not appended-to-body dropdown). So, it tests If it's inside Taginput (test already used inside component) and chooses right trigger accordingly.

kikuomax:
@francisfontoura Could you provide any example code that reproduces the issue?

kikuomax:
> @francisfontoura Could you provide any example code that reproduces the issue?

I reproduced the bug with the following code and finally got your intention:
```vue
<template>
    <div>
        <b-taginput :data="tags" append-to-body autocomplete></b-taginput>
    </div>
</template>

<script>
export default {
    data() {
        return {
            tags: [
                'Apple',
                'Banana',
                'Grape',
                'Kiwifruit',
                'Mango',
                'Orange',
                'Pineapple'
            ]
        }
    }
}
</script>
```

kikuomax (approved):
I have confirmed the bug fix.

Issue #3827, opened 2023-01-16
Add "always-render" option to modal


timwoelfle:
### Description

When setting a modal's "destroy-on-hide" option to false, it stays in the DOM when inactivated. However, this requires it to be active first - if it's initialized inactive, it's not in the DOM. I kindly ask for an option to make sure it's always in the DOM from initialization on, even when inactive. This may be called "always-render" or "render-on-initialization" or something like this.

### Why Buefy need this feature

Some use cases (mine at least ;-) ) require the modal to be in the DOM from the beginning (in my case so that vis.js network can add elements to it) -- yet I still want it to be inactive by default. You can see what I need on https://localcitationnetwork.github.io/?examples -- the modal opens when you click "network settings" on the top right. I'm currently circumventing the issue with an ugly hack, see below (from https://github.com/LocalCitationNetwork/LocalCitationNetwork.github.io/blob/master/index.js): I'm initializing the modals' binding variables "showCitationNetworkSettings" and "showAuthorNetworkSettings" as true (thus the modals are active) but setting their containers to visibility: hidden. After 1ms I inactive the modals and after 3s (when the fade-out is over) I set visibility to "visible".

```
// Hack: showCitationNetworkSettings and showAuthorNetworkSettings have to be true initially for modal containers to be created and available for vis.js network
// They have visibility: hidden initially
// Make network settings visible (they briefly flash up after loading the page so that the DOM elements are created for vis.js)
setTimeout(function () {
  vm.showCitationNetworkSettings = false
  vm.showAuthorNetworkSettings = false
}, 1)
setTimeout(function () {
  document.getElementById('citationNetworkSettingsModal').style.visibility = 'visible'
  document.getElementById('authorNetworkSettingsModal').style.visibility = 'visible'
}, 3000)
```

wesdevpro:
I will leave most feature requests up to @jtommy as my goal is just to maintain and fix bugs for Buefy v2. Thank you for your feature request. However, has your example changed? I don't see network settings on your website

timwoelfle:
Thanks for the update! The issue is still up to date, however the example link has indeed changed (I've updated the original post): When you go to https://localcitationnetwork.github.io/?examples, you see "network settings" in the top right right below "Citation network". I don't seem to be able to open this issue back up but I would love to see this implemented eventually!

wesdevpro:
[...]

PR #3880, opened 2023-09-05
Modal render on mounted (#3827)

kikuomax:
Fixes
- #3827

## Proposed Changes

- Introduce `renderOnMounted` prop for `Modal`. If it is `true`, `Modal` creates DOM for its content whether it is active or not.

### DIFF
--- a/docs/pages/components/modal/api/modal.js
+++ b/docs/pages/components/modal/api/modal.js
@@ -149,6 +149,13 @@ export default [
                 type: 'String',
                 values: '—',
                 default: '—'
+            },
+            {
+                name: '<code>render-on-mounted</code>',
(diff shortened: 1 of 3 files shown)


kikuomax:
Alternatively we could change the behavior of `destroyOnHide` so that DOM is created on mounted if `destroyOnHide` is `false`.

wesdevpro:
@kikuomax I just want to confirm that this PR is for Buefy and not Buefy-next

kikuomax:
@wesdevpro 
> @kikuomax I just want to confirm that this PR is for Buefy and not Buefy-next

Yes, this is for Buefy.

wesdevpro (approved):

PR #3881, opened 2023-09-05
Add nearbyMonthDays prop support to Datetimepicker

gmercey:
## Proposed Changes

- Add Datepicker component prop `nearby-month-days` to Datetimepicker component

### DIFF
--- a/src/components/datetimepicker/Datetimepicker.vue
+++ b/src/components/datetimepicker/Datetimepicker.vue
@@ -18,6 +18,7 @@
         :date-parser="defaultDatetimeParser"
         :min-date="minDate"
         :max-date="maxDate"
+        :nearby-month-days="nearbyMonthDays"
         :icon="icon"
         :icon-right="iconRight"
         :icon-right-clickable="iconRightClickable"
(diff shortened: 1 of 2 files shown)


src/components/datetimepicker/Datetimepicker.vue
```diff
@@ -145,6 +146,10 @@ export default {
         },
         minDatetime: Date,
         maxDatetime: Date,
+        nearbyMonthDays: {
+            type: Boolean,
+            default: true
```
kikuomax, line 151: To respect the default behavior of `Datepicker`, I think the following would be better:
```js
default: () => config.defaultDatepickerNearbyMonthDays
```
Or leave it `undefined` by default.

kikuomax:
@gmercey Could you also update the [docs](https://github.com/buefy/buefy/blob/dev/docs/pages/components/datetimepicker/api/datetimepicker.js)?

gmercey:
@kikuomax I have updated the prop default behavior and the doc

kikuomax (approved):

Issue #3896, opened 2023-10-10
Add loading to the button in diglog component.


zhanghengxin:
## How to use the feature.
```js
this.$buefy.dialog.prompt({
   ...
   onConfirm: (value, { close, startLoading, cancelLoading }) => {
          StartLoading();
          this.$buefy.toast.open(`Your message is sending...`);
          setTimeout(() => {
              this.$buefy.toast.open(`Success message send!`);
              // cancelLoading(); or close();
          }, 2000);
      }
  })
  
  // close function will destroys components, StopLoading is not necessary, but is added for semantic integrity.
```

PR #3897, opened 2023-10-10
🪶 :Add loading to the button in diglog component.

zhanghengxin:
Fixes #

resolve https://github.com/buefy/buefy/issues/3896

## Proposed Changes

- support loading state in dialog while confirmed.
- [Add unit test for dialog loading.](https://github.com/buefy/buefy/pull/3897/commits/7d21822b1c215d6edec6e18dc92d8cb7c723df39)
-

### DIFF
--- a/docs/pages/components/dialog/examples/ExPromptDialog.vue
+++ b/docs/pages/components/dialog/examples/ExPromptDialog.vue
@@ -5,19 +5,29 @@
                 label="Launch prompt (default)"
                 type="is-dark"
                 size="is-medium"
-                @click="prompt" />
+                @click="prompt"
+            />
 
             <b-button
                 label="Launch prompt (number)"
                 type="is-dark"
                 size="is-medium"
-                @click="promptNumber" />
+                @click="promptNumber"
+            />
 
             <b-button
                 label="Launch prompt (Not closing default)"
                 type="is-dark"
                 size="is-medium"
-                @click="promptNotClosed" />
+                @click="promptNotClosed"
+            />
+
+            <b-button
+                label="Launch prompt (Not closing with loading)"
+                type="is-dark"
+                size="is-medium"
+                @click="promptNotClosedWithLoading"
+            />
         </div>
     </section>
 </template>
@@ -29,47 +39,69 @@ export default {
             this.$buefy.dialog.prompt({
                 message: `What's your name?`,
                 inputAttrs: {
-                    placeholder: 'e.g. Walter',
+                    placeholder: "e.g. Walter",
                     maxlength: 10
                 },
                 trapFocus: true,
-                onConfirm: (value) => this.$buefy.toast.open(`Your name is: ${value}`)
-            })
+                onConfirm: value =>
+                    this.$buefy.toast.open(`Your name is: ${value}`)
+            });
         },
         promptNumber() {
             this.$buefy.dialog.prompt({
                 message: `What's your age?`,
                 inputAttrs: {
-                    type: 'number',
-                    placeholder: 'Type your age',
-                    value: '18',
+                    type: "number",
+                    placeholder: "Type your age",
+                    value: "18",
                     min: 18,
                     max: 99
                 },
                 trapFocus: true,
-                onConfirm: (value) => this.$buefy.toast.open(`Your age is: ${value}`)
-            })
+                onConfirm: value =>
+                    this.$buefy.toast.open(`Your age is: ${value}`)
+            });
         },
         promptNotClosed() {
             this.$buefy.dialog.prompt({
-                message: 'Send your message to moon 🚀',
+                message: "Send your message to moon 🚀",
                 inputAttrs: {
-                    type: 'text',
-                    placeholder: 'My message is...',
-                    value: 'Hello moon!'
+                    type: "text",
+                    placeholder: "My message is...",
+                    value: "Hello moon!"
                 },
-                confirmText: 'Send',
+                confirmText: "Send",
                 trapFocus: true,
                 closeOnConfirm: false,
-                onConfirm: (value, {close}) => {
-                    this.$buefy.toast.open(`Your message is sending...`)
+                onConfirm: (value, { close }) => {
+                    this.$buefy.toast.open(`Your message is sending...`);
                     setTimeout(() => {
-                        this.$buefy.toast.open(`Success message send!`)
-                        close()
-                    }, 2000)
+                        this.$buefy.toast.open(`Success message send!`);
+                        close();
+                    }, 2000);
                 }
-            })
+            });
+        },
+        promptNotClosedWithLoading() {
+            this.$buefy.dialog.prompt({
+                message: "Send your message to moon 🚀",
+                inputAttrs: {
+                    type: "text",
+                    placeholder: "My message is...",
+                    value: "Hello moon!"
+                },
+                confirmText: "Send",
+                trapFocus: true,
+                closeOnConfirm: false,
+                onConfirm: (value, { cancelLoading }) => {
+                    this.$buefy.toast.open(`Your message is sending...`);
+                    setTimeout(() => {
+                        this.$buefy.toast.open(`Success message send!`);
+                        cancelLoading();
+                    }, 2000);
+                }
+            });
         }
     }
-}
+};
 </script>
--- a/docs/pages/components/dialog/examples/ExPromptDialog.vue
+++ b/docs/pages/components/dialog/examples/ExPromptDialog.vue
@@ -73,7 +73,8 @@ export default {
                 confirmText: "Send",
                 trapFocus: true,
                 closeOnConfirm: false,
-                onConfirm: (value, { close }) => {
+                onConfirm: (value, { close, StartLoading }) => {
+                    StartLoading();
                     this.$buefy.toast.open(`Your message is sending...`);
                     setTimeout(() => {
                         this.$buefy.toast.open(`Success message send!`);
@@ -93,11 +94,12 @@ export default {
                 confirmText: "Send",
                 trapFocus: true,
                 closeOnConfirm: false,
-                onConfirm: (value, { cancelLoading }) => {
+                onConfirm: (value, { StartLoading, CancelLoading }) => {
+                    StartLoading();
                     this.$buefy.toast.open(`Your message is sending...`);
                     setTimeout(() => {
                         this.$buefy.toast.open(`Success message send!`);
-                        cancelLoading();
+                        CancelLoading();
                     }, 2000);
                 }
             });
--- a/docs/pages/components/dialog/examples/ExPromptDialog.vue
+++ b/docs/pages/components/dialog/examples/ExPromptDialog.vue
@@ -73,8 +73,8 @@ export default {
                 confirmText: "Send",
                 trapFocus: true,
                 closeOnConfirm: false,
-                onConfirm: (value, { close, StartLoading }) => {
-                    StartLoading();
+                onConfirm: (value, { close, startLoading }) => {
+                    startLoading();
                     this.$buefy.toast.open(`Your message is sending...`);
                     setTimeout(() => {
                         this.$buefy.toast.open(`Success message send!`);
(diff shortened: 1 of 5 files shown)


src/components/dialog/Dialog.vue
```diff
@@ -164,7 +165,8 @@ export default {
             prompt,
             isActive: false,
             validationMessage: '',
-            isCompositing: false
+            isCompositing: false,
+            Loading: false
```
kikuomax, line 169: @zhanghengxin @wesdevpro 
I am wondering if the name `Loading` is relevant, because the consequence of `confirm` is not necessarily "loading". How about `isConfirming` or `isProcessing`?

src/components/dialog/Dialog.spec.js
```diff
@@ -69,6 +69,19 @@ describe('BDialog', () => {
         expect(wrapper.vm.onConfirm).toHaveBeenCalled()
     })
 
+    it('close on async confirm', async () => {
+        wrapper.setProps({
+            onConfirm: jest.fn((confirm, {StartLoading}) => {
+                StartLoading()
```
kikuomax, line 75: @zhanghengxin @wesdevpro
I am thinking enabling the "Loading" behavior in `Dialog` by default should not harm. I mean clients does not have to call `StartLoading` themselves, but `Dialog` sets itself the "Loading" state whenever `confirm` is called. The "Loading" state will be cancelled when `close` or `cancelLoading` (or `cancelConfirm`) is called.

src/components/dialog/Dialog.vue
zhanghengxin: Whether it is in definition or development, isConfirming, isProcessing are unrelated to Loading. @kikuomax 

1、The original intent of Loading is to express the loading status of the confirm button. It is defined as a state machine and follows the old way of exposing the state machine change methods: startLoading, stopLoading.

2、There is a change to the line of code for isCompositing in the git record, which only added a comma.

3、Whether the Loading state changes is decided by the user, the user enables it through StartLoading. The default state of Loading is false.

I'm not sure if my replies have answered your question. Please let me know if you need any clarification or have additional questions.

src/components/dialog/Dialog.spec.js
zhanghengxin: It makes sense, it can completely be an internal state.

src/components/dialog/Dialog.vue
kikuomax: @zhanghengxin Thanks for clarification.

"Loading" makes sense in terms of the internal state of the confirm button. But how about if we want to, I think we should, disable the cancel button while "Loading" is on?

And how about "StartLoading" and "StopLoading" which are exposed to the user? It won't make sense if what the user is doing is not "loading".

zhanghengxin: 1、To maintain naming consistency and follow the naming rule aligned with isActive, change Loading to isLoading.

2、Then when Loading, change the status of the cancel button to disabled? Is this possible?

zhanghengxin: But, if the cancel button is disabled, there may be a case where - the user's loading result fails, should it close the dialog, or keep it in the dialog interface?

zhanghengxin: In other words, if we give the user control over the Loading state, then the user will be able to stay on the dialog interface when the loading result fails. I'm unsure if this violates the original intent of designing the dialog component.

zhanghengxin: Currently, even if there is a possibility of a second dialog, the component does not have the ability to change feedback content. Generally feedback is provided via the Toast component.

Therefore, I think I can define the Dialog component as designed for single use dialogs. Then I can simply handle it by disabling the cancel button when isLoading is true.

kikuomax: > But, if the cancel button is disabled, there may be a case where - the user's loading result fails, should it close the dialog, or keep it in the dialog interface?

It is up to the user whether the dialog is closed on failure. I mentioned `cancelLoading` (or `cancelConfirm`) method in [another comment](https://github.com/buefy/buefy/pull/3897#discussion_r1353830769) which was intended to be called on failure if the user wants to keep the dialog opened. The method will clear the "Loading" state.

zhanghengxin: done

wesdevpro: @kikuomax waiting your review 😄

wesdevpro (approved):
LGTM! @zhanghengxin wait for @kikuomax before we pull these changes into `dev`

wesdevpro (commented):
Do we need to also add documentation? @zhanghengxin @kikuomax

src/components/dialog/Dialog.spec.js
kikuomax: > @zhanghengxin @wesdevpro I think enabling the "Loading" behavior in `Dialog` by default should not harm. I mean clients do not have to call `StartLoading` themselves, but `Dialog` sets itself the "Loading" state whenever `confirm` is called. The "Loading" state will be cancelled when `close` or `cancelLoading` (or `cancelConfirm`) is called.

I realized I was wrong about the above comment. It may be a breaking change for some applications that intentionally keep the `Dialog` opened after the confirm button is clicked, which will end up with being stuck in the loading state unless they change their code.

@zhanghengxin Could you revert to your initial idea of calling `startLoading` explicitly?

zhanghengxin:
> Do we need to add or update documentation? @zhanghengxin @kikuomax

Done

src/components/dialog/Dialog.spec.js
zhanghengxin: @kikuomax 

I have restored it.

src/components/dialog/Dialog.vue
```diff
@@ -218,11 +222,26 @@ export default {
...
             setTimeout(() => {
                 this.$destroy()
                 removeElement(this.$el)
             }, 150)
+        },
+
+        /**
+        * Start the Loading.
+        */
+        StartLoading() {
```
kikuomax, line 236: `startLoading` please.

src/components/dialog/Dialog.vue
```diff
@@ -218,11 +222,26 @@ export default {
...
+            this.isLoading = true
+        },
+
+        /**
+        * Cancel the Loading.
+        */
+        CancelLoading() {
```
kikuomax, line 243: `cancelLoading` please.

src/components/dialog/Dialog.vue
zhanghengxin: done

src/components/dialog/Dialog.vue
zhanghengxin: done

docs/pages/components/dialog/examples/ExPromptDialog.vue
```diff
@@ -29,47 +39,71 @@ export default {
...
+                    value: "Hello moon!"
                 },
-                confirmText: 'Send',
+                confirmText: "Send",
                 trapFocus: true,
                 closeOnConfirm: false,
-                onConfirm: (value, {close}) => {
-                    this.$buefy.toast.open(`Your message is sending...`)
+                onConfirm: (value, { startLoading, cancelLoading }) => {
+                    startLoading();
```
kikuomax, line 98: @zhanghengxin Sorry for confusion, but I meant to remove this line from the library by the following comment:
> Could you revert to your initial idea of calling startLoading explicitly?

zhanghengxin: What's meaning？
What should I do?

kikuomax: Please remove the line 98: `startLoading();`

zhanghengxin: If there is no startLoading(), then there is no Loading

kikuomax: Sorry, I was wrong.

src/components/dialog/Dialog.vue
```diff
@@ -200,6 +203,7 @@ export default {
         * Call the onConfirm prop (function) and close the Dialog.
         */
         confirm() {
+            this.isLoading = true
```
kikuomax, line 206: Please remove this line 206: `this.isLoading = true`.

docs/pages/components/dialog/examples/ExPromptDialog.vue
zhanghengxin: hah，I'm glad to see even experts make mistakes sometimes. It makes me feel better about my own mistakes.

kikuomax: > Sorry, I was wrong. I was looking at a wrong file.

I meant: https://github.com/buefy/buefy/pull/3897/files#r1356019615

src/components/dialog/Dialog.spec.js
```diff
@@ -69,6 +69,38 @@ describe('BDialog', () => {
         expect(wrapper.vm.onConfirm).toHaveBeenCalled()
     })
 
+    it('async confirm and keep Loading state', async () => {
+        wrapper.setProps({
+            onConfirm: jest.fn((confirmValue, { startLoading }) => {
+                startLoading()
+                expect(wrapper.vm.isLoading).toBeTruthy()
+                return new Promise(() => {})
```
kikuomax, line 77: This line is unnecessary because `Dialog` discards the returned value from `onConfirm`.
Please see https://github.com/buefy/buefy/blob/f4c72b9d897b04c75d5a439c6de61a9ae66138dc/src/components/dialog/Dialog.vue#L212

src/components/dialog/Dialog.spec.js
```diff
@@ -69,6 +69,38 @@ describe('BDialog', () => {
...
+            }),
+            closeOnConfirm: false
+        })
+        expect(wrapper.vm.isLoading).toBeFalsy()
+        await wrapper.vm.confirm()
```
kikuomax, line 82: This `await` has no effect, because `confirm` method is "sync" and returns nothing.

src/components/dialog/Dialog.spec.js
```diff
@@ -69,6 +69,38 @@ describe('BDialog', () => {
...
+                startLoading()
+                expect(wrapper.vm.isLoading).toBeTruthy()
+                return new Promise(() => {
+                    cancelLoading()
+                })
+            }),
+            closeOnConfirm: false
+        })
+        expect(wrapper.vm.isLoading).toBeFalsy()
+        await wrapper.vm.confirm()
```
kikuomax, line 99: This `await` has no effect, because `confirm` method is "sync" and returns nothing.

src/components/dialog/Dialog.spec.js
zhanghengxin: If I remove the return statement, the new keyword will cause an error. Should I delete that entire line? Or is there another way to fix it?

zhanghengxin: Let's move on to another question. How can we simulate asynchronous requests?

kikuomax: > If I remove the return statement, the new keyword will cause an error. Should I delete that entire line? Or is there another way to fix it?

Please remove the entire line 77. It has no effect anyway.

zhanghengxin: done

src/components/dialog/Dialog.spec.js
```diff
@@ -69,6 +69,38 @@ describe('BDialog', () => {
...
+    })
+
+    it('async confirm and close Loading state', async () => {
+        wrapper.setProps({
+            onConfirm: jest.fn((confirmValue, { startLoading, cancelLoading }) => {
+                startLoading()
+                expect(wrapper.vm.isLoading).toBeTruthy()
+                return new Promise(() => {
+                    cancelLoading()
+                })
```
kikuomax, line 94: Above three lines are almost equivalent to just calling `cancelLoading()`.

kikuomax: It is OK since it works, though, could you replace the three lines with a simple line:
```
cancelLoading()
```

zhanghengxin: @kikuomax 

Done

Is this what you wanted?

kikuomax: Perfect!

kikuomax (approved):

Issue #3912, opened 2023-12-07
Issue on - postcss-import: @charset must precede all other statements


mary-josephine-1:
### Overview of the problem

**OS/Browser**: Ubuntu 22.04.3 LTS  - Firefox

**Packages**: package.json
> {
  "name": "testApp",
  "version": "1.0.0",
  "private": true,
  "scripts": {
    "dev": "nuxt",
    "build": "nuxt build",
    "start": "nuxt start",
    "generate": "nuxt generate",
    "lint:js": "eslint --ext \".js,.vue\" --ignore-path .gitignore .",
    "lint:prettier": "prettier --check .",
    "lint": "npm run lint:js && npm run lint:prettier",
    "lintfix": "prettier --write --list-different . && npm run lint:js -- --fix"
  },
  "dependencies": {
    "@nuxtjs/axios": "^5.13.6",
    "@nuxtjs/pwa": "^3.3.5",
    "core-js": "^3.25.3",
    "nuxt": "^2.15.8",
    "nuxt-buefy": "^0.4.24",
    "vue": "^2.7.10",
    "vue-server-renderer": "^2.7.10",
    "vue-template-compiler": "^2.7.10"
  },
  "devDependencies": {
    "@babel/eslint-parser": "^7.19.1",
    "@nuxtjs/eslint-config": "^11.0.0",
    "@nuxtjs/eslint-module": "^3.1.0",
    "eslint": "^8.24.0",
    "eslint-config-prettier": "^8.5.0",
    "eslint-plugin-nuxt": "^4.0.0",
    "eslint-plugin-vue": "^9.5.1",
    "prettier": "^2.7.1"
  }
}

**Config**: nuxt.config.js

> export default {
  // Disable server-side rendering: https://go.nuxtjs.dev/ssr-mode
  ssr: false,
  // Global page headers: https://go.nuxtjs.dev/config-head
  head: {
    title: 'testApp',
    meta: [
      { charset: 'utf-8' },
      { name: 'viewport', content: 'width=device-width, initial-scale=1' },
      { hid: 'description', name: 'description', content: '' },
      { name: 'format-detection', content: 'telephone=no' },
    ],
    link: [{ rel: 'icon', type: 'image/x-icon', href: '/favicon.ico' }],
  },
  // Global CSS: https://go.nuxtjs.dev/config-css
  css: [],
  // Plugins to run before rendering page: https://go.nuxtjs.dev/config-plugins
  plugins: [],
[...]

PR #3913, opened 2023-12-07
fix(lib): update `build:scss` script

wesdevpro:
Fixes:
     #3912 

## Proposed Changes
- Update the `build:scss` script that compiles scss files and minifies the resulting CSS file. The `build:scss` script inserts the banner before compiling scss files, because some frameworks, e.g., Nuxt, complain about the `@charset "UTF-8"` not being at the top of the CSS file. FYI: `@charset "UTF-8"` is prepended by `sass` command.


kikuomax (approved):
Looks good to me!

PR #3963, opened 2024-01-10
Fix check to display Autocomplete

GMartigny:
Display Autocomplete list if footer slot is registered

## Proposed Changes

- Add a check for footer slot in Autocomplete list display condition

### DIFF
--- a/src/components/autocomplete/Autocomplete.vue
+++ b/src/components/autocomplete/Autocomplete.vue
@@ -31,7 +31,7 @@
                 class="dropdown-menu"
                 :class="{ 'is-opened-top': isOpenedTop && !appendToBody }"
                 :style="style"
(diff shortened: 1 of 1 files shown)


kikuomax:
@GMartigny I have tested the behavior with the following code:

```vue
<template>
    <div class="container is-max-desktop">
        <b-field label="Only footer">
            <b-autocomplete :data="[]">
                <template #footer>
                    <a><span>Add new...</span></a>
                </template>
            </b-autocomplete>
        </b-field>
    </div>
</template>
```

**Without** your PR: no dropdown showed up

**With** your PR: the dropdown with "Add new..." footer showed up

Thanks for your PR! Please attach any reproduction code next time.

kikuomax (approved):
Looks good to me!

PR #3991, opened 2024-01-15
Clear event listeners and timers before component unmounts

Arooba-git:
Hi 👋 , thank you for this library!

We found that some components used event listeners and timing events, but did not remove them before unmounting. This means an accumulation of listeners and timers each time a component is mounted, on top of the previous ones, leading to a memory leak. This PR fixes this issue :)

### DIFF
--- a/docs/components/TheNavbar.vue
+++ b/docs/components/TheNavbar.vue
@@ -133,6 +133,10 @@ export default {
     },
     mounted() {
         this.$eventHub.$on('navigate', this.closeMenu)
+    },
+
+    beforeDestroy() {
+        this.$eventHub.$off('navigate', this.closeMenu)
     }
 }
 </script>
--- a/docs/templates/Documentation.vue
+++ b/docs/templates/Documentation.vue
@@ -62,6 +62,10 @@ export default {
         if (this.$route.hash) {
             this.$nextTick(() => this.scrollTo(this.$route.hash))
         }
+    },
+
+    beforeDestroy() {
+        this.$eventHub.$off('navigate', this.setMeta)
     }
 }
 </script>
--- a/src/components/sidebar/Sidebar.vue
+++ b/src/components/sidebar/Sidebar.vue
@@ -295,6 +295,7 @@ export default {
         if (this.isFixed) {
             removeElement(this.$el)
         }
+        clearTimeout(this.timer)
     }
 }
 </script>
--- a/src/components/slider/SliderThumb.vue
+++ b/src/components/slider/SliderThumb.vue
@@ -222,6 +222,14 @@ export default {
                 this.oldValue = value
             }
         }
+    },
+
+    beforeDestroy() {
+        document.removeEventListener('mousemove', this.onDragging)
+        document.removeEventListener('touchmove', this.onDragging)
+        document.removeEventListener('mouseup', this.onDragEnd)
+        document.removeEventListener('touchend', this.onDragEnd)
+        document.removeEventListener('contextmenu', this.onDragEnd)
     }
 }
 </script>
--- a/src/components/numberinput/Numberinput.vue
+++ b/src/components/numberinput/Numberinput.vue
@@ -312,6 +312,10 @@ export default {
             clearTimeout(this._$intervalRef)
             this._$intervalRef = null
         }
+    },
+
+    beforeDestroy() {
+        clearTimeout(this._$intervalRef)
     }
 }
 </script>


wesdevpro:
@Arooba-git thanks for identifying this problem and fixing it. @kikuomax and I will review these changes as soon as possible. Thanks again!

wesdevpro (approved):
@kikuomax all these changes make sense to me. However, I have not tested for any memory leaks and or if there are no longer any memory leaks. Is there a way for us to test that?

src/components/carousel/Carousel.vue
```diff
@@ -345,6 +345,7 @@ export default {
     },
     beforeDestroy() {
         this.pauseTimer()
+        clearInterval(this.timer)
```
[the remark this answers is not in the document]
wesdevpro: Oops I do see that now.

Arooba-git:
@kikuomax Good that you pointed that out. Yes, actually we use [Memlab](https://facebook.github.io/memlab/docs/getting-started) to gauge heap size and memory leaks in applications. Memlab requires us to write a scenario file as mentioned in their documentation. I created a scenario file for this project as well. Sadly currently memlab has a limitation of a _protocolTimeout_ error for some cases, this one falls into that. I have reported the issue, they are working to fix within this week. After that I will attach the memory footprint results before and after the leak fixes in this project.

src/components/sidebar/Sidebar.vue
```diff
@@ -295,6 +295,7 @@ export default {
         if (this.isFixed) {
             removeElement(this.$el)
         }
+        clearTimeout(this.timer)
```
kikuomax, line 298: Same as https://github.com/buefy/buefy/pull/3991/files#r1452848601.

src/components/slider/SliderThumb.vue
```diff
@@ -222,6 +222,14 @@ export default {
...
             }
         }
+    },
+
+    beforeDestroy() {
+        document.removeEventListener('mousemove', this.onDragging)
+        document.removeEventListener('touchmove', this.onDragging)
+        document.removeEventListener('mouseup', this.onDragEnd)
+        document.removeEventListener('touchend', this.onDragEnd)
+        document.removeEventListener('contextmenu', this.onDragEnd)
```
kikuomax, line 232: I do not see any problem with the original code. But there might be some edge cases I overlooked. So it will be safer to have these lines here.

docs/components/TheNavbar.vue
```diff
@@ -133,6 +133,10 @@ export default {
     },
     mounted() {
         this.$eventHub.$on('navigate', this.closeMenu)
+    },
+
+    beforeDestroy() {
+        this.$eventHub.$off('navigate', this.closeMenu)
```
kikuomax, line 139: I confirm this should be done.

docs/templates/Documentation.vue
```diff
@@ -62,6 +62,10 @@ export default {
         if (this.$route.hash) {
             this.$nextTick(() => this.scrollTo(this.$route.hash))
         }
+    },
+
+    beforeDestroy() {
+        this.$eventHub.$off('navigate', this.setMeta)
```
kikuomax, line 68: I confirm this should be done.

Arooba-git:
> I have reported the issue, they are working to fix within this week. 

I heard back from Memlab developer regarding the timeout and page crash issue I was facing when executing their tool on the scenario file I wrote for Buefy. 
Screen Shot 2024-01-18 at 1 22 13 PM

This is what their developer has to say:

> My guess is that either 1) 1.4GB of memory is too much for Chrome to capture the heap snapshot (you could try taking the 1.4GB heap snapshot manually in Chrome DevTools to see if the Browser DevTools times out) or 2) the connection between Puppeteer and Chrome was not designed to handle such a large heap snapshot.
> 
> Moreover, if a web app uses more than 1GB JS heap, it has some serious memory issues. I would recommend testing on a scenario where the memory leak is less severe.

The code patterns fixed in this PR are some of the common sources of memory leaks, as per our research on Github and StackOverflow history. Nevertheless, if I manage to resolve the Memlab crash issue and provide concrete heap/leak reduction after these fixes, I will let you know as soon as is possible..

kikuomax:
@Arooba-git What kind of scenario are you running? 1.4GB of heap usage by Buefy does not make sense to me. If just browsing Buefy docs used 1.4GB of heap, we should have gotten a lot of serious problem reports.

kikuomax:
@Arooba-git By the way, could you update your PR if my comments make sense to you?

Arooba-git:
> What kind of scenario are you running? 

TBH It's a pretty normal scenario file that follows puppeteer syntax, similar to what I have been using for multiple other projects 🤔 . I have attached it here for reference:

[test-scenario-buefy.js.zip](https://github.com/buefy/buefy/files/13973609/test-scenario-buefy.js.zip)

Arooba-git:
> @Arooba-git By the way, could you update your PR if my comments make sense to you?

Sure, totally makes sense. I'll try to update by today 👍

kikuomax:
> > What kind of scenario are you running?
> 
> TBH It's a pretty normal scenario file that follows puppeteer syntax, similar to what I have been using for multiple other projects 🤔 . I have attached it here for reference:
> 
> [test-scenario-buefy.js.zip](https://github.com/buefy/buefy/files/13973609/test-scenario-buefy.js.zip)

@Arooba-git Does the script suppose the Buefy docs app running at `http://localhost:8080`; i.e., running `npm run dev`?

Arooba-git:
> Does the script suppose the Buefy docs app running at `http://localhost:8080`; i.e., running `npm run dev`?

Yes, we need to first run the app and pass the URL to Memlab's script accordingly. In my case (when I created this script), the app was running on the said address.

kikuomax (approved):
Looks good to me!

kikuomax:
@Arooba-git @wesdevpro I am going to merge this PR and close the conversation. Please update us in the new issue #3992.

Issue #3992, opened 2024-01-18
[Need more info] Huge memory leak?


kikuomax:
> I have reported the issue, they are working to fix within this week. 

I heard back from Memlab developer regarding the timeout and page crash issue I was facing when executing their tool on the scenario file I wrote for Buefy. 
<img width="478" alt="Screen Shot 2024-01-18 at 1 22 13 PM" src="https://github.com/buefy/buefy/assets/56495631/7d2eeac9-c8bc-42aa-8063-8216e65afcb8">

This is what their developer has to say:

> My guess is that either 1) 1.4GB of memory is too much for Chrome to capture the heap snapshot (you could try taking the 1.4GB heap snapshot manually in Chrome DevTools to see if the Browser DevTools times out) or 2) the connection between Puppeteer and Chrome was not designed to handle such a large heap snapshot.
> 
> Moreover, if a web app uses more than 1GB JS heap, it has some serious memory issues. I would recommend testing on a scenario where the memory leak is less severe.

The code patterns fixed in this PR are some of the common sources of memory leaks, as per our research on Github and StackOverflow history. Nevertheless, if I manage to resolve the Memlab crash issue and provide concrete heap/leak reduction after these fixes, I will let you know as soon as is possible..

_Originally posted by @Arooba-git in https://github.com/buefy/buefy/issues/3991#issuecomment-1897860167_

PR #4000, opened 2024-01-20
Clear timing events and event listeners before component unmounts

Arooba-git:
In reference to issue [#3992](https://github.com/buefy/buefy/issues/3992), I was able to measure memory foot prints before and after the leak patterns's removal. I reduced the elements that were triggered to 20 links, instead of triggering all the links, that helped prevent the Memlab crash (mentioned in the issue). 

While the leak counts discovered by Memlab remained the same, the heap size saw a noticeable reduction:
**Before**
buefy-heap-before

**After**
buefy-heap-after

### DIFF
--- a/src/components/dialog/Dialog.vue
+++ b/src/components/dialog/Dialog.vue
@@ -167,8 +167,7 @@ export default {
             isActive: false,
             validationMessage: '',
             isCompositing: false,
-            isLoading: false,
-            timeOutID: null
+            isLoading: false
         }
     },
     computed: {
@@ -224,7 +223,7 @@ export default {
             this.isActive = false
             this.isLoading = false
             // Timeout for the animation complete before destroying
-            this.timeOutID = setTimeout(() => {
+            setTimeout(() => {
                 this.$destroy()
                 removeElement(this.$el)
             }, 150)
@@ -270,10 +269,6 @@ export default {
                 this.$refs.confirmButton.$el.focus()
             }
         })
-    },
-
-    beforeDestroy() {
-        clearTimeout(this.timeOutID)
     }
 }
 </script>
--- a/src/components/dialog/Dialog.vue
+++ b/src/components/dialog/Dialog.vue
@@ -167,7 +167,8 @@ export default {
             isActive: false,
             validationMessage: '',
             isCompositing: false,
-            isLoading: false
+            isLoading: false,
+            timeOutID: null
         }
     },
     computed: {
@@ -223,7 +224,7 @@ export default {
             this.isActive = false
             this.isLoading = false
             // Timeout for the animation complete before destroying
-            setTimeout(() => {
+            this.timeOutID = setTimeout(() => {
                 this.$destroy()
                 removeElement(this.$el)
             }, 150)
@@ -269,6 +270,10 @@ export default {
                 this.$refs.confirmButton.$el.focus()
             }
         })
+    },
+
+    beforeDestroy() {
+        clearTimeout(this.timeOutID)
     }
 }
 </script>
--- a/src/components/loading/Loading.vue
+++ b/src/components/loading/Loading.vue
@@ -47,7 +47,8 @@ export default {
     data() {
         return {
             isActive: this.active || false,
-            displayInFullPage: this.isFullPage
+            displayInFullPage: this.isFullPage,
+            timeOutID: null
         }
     },
     watch: {
@@ -78,7 +79,7 @@ export default {
             // Timeout for the animation complete before destroying
             if (this.programmatic) {
                 this.isActive = false
-                setTimeout(() => {
+                this.timeOutID = setTimeout(() => {
                     this.$destroy()
                     removeElement(this.$el)
                 }, 150)
@@ -116,6 +117,7 @@ export default {
         if (typeof window !== 'undefined') {
             document.removeEventListener('keyup', this.keyPress)
         }
+        clearTimeout(this.timeOutID)
     }
 }
 </script>
--- a/src/components/loading/Loading.vue
+++ b/src/components/loading/Loading.vue
@@ -47,8 +47,7 @@ export default {
     data() {
         return {
             isActive: this.active || false,
-            displayInFullPage: this.isFullPage,
-            timeOutID: null
+            displayInFullPage: this.isFullPage
         }
     },
     watch: {
@@ -79,7 +78,7 @@ export default {
             // Timeout for the animation complete before destroying
             if (this.programmatic) {
                 this.isActive = false
-                this.timeOutID = setTimeout(() => {
+                setTimeout(() => {
                     this.$destroy()
                     removeElement(this.$el)
                 }, 150)
@@ -117,7 +116,6 @@ export default {
         if (typeof window !== 'undefined') {
             document.removeEventListener('keyup', this.keyPress)
         }
-        clearTimeout(this.timeOutID)
     }
 }
 </script>
--- a/src/components/modal/Modal.vue
+++ b/src/components/modal/Modal.vue
@@ -151,7 +151,8 @@ export default {
                 ? this.width + 'px'
                 : this.width,
             animating: !this.active,
-            destroyed: !(this.active || this.renderOnMounted)
+            destroyed: !(this.active || this.renderOnMounted),
+            timeOutID: null
         }
     },
     computed: {
@@ -240,7 +241,7 @@ export default {
             // Timeout for the animation complete before destroying
             if (this.programmatic) {
                 this.isActive = false
-                setTimeout(() => {
+                this.timeOutID = setTimeout(() => {
                     this.$destroy()
                     removeElement(this.$el)
                 }, 150)
@@ -305,6 +306,7 @@ export default {
             document.documentElement.scrollTop = savedScrollTop
             document.body.style.top = null
         }
+        clearTimeout(this.timeOutID)
     }
 }
 </script>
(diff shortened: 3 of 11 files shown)


kikuomax:
@Arooba-git Thanks for the update. Calling `clearTimeout` will prevent potential access to unmounted components. However, it should not contribute to reducing memory leaks in the real world scenarios, because the resource allocated for `setTimeout` must be released when the timeout comes.

kikuomax:
@Arooba-git I came up with a possible cause of the leak.
How about to change the following line:
https://github.com/buefy/buefy/blob/7a0c69055be6fdbcb2519e820386f3916b1a089f/docs/main.js#L104-L106

to:
```js
const loader = document.addEventListener('DOMContentLoaded', function () {
    root.$mount('#app')
    document.removeEventListener('DOMContentLoaded', loader)
})
```

Arooba-git:
> However, it should not contribute to reducing memory leaks in real world scenarios, because the resource allocated for `setTimeout` must be released when the timeout expires.

That is correct. The case we're targeting (and I think what might make it add to the memory leakage) is if the user is viewing a component, a timeout has started after the component mounts, but they navigate away _before_ the timeout got a chance to be complete. Sounds plausible?

Arooba-git:
> @Arooba-git I came up with a possible cause of the leak. How about to change the following lines:
> 
> https://github.com/buefy/buefy/blob/7a0c69055be6fdbcb2519e820386f3916b1a089f/docs/main.js#L104-L106
> 
> to:
> 
> ``` js
> const loader = document.addEventListener('DOMContentLoaded', function () {
>     root.$mount('#app')
>     document.removeEventListener('DOMContentLoaded', loader)
> })
> ```

Hmm, I never thought about this kind of resolution. The reason is similar to my other comment: the event listener is attached to the `document `on the component mount (or any action), but once the event is attached, it depends on whether or not the _event_ is called that the listener can be removed.. What if the user's action does _not_ trigger the event and they navigate away? In such a scenario, the callback will not get a chance to be executed right (and hence no removal of listener either)..

Arooba-git:
[This might be a bit of help] If you run Memlab, you be able to see the entire [retainer trace](https://facebook.github.io/memlab/docs/how-memlab-works/#3-retainer-traces-for-memory-leaks) for all the reported leaks i.e all the objects (along with their IDs in heap)  in the path from the actual leaking object to the root object. You can then use the browser's dev tools to check these object IDs one by one in the [heap-snap shot](https://developer.chrome.com/docs/devtools/memory-problems/heap-snapshots), to identify where, in the chain of objects, is the actual leak-causing object. Sometimes it's a global array that needs to be cleared, it may even be a third-party library that is leaky. Again, you can check it by manually analyzing those objects, in your captured heap snapshots,  that are present in Memlab's results:
Screen Shot 2024-01-19 at 11 42 44 PM

kikuomax:
> Hmm, I never thought about this kind of resolution. The reason is similar to my other comment: the event listener is attached to the `document` on the component mount (or any action), but once the event is attached, it depends on whether or not the _event_ is called that the listener can be removed.. What if the user's action does _not_ trigger the event and they navigate away? In such a scenario, the callback will not get a chance to be executed right (and hence no removal of listener either)..

@Arooba-git [`DOMContentLoaded`](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event) does not require any user interaction. It is always[^1] notified when a document is loaded.

[^1]: There might be [some edge cases](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event#checking_whether_loading_is_already_complete) in which `DOMContentLoaded` has been notified before `main.js` is executed. But this is another issue.

kikuomax:
@Arooba-git @wesdevpro I have started looking into memlab. I misunderstood [how memlab detects leaks](https://facebook.github.io/memlab/docs/how-memlab-works) and found [my fix](https://github.com/buefy/buefy/pull/4000#issuecomment-1902271924) was pointless.

kikuomax:
> > However, it should not contribute to reducing memory leaks in real world scenarios, because the resource allocated for `setTimeout` must be released when the timeout expires.
> 
> That is correct. The case we're targeting (and I think what might make it add to the memory leakage) is if the user is viewing a component, a timeout has started after the component mounts, but they navigate away _before_ the timeout got a chance to be complete. Sounds plausible? (I think in even Memlab scenario file I did not add any code to wait for a certain action to be completed so it navigates away (clicks multiple links) pretty quickly)

@Arooba-git Now I understood `clearTimeout` matters to leaks in terms of memlab.

kikuomax:
@Arooba-git @wesdevpro I tweaked your scenario so that we can evaluate one page at once and do more granular analysis. https://github.com/kikuomax/buefy/tree/memlab-scenarios/memlab-scenarios

For instance,
`memlab run --scenario memlab-scenarios/test-documentation-start.js` did not report any leak.

but `memlab run --scenario memlab-scenarios/test-documentation-tag.js` reported leaks:
image

kikuomax:
@Arooba-git @wesdevpro I found even the `/documentation/start` page caused leaks depending on which element we wait for before leaving the page; this time I waited for `.codeview` that is deeply nested in the start page.

wesdevpro:
> @Arooba-git @wesdevpro I found even the `/documentation/start` page caused leaks depending on which element we wait for before leaving the page; this time I waited for `.codeview` instead of `.title`, which is deeply nested in the start page. Waiting for `.title` may not give enough time to mount a Vue component of interest, because an element with `.title` is included in the header existing in all the documentation pages.
> 
> Interestingly, `memlab` still detected memory leaks even if I replaced `docs/pages/installation/Start.vue` with the following simple component and waited for `.start-container`:
> 
> ```
> <template>
>     <div class="start-container">
>         Hello!
>     </div>
> </template>
> ```

Does Vue naturally have memory leaks?

src/components/dialog/Dialog.vue
```diff
@@ -269,6 +270,10 @@ export default {
                 this.$refs.confirmButton.$el.focus()
             }
         })
+    },
+
+    beforeDestroy() {
+        clearTimeout(this.timeOutID)
```
kikuomax, line 276: @Arooba-git This should not be necessary. Because `timeOutID` is set only if `Dialog` has programmatically been opened (`this.programmatic` is `true`) and a programmatically opened `Dialog` won't be destroyed until the timeout expires. Please see the line 228.

src/components/loading/Loading.vue
```diff
@@ -116,6 +117,7 @@ export default {
         if (typeof window !== 'undefined') {
             document.removeEventListener('keyup', this.keyPress)
         }
+        clearTimeout(this.timeOutID)
```
kikuomax, line 120: @Arooba-git This should not be necessary. Because `timeOutID` is set only if `Loading` has programmatically been opened (`this.programmatic` is `true`) and programmatically opened `Loading` won't be destroyed until the timeout expires. Please see the line 83.

src/components/modal/Modal.vue
```diff
@@ -305,6 +306,7 @@ export default {
             document.documentElement.scrollTop = savedScrollTop
             document.body.style.top = null
         }
+        clearTimeout(this.timeOutID)
```
kikuomax, line 309: @Arooba-git This should be unnecessary, because `timeOutID` is set only if `Modal` has programmatically been opened (`this.programmatic` is `true`) and programmatically opened `Modal` won't be destroyed until the timeout expires. Please see the line 245.

src/components/notification/NotificationNotice.vue
```diff
@@ -31,12 +32,16 @@ export default {
...
+                this.timeOutID = setTimeout(() => {
                     this.$destroy()
                     removeElement(this.$el)
                 }, 150)
             }
         }
+    },
+
+    beforeDestroy() {
+        clearTimeout(this.timeOutID)
```
kikuomax, line 44: @Arooba-git This should be unnecessary. Because `NotificationNotice` is intended to be programmatically opened and it won't be destroyed until the timeout expires. Please see the line 36.

src/components/tooltip/Tooltip.vue
```diff
@@ -295,7 +297,9 @@ export default {
                     this.updateAppendToBody()
                     animation.removeEventListener('transitionend', listener)
                 }
-                animation.addEventListener('transitionend', listener)
+                animation.addEventListener('transitionend', listener, {
+                    signal: this.controller.signal
```
kikuomax, line 301: @Arooba-git I have learned [`signal` option](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#signal). Thanks!

src/components/dialog/Dialog.vue
kikuomax: @Arooba-git I have reverted the changes in your fork.

src/components/loading/Loading.vue
kikuomax: @Arooba-git I have reverted the changes in your fork.

src/components/modal/Modal.vue
kikuomax: @Arooba-git I have reverted the changes in your fork.

src/components/notification/NotificationNotice.vue
kikuomax: @Arooba-git I have reverted the changes in your fork.

kikuomax (approved):
Looks good to me!
We are going to merge and close this PR, though, continue investigation in Issue #3992.

wesdevpro (approved):
⚡️

glumb:
> Looks good to me! We are going to merge and close this PR, though, continue investigation in Issue #3992.

Just for clarification and planning purposes, is PR #4000 / #3992 still a blocker for the upcoming release, or will it be merged and released shortly? :)

kikuomax:
@glumb @wesdevpro We are going to release v0.9.28.
- #4009

PR #4019, opened 2024-03-04
Fix: Tooltip - AbortController is not defined in SSR mode (Nuxt) (#4017)

### LINKED ISSUE #4017: Tooltip - `AbortController` is not defined in SSR mode (Nuxt)
**Buefy** version: [0.9.28]
**Nuxt** version: [2.17.3]

### Description

In the last update of the **Tooltip** component - there is an addon to remove animation on `beforeDestroy` event.
The problem is that it is using `window.AbortController()` without checking are `window` and/or `AbortController` defined.
Both of them are not defined when using **Nuxt with SSR mode**.

### Code source
https://github.com/buefy/buefy/blob/a13d8445ec263cbb817820b45a7ac56b5e7f5648/src/components/tooltip/Tooltip.vue#L288
https://github.com/buefy/buefy/blob/a13d8445ec263cbb817820b45a7ac56b5e7f5648/src/components/tooltip/Tooltip.vue#L334

First - when `AbortController` is registered (row 288) in SSR mode it will be just undefined.
Second -  when using the `abort` method (row 334) without checking it is defined - is causing the error:
```
TypeError
Cannot read property 'abort' of undefined
```

### Expected behavior
Use `AbortController` only after check if `window` and/or `AbortController` are defined.

### Actual behavior

`AbortController` is used without checking for `window` object and/or `AbortController` instance which cause errors.

kikuomax:
Fixes
- fixes #4017 

## Proposed Changes

- In `Tooltip`, move `window.AbortController` access inside the `window` availability check
- Prepare for `controller` possibly `null`

### DIFF
--- a/src/components/tooltip/Tooltip.vue
+++ b/src/components/tooltip/Tooltip.vue
@@ -285,8 +285,8 @@ export default {
         }
     },
     mounted() {
-        this.controller = new window.AbortController()
         if (this.appendToBody && typeof window !== 'undefined') {
+            this.controller = new window.AbortController()
             this.$data._bodyEl = createAbsoluteElement(this.$refs.content)
             this.updateAppendToBody()
             // updates the tooltip position if the tooltip is inside
@@ -331,7 +331,9 @@ export default {
         if (this.appendToBody) {
             removeElement(this.$data._bodyEl)
         }
-        this.controller.abort()
+        if (this.controller != null) {
+            this.controller.abort()
+        }
         clearTimeout(this.timer)
         clearTimeout(this.timeOutID)
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Buefy Changelog
 
+## 0.9.29 Unreleased
+
+### Fixes
+
+* Fix [#4017](https://github.com/buefy/buefy/issues/4017) `Tooltip` - AbortController is not defined in SSR mode (Nuxt)
+
 ## 0.9.28
 
 ### New features


kikuomax:
Stays draft until we confirm it works with Nuxt 2.

src/components/tooltip/Tooltip.vue
```diff
@@ -331,7 +331,9 @@ export default {
         if (this.appendToBody) {
             removeElement(this.$data._bodyEl)
         }
-        this.controller.abort()
+        if (this.controller != null) {
```
miroslavgeorgiev, line 334: we need:
`this.controller != undefined || this.controller != null`

If `window` is not accessible - `controller` would be undefined, too - not null

kikuomax: Thanks for your review. But we do not have to be as strict as you suggested:
1. `this.controller != null` turns `false` if `this.controller` is either `null` or `undefined`. Please see [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null#difference_between_null_and_undefined).
2. `this.controller` is safely evaluated as `undefined` even if `this.controller` has not been assigned. [`typeof` check is only useful if the variable (`this`) may not be defined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#typeof_operator_and_undefined), but `this` is always defined in this context. So the `typeof` check is redundant.

Thus, the current check is fine.

miroslavgeorgiev: With loose equality check `==` I guess you're right 👍 
I always prefer strict equality check `===` and in that case this won't work.

And I also noticed we have in the code `typeof window !== 'undefined'` not `window != null` right?

miroslavgeorgiev (approved):

src/components/tooltip/Tooltip.vue
kikuomax: Since `window` variable may not exist, e.g., in SSR, `typeof window !== 'undefined'` comes in.

wesdevpro (approved):
Looks great to me!

PR #4021, opened 2024-03-09
Fix: Browser tab freezes, unresponsive, plus possible memory leak with form in modal (#4018)

### LINKED ISSUE #4018: Browser tab freezes, unresponsive, plus possible memory leak with form in modal
### Overview of the problem

**Buefy** version: 0.9.28
**Vuejs** version: 2.7.16
**OS/Browser**: Ubuntu 22, Firefox, Brave

### Description

Whenever the modal opens and captures focus or when you start typing into the form fields, the tab will freeze.

### Steps to reproduce

1. Navigate to https://jsbin.com/baluzoquvo/edit?html,output
2. Open your browser's "Task Manager". (If you can, mark the window to stay ontop of other windows for better access)
3. Click on "Launch card modal (Component)" button to open the modal
4. If you are lucky, the tab will immediately freeze. If not, simply just start typing into the input fields, pressing tab to advance to the next field. By the time you get to the "Phone" field, your tab will freeze. Another way is to click into any field the first time the modal is open. It seems like the issue has to do with field focus inside the modal?
5. Observe the CPU and memory usage of the tab. Clicking in the unresponsive fields will also increase the memory usage even higher. 

### Expected behavior

Tab should not lock up or have memory spike

### Actual behavior

Inspecting CPU usage in the browser's Task Manager shows memory usage growing rapidly and CPU usage stays at 100%

kikuomax:
Fixes #
- fixes #4018

## Proposed Changes

- `Field` makes sure that the contents of a new value for `newMessage` are different from the current `newMessage` before updating it to avoid infinite recursions that made a browser tab freeze
- Add warnings of not mixing `grouped` and `horizontal` props of `Field` to the documentation

## Remarks

Since `Field` was not designed to turn on `grouped` and `horizontal` props simultaneously, another issue will emerge after merging this PR, where once a validation error state is set, the error state won't be reset even after correcting the input. A workaround for now is not to mix `grouped` and `horizontal`. This PR will mitigate the worst scenario anyway.

### DIFF
--- a/docs/pages/components/field/api/field.js
+++ b/docs/pages/components/field/api/field.js
@@ -41,7 +41,7 @@ export default [
             },
             {
                 name: '<code>grouped</code>',
-                description: 'Direct child components/elements of Field will be grouped horizontally (see which ones at the top of the page)',
+                description: 'Direct child components/elements of Field will be grouped horizontally (see which ones at the top of the page). Do not mix with <code>horizontal</code>.',
                 type: 'Boolean',
                 values: '—',
                 default: '<code>false</code>'
@@ -69,7 +69,7 @@ export default [
             },
             {
                 name: '<code>horizontal</code>',
-                description: 'Group label and control on the same line for horizontal forms',
+                description: 'Group label and control on the same line for horizontal forms. Do not mix with <code>grouped</code>.',
                 type: 'Boolean',
                 values: '—',
                 default: '<code>false</code>'
--- a/docs/pages/components/field/api/field.js
+++ b/docs/pages/components/field/api/field.js
@@ -41,7 +41,7 @@ export default [
             },
             {
                 name: '<code>grouped</code>',
-                description: 'Direct child components/elements of Field will be grouped horizontally (see which ones at the top of the page). Do not mix with <code>horizontal</code>.',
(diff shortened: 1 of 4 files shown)


docs/pages/components/field/api/field.js
```diff
@@ -69,7 +69,7 @@ export default [
             },
             {
                 name: '<code>horizontal</code>',
-                description: 'Group label and control on the same line for horizontal forms',
+                description: 'Group label and control on the same line for horizontal forms. Do not mix with <code>grouped</code>.',
```
wesdevpro, line 72: Is there a more clear way to say this? And should we also explain, in a simple form, why we shouldn't mix them?

wesdevpro (changes requested):

docs/pages/components/field/api/field.js
kikuomax: I briefly explained the reason why we should not mix them. However, there would not be a reasonable explanation because it is a design flaw we have to address.

wesdevpro (approved):

docs/pages/components/field/api/field.js
wesdevpro: DM over discord. Maybe we could discuss this flaw over the weekend?

wesdevpro (approved):

PR #4038, opened 2024-05-23
TOAST: change type declaration and update doc

GorbunovAlex:
Working with Toast component and using it outside the component noticed that trying to invoke it as stated in docs results in a `TypeError: Toast.open is not a function`

Reproduction: https://codesandbox.io/p/live/1d49f415-fba6-4a78-93da-316fc1c22f49

## Proposed Changes

- change type declaration of toast to class
- change information in documentation

### DIFF
--- a/docs/pages/components/toast/Toast.vue
+++ b/docs/pages/components/toast/Toast.vue
@@ -33,7 +33,7 @@
                 ExSimpleCode,
                 outsideVueInstance: `
                 import { ToastProgrammatic as Toast } from 'buefy'
-                Toast.open('Toasty!')`
+                new Toast().open('Toasty!')`
             }
         }
     }
--- a/types/components.d.ts
+++ b/types/components.d.ts
@@ -387,8 +387,8 @@ export declare const SnackbarProgrammatic: {
 }
 
 
-export declare const ToastProgrammatic: {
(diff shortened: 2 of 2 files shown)


wesdevpro:
Ths Toast component in Buefy vue2 (not Buefy-next) is not a class but an object literal 

https://github.com/buefy/buefy/blob/f6fe38ac2305e37a0fa33d4b7814a943e838a770/src/components/toast/index.js#L9

what may need to happen to fix

wesdevpro:
@GorbunovAlex  something similar to the source below would be the correct change for Buefy vue2

https://stackoverflow.com/questions/12787781/type-definition-in-object-literal-in-typescript

wesdevpro (changes requested):

dagnelies:
I wonder if there is a confusion in this PR. Shouldn't that go to https://github.com/ntohq/buefy-next/ instead?

wesdevpro:
> I wonder if there is a confusion in this PR. Shouldn't that go to https://github.com/ntohq/buefy-next/ instead?

Hmm I don't think so (See: https://github.com/buefy/buefy/pull/4038#discussion_r1721989786 ) Why should this PR be at buefy-next?

PR #4044, opened 2024-06-24
Add conditional rendering to Collapse component

cwilby:
## Proposed Changes

This pull request adds the ability to use a named `content` slot in the `BCollapse` component. This feature provides developers with the flexibility to delay rendering of the collapsible content until the `BCollapse` is open, which can improve performance when the content is complex or data-heavy (such as loading one or more images in every collapsible component in a list, or a complex iteration template that can cause performance degradation).

## Impact
No breaking changes to existing usage. Developers can opt-in to the feature by using the `content` slot instead of the default slot for deferred content rendering.  

## Testing
- Added unit tests for new slot behavior in `BCollapse.spec.js`.
- Ensured all existing tests pass with no regressions.

### DIFF
--- a/docs/pages/components/collapse/Collapse.vue
+++ b/docs/pages/components/collapse/Collapse.vue
@@ -9,6 +9,8 @@
         <Example :component="ExPosition" :code="ExPositionCode" title="Position" vertical/>
 
         <Example :component="ExAccordion" :code="ExAccordionCode" title="Accordion Effect" vertical/>
+        
+        <Example :component="ExConditionalRendering" :code="ExConditionalRenderingCode" title="Conditional Rendering" vertical/>
 
         <ApiView :data="api"/>
     </div>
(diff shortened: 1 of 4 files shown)


cwilby:
Open to closing this since there is a way to accomplish this without modifying Buefy.

```vue
<template>
  <b-collapse>
    <template #trigger="props">
      <div class="card-header" role="button" :aria-controls="'contentIdForA11y5-' + fieldIndex" :aria-expanded="props.open">
        <span>
          {{ fieldIsOpen = props.open }}
        </span>
      </div>
      <div class="card-body">
        <div>Always renders</div>
        <div v-if="fieldIsOpen">
          Only renders when open
          <img src="https://placehold.it/200x200">
        </div>
      </div>
    </template>
  </b-collapse>
</template>

<script>
  export default {
    data() {
      return {
        fieldIsOpen: false
      }
    }
  }
</script>
```

PR #4045, opened 2024-07-09
Change order to prevent slot from messing with focus

GMartigny:
Fixes #

## Proposed Changes

- Change order of tag inside Upload component

### DIFF
--- a/src/components/upload/Upload.vue
+++ b/src/components/upload/Upload.vue
@@ -1,5 +1,14 @@
 <template>
     <label class="upload control" :class="{'is-expanded' : expanded, 'is-rounded' : rounded}">
+        <input
+            ref="input"
+            type="file"
+            v-bind="$attrs"
+            :multiple="multiple"
+            :accept="accept"
+            :disabled="disabled"
+            @change="onFileChange">
+
         <template v-if="!dragDrop">
             <slot/>
         </template>
@@ -19,15 +28,6 @@
             @drop.prevent="onFileChange">
             <slot/>
         </div>
-
-        <input
-            ref="input"
-            type="file"
-            v-bind="$attrs"
-            :multiple="multiple"
-            :accept="accept"
-            :disabled="disabled"
-            @change="onFileChange">
     </label>
 </template>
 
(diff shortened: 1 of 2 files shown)


GMartigny:
Bug reproduction :
1. Go to https://codepen.io/GMartigny/pen/eYwmOwx?editors=1010
2. Add a file to the Upload component
3. See that you can update the file, because the focus goes to the text input

wesdevpro:
> Bug reproduction :
> 
> 1. Go to https://codepen.io/GMartigny/pen/eYwmOwx?editors=1010
> 2. Add a file to the Upload component
> 3. See that you can update the file, because the focus goes to the text input

@kikuomax the bug was reproducible on my end.

wesdevpro:
> @GMartigny @kikuomax the bug was reproducible on my end.

@GMartigny do you know if this is also ab issue with Buefy-next?

GMartigny:
> @GMartigny do you know if this is also an issue with Buefy-next?

It is, since tags are in the same order there too. https://github.com/ntohq/buefy-next/blob/dev/packages/buefy-next/src/components/upload/Upload.vue#L7-L9

wesdevpro:
> > @GMartigny do you know if this is also an issue with Buefy-next?
> 
> 
> 
> It is, since tags are in the same order there too. https://github.com/ntohq/buefy-next/blob/dev/packages/buefy-next/src/components/upload/Upload.vue#L7-L9

That makes sense. Could you also make a PR there as well?

wesdevpro (approved):
I have not tested this component but logically this makes sense and the code changes look good. Please wait for @kikuomax to review. I will also take sometime to test these changes to see if I find any abnormal behaviors from the component

kikuomax (approved):
@GMartigny I have confirmed your PR fixes the issue.

Btw, I was wondering why the order mattered. According to the [MDN documentation about `<label>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label):
> When a user clicks or touches/taps a label, the browser passes the focus to its associated input (the resulting event is also raised for the input).

I am not 100% sure about the behavior of `<label>` when it contains multiple nested `<input>` elements, though, it makes sense if the first `<input>` gets the focus.

Another possible workaround might be to add the `for="unique-upload-id"` attribute to `<label>`, and the `id="unique-upload-id"` to `<input>`. However, we have to make sure `unique-upload-id` globally unique to adopt this approach.

GMartigny:
> Btw, I was wondering why the order mattered.

As you expected, when a `<label>` has multiple focusable element inside, the first one get it.

Also, yes, it would be possible (I guess) to generate random id an attribute it to the `<input type="upload">` and `<label>`. With a large enough cryptographic randomness, collision chances are pretty much 0.

Issue #4130, opened 2024-01-08
Autocomplete does not allow us to turn `appendToBody` on after mounted


kikuomax:
### Overview of the problem

**Buefy** version: [[0.1.3-5c420e09933677691b004d437c7c7c684070b972](https://github.com/orgs/ntohq/packages/npm/buefy-next/159834561)]
**Vuejs** version: [3.3.4]
**OS/Browser**: macOS/Safari

### Description

Autocomplete does not allow us to turn `appendToBody` on after mounted.

### Steps to reproduce

1. Mount `<b-autocomplete>` with `appendToBody=false`
2. Turn `appendToBody` into `true`
3. Type something in the text field

```vue
<template>
    <section>
        <div class="block">
            <b-switch v-model="appendToBody">Append to body</b-switch>
        </div>
        <p class="content"><b>Selected:</b> {{ selected }}</p>
        <b-field label="Find a JS framework">
            <b-autocomplete
                rounded
                v-model="name"
                :data="filteredDataArray"
                placeholder="e.g. jQuery"
                icon="magnify"
                clearable
                @select="option => selected = option"
                :append-to-body="appendToBody"
            >
                <template #empty>No results found</template>
            </b-autocomplete>
        </b-field>
    </section>
</template>

<script>
export default {
    data() {
        return {
            data: [
                'Angular',
                'Angular 2',
                'Aurelia',
                'Backbone',
                'Ember',
                'jQuery',
                'Meteor',
                'Node.js',
                'Polymer',
                'React',
                'RxJS',
                'Vue.js'
            ],
            name: '',
            selected: null,
            appendToBody: false
        }
    },
    computed: {
        filteredDataArray() {
            return this.data.filter((option) => {
                return option
                    .toString()
                    .toLowerCase()
                    .indexOf(this.name.toLowerCase()) >= 0
            })
        }
    }
}
</script>
```

### Expected behavior

Autocomplete appends the dropdown to the body.

### Actual behavior

Autocomplete crashed with the following error:
```
[Error] Unhandled Promise Rejection: TypeError: undefined is not an object (evaluating 'root.classList')
	updateAppendToBody (Autocomplete.vue:581)
	（anonymous関数） (Autocomplete.vue:456)
```

Issue #4098, opened 2024-06-07
BAutoComplete arrow key selection doesn't work with computed data


hchamre:
### Overview of the problem

**Buefy** version: buefy-next 0.1.3
**Vuejs** version: 3.4.27
**OS/Browser**: 

### Description

When using a computed property as the data source for `BAutoComplete`, the arrow key selection doesn't work. The issue seems to be that the autocomplete component's selection variable is wrapped in a proxy, which changes how the item selected matches with those in the suggested list.

### Steps to reproduce

1. Create a computed property that returns an array of objects.
2. Use the computed property as the data source for `BAutoComplete`.
3. Try to navigate through the suggestions using the arrow keys.

### Expected behavior

The arrow key selection should work as expected.

### Actual behavior

The arrow key selection doesn't work:

https://github.com/buefy/buefy/assets/108055897/252674de-11f7-4084-a184-b2c0e019271f

### Code

<details>
  <summary><code>src/App.vue</code>:</summary>

```vue
<script setup>
import { computed, ref } from 'vue'

function getFruits() {
  return [
    { name: 'Apple', id: 1 },
    { name: 'Banana', id: 2 },
    { name: 'Orange', id: 3 },
    { name: 'Mango', id: 4 },
    { name: 'Grapes', id: 5 }, 
  ];
}

const suggestionsAsRef = ref(getFruits());

const suggestionsAsComputed = computed(getFruits);
</script>

<template>
  <section @keydown="onKeyDown">
    <!-- works -->
    <b-field label="Suggestions as ref">
      <b-autocomplete id="autocomplete" :data="suggestionsAsRef" field="name" />
    </b-field>
    <!-- doesn't work -->
    <b-field label="Suggestions as computed">
      <b-autocomplete id="autocomplete" :data="suggestionsAsComputed" field="name" />
    </b-field>
  </section>
</template>
```

PR #4139, opened 2025-06-15
FIX various bugs on Autocomplete component

gmercey:
## Fixes
- https://github.com/buefy/buefy/issues/4098
- https://github.com/buefy/buefy/issues/4130
- Autocomplete selection and dropdown logic

## Improvements
- recompute whitelist dynamically
- update ariaAutocomplete when keepFirst changes
- adjust scroll listener when appending to body
- test Autocomplete for 0 values, clearOnSelect, whitelist updates, etc.

### DIFF
--- a/packages/buefy/src/components/autocomplete/Autocomplete.vue
+++ b/packages/buefy/src/components/autocomplete/Autocomplete.vue
@@ -442,6 +442,25 @@ export default defineComponent({
                     }
                 }
             }
+        },
+
+        /*
+         * When appendToBody property changes, handle the transition properly
+         */
+        appendToBody(newValue, oldValue) {
+            if (newValue && !oldValue) {
+                // Changing from false to true - need to create _bodyEl if dropdown is active
+                if (this.isActive && this.$refs.dropdown && !this.$data._bodyEl) {
+                    this.$data._bodyEl = createAbsoluteElement(this.$refs.dropdown as Element)
+                    this.updateAppendToBody()
+                }
+            } else if (!newValue && oldValue) {
+                // Changing from true to false - need to clean up _bodyEl
+                if (this.$data._bodyEl) {
+                    removeElement(this.$data._bodyEl)
+                    this.$data._bodyEl = undefined
+                }
+            }
         }
     },
     methods: {
@@ -758,6 +777,10 @@ export default defineComponent({
                 ? this.$parent!.$el
                 : (this.$refs.input as BInputComponent).$el
             if (dropdownMenu && trigger) {
+                // Ensure _bodyEl exists before trying to update it
+                if (!this.$data._bodyEl) {
+                    this.$data._bodyEl = createAbsoluteElement(dropdownMenu)
+                }
                 // update wrapper dropdown
                 const root = this.$data._bodyEl!
                 root.classList.forEach((item) => root.classList.remove(item))
@@ -816,8 +839,8 @@ export default defineComponent({
             const list = (this.$refs.dropdown as Element).querySelector('.dropdown-content')!
             list.removeEventListener('scroll', this.checkIfReachedTheEndOfScroll)
         }
-        if (this.appendToBody) {
-            removeElement(this.$data._bodyEl!)
+        if (this.appendToBody && this.$data._bodyEl) {
+            removeElement(this.$data._bodyEl)
         }
         clearTimeout(this.timeOutID)
     }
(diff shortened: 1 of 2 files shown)


dagnelies:
This seems to break key navigation with normal async data

gmercey:
> This seems to break key navigation with normal async data

how did you get the broken behavior? I tested the component manually and the key navigation works in all cases.

gmercey:
> Here: https://deploy-preview-4139--buefy-org.netlify.app/documentation/autocomplete#async-with-custom-template
> 
> I can only navigate the two first elements with the up/down keys... depending on what I type it the search. Sometimes it works, sometimes not, which is really fishy. It looks like long lists are more likely to be buggy. The same happens with both async demos.

Though one to debug, but here's an update

packages/buefy/src/components/autocomplete/Autocomplete.vue
```diff
@@ -35,7 +35,10 @@
                 class="dropdown-menu"
                 :class="{ 'is-opened-top': isOpenedTop && !appendToBody }"
                 :style="style"
-                v-show="isActive && (!isEmpty || hasEmptySlot || hasHeaderSlot || hasFooterSlot)"
+                v-show="
```
dagnelies, line 38: unusual formatting ...I found it more intuitive on a single line ...just nitpicking here, you can ignore this if you want to

packages/buefy/src/components/autocomplete/Autocomplete.vue
```diff
@@ -125,17 +128,17 @@ import CompatFallthroughMixin from '../../utils/CompatFallthroughMixin'
...
 // `$data` interface of `Taginput` component.
 // we cannot directly import `Taginput` because it depends on `Autocomplete`.
 interface TaginputData {
-    _isTaginput: boolean
+    _isTaginput: boolean;
 }
 
 interface DataItem {
-    group?: string
+    group?: string;
```
dagnelies, line 139: most of this codebase favors lines without semicolons, so I wonder why you added those, it makes it more "heterogenous". Again, it's just nitpicking here, you can ignore this if you want to

packages/buefy/src/components/autocomplete/Autocomplete.vue
```diff
@@ -253,28 +256,35 @@ export default defineComponent({
...
             const whiteList = []
-            whiteList.push((this.$refs.input as BInputComponent).$el.querySelector('input'))
+            whiteList.push(
+                (this.$refs.input as BInputComponent).$el.querySelector('input')
+            )
             whiteList.push(this.$refs.dropdown)
-            // Add all children from dropdown
             if (this.$refs.dropdown != null) {
-                const children = (this.$refs.dropdown as Element).querySelectorAll('*')
+                const children = (
```
dagnelies, line 277: Actually, those formatting changes disturb reviewing because real changes are interleaved in stylistic ones. Perhaps it would be wise to minimize format changes and stick to the current formatting, and just commit the changes in "logic" next time. That would make it easier / faster for the reviewer ...just nitpicking, no biggy.

packages/buefy/src/components/autocomplete/Autocomplete.vue
```diff
@@ -397,12 +396,12 @@ export default defineComponent({
...
             // Check if selected is invalid
             const currentValue = this.getValue(this.selected)
-            if (currentValue && currentValue !== value) {
+            if (currentValue !== undefined && currentValue !== null && currentValue !== value) {
                 this.setSelected(null, false)
             }
             // Close dropdown if input is clear or else open it
-            if (this.hasFocus && (!this.openOnFocus || value)) {
-                this.isActive = !!value
+            if (this.hasFocus && (!this.openOnFocus || value !== '')) {
```
dagnelies, line 426: ...that's what I meant ...logic changes appear in the middle of similar "formatting changes". What happens if the new `value` is `null`. Before `isActive` would be left unchanged, and now it would become `false`. It's just not clear if it's an intended change or a mishap of the stylistic change. Personaly I also prefer the more readable `!!value` over the long `value !== '' && value !== undefined && value !== null` which are not completely equivalent. What happens if an autofill is filled with various numbers instead of strings? (dunno if that's possible though)

dagnelies (approved):
I'll approve this. Most of my comments are just nitpicking but due to formatting changes the logic changes were kind of difficult to spot. I hope this doesn't introduce subtle bugs in edge cases with autocomplete values such as `null`, `''` or `0` as side effect.

packages/buefy/src/components/autocomplete/Autocomplete.vue
gmercey: I believe my prettier or linter is acting out

dagnelies (approved):
LGTM. 🎉 Thank you for your contribution! 👍

PR #4140, opened 2025-06-15
Autocomplete optimization

gmercey:
Improve performance and stability in Autocomplete component with race condition prevention and throttling

## Proposed Changes
- **Race Condition Prevention:** Fixed timeout management in the `isActive` watcher to prevent race conditions
- **Error Handling:** Added robust error handling for DOM operations with try/catch blocks and fallback values
- **Performance Optimization:** Implemented scroll event throttling using `_scrollThrottled` flag
- **Memory Leak Prevention:** Enhanced cleanup in `beforeUnmount` to properly clear timeouts and event listeners
- **Type Safety:** Improved event handling with proper type checking for event targets


dagnelies (changes requested):
Same as before: breaks keyboard navigation with async data ;)

PR #4141, opened 2025-06-15
FIX #4120 - bSelect required not working

### LINKED ISSUE #4120: Select required isn't working
### Overview of the problem

**Buefy** version: [0.2-next]
**Vuejs** version: [3.5.13]
**OS/Browser**: edge

### Description

the tag required doen't work anymore on a b-select.

### Steps to reproduce

If you use the website https://v3.buefy.org/documentation/select, and you make dirty the Required select available, it will not turn red (and in my code doesn't prevent the form to be submitted)

If you go to the documentation of buefy https://buefy.org/documentation/select you can see that before, making dirty the select field will turn it red and prevent the form submitting.

### Expected behavior

Image

### Actual behavior

Image

gmercey:
Fixes #4120

The `required` attribute on b-select was not triggering HTML5 validation properly. When a user focused on a required select field with a placeholder and then blurred without selecting a value, the field would not turn red (show invalid state) and would not prevent form submission.

- **Placeholder Value Issue:** The placeholder option was using :value="null" which doesn't create a proper empty string value in the DOM needed for HTML5 validation.
- **Initial Value Handling:** The component was initializing with null instead of an empty string when a placeholder was present, which is incompatible with HTML5 validation for required fields.
- **Validation Trigger Issues:** The validation was not consistently re-triggered when values changed.


dagnelies:
Once #4142 is merged, the diffs should drastically reduce

dagnelies:
Also, I didn't delve into details, but here is a general question. A common use case in selects, is to use an empty option (using empty string `""` as value) to denote some default. Would the current PR pose a problem using such an empty option or not?

wesdevpro:
@gmercey Can you resolve these conflicts?

PR #4143, opened 2025-06-20
FIX timepickermixin unmount

gmercey:
TimepickerMixin was using an invalid `beforeUumounted` hook to remove an eventListener.


wesdevpro:
@gmercey is this PR still in progress?

PR #4150, opened 2025-06-25
Added ref to clockpicker face in the Clockpicker Component

julzELO:
## Proposed Changes

This change makes it possible to access the clockpickerface and use its variables such as isDragging.

This is useful when closing and switching current values through the dragging info of the clockpicker face.

### DIFF
--- a/packages/buefy/src/components/clockpicker/Clockpicker.vue
+++ b/packages/buefy/src/components/clockpicker/Clockpicker.vue
@@ -121,6 +121,7 @@
                             </div>
                         </div>
                         <b-clockpicker-face
+                            ref="clockpickerFace"
                             :picker-size="faceSize"
                             :min="minFaceValue"
                             :max="maxFaceValue"


dagnelies (approved):
This is pretty harmless, so I'll approve ....but I think you should describe a little better, with an example perhaps, why / how it should be used.

julzELO:
> This is pretty harmless, so I'll approve ....but I think you should describe a little better, with an example perhaps, why / how it should be used.

any idears how to example/write docs to a ref in buefy?

wesdevpro:
> > This is pretty harmless, so I'll approve ....but I think you should describe a little better, with an example perhaps, why / how it should be used.
> 
> any idears how to write example/docs to a ref in buefy?

Yes. I would recommend coming up with a scenario that is simple for using the ref then writing a component for that example. After doing so add that component to the docs for the clock picker section.

wesdevpro:
@julzELO any update on this?

PR #4161, opened 2025-07-11
Update Discord Links

### LINKED ISSUE #4155: Outdated Discord Invite Links
### Overview of the problem

**Buefy** version: [X.X.X]
**Vuejs** version: [X.X.X]
**OS/Browser**:

wesdevpro:
Fixes #4155

### What's Changed?
- Updated the discord link in the navbar
- Updated the discord link in the READMEs


wesdevpro:
> Should we remove the twitter link as it is not used? And add the discord link in the footer in place of twitter?

How about we update the footer social links to include all three? I think it would be best to keep the twitter account as I still have access to it and can post announcement updates there alongside discord and the announcement channel here on GitHub.

gmercey:
Then you may want to use the X icon instead

wesdevpro:
> Then you may want to use the X icon instead

We can us the following classes from font awesome.

`fa fa-brands fa-x-twitter`

PR #4174, opened 2025-07-20
Publish Buefy v1.0.2

wesdevpro:


dagnelies (commented):
Related to the last commit, I'm personally not fond of ads. They provide mediocre revenue and often drive traffic off the site to some related products. While I agree money is a key to a long-term sustainable maintenance, I don't think ads is the right way to do so. It's a challenging problem though

wesdevpro:
> They provide mediocre revenue and often drive traffic off the site to some related products. While I agree money is a key to a long-term sustainable maintenance, I don't think ads is the right way to do so. It's a challenging problem though

There are a few other ways to generate revenue the only cost as far as I am aware of is our domain name. @dagnelies

PR #4175, opened 2025-07-20
Bump Typescript ESBuild Dependencies

wesdevpro:
### What's Changed?
- Upgrades ESBuild dependencies to their latest version so that the version of TS we are using is supported.

### DIFF
--- a/packages/docs/package.json
+++ b/packages/docs/package.json
@@ -29,6 +29,8 @@
   "dependencies": {
     "buefy": "*",
     "bulma": "^0.9.4",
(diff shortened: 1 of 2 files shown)


packages/docs/package.json
```diff
@@ -29,6 +29,8 @@
   "dependencies": {
     "buefy": "*",
     "bulma": "^0.9.4",
+    "run": "^1.5.0",
```
wesdevpro, line 32: Not sure what happen here... working

PR #4183, opened 2025-07-24
Clockpicker ref example

julzELO:
Fixes #
this pr is a follow up to [#4150](https://github.com/buefy/buefy/pull/4150#issuecomment-3111275265).

## Proposed Changes

- docs and example how to use clockpickerFace ref


dagnelies:
rather than adding a ref just to do that, I wonder if it makes more sense to have the clockpicker close by default once the time is selected

julzELO:
Its just a example for the ref. i thjink would still use the var isDragging in other use cases

dagnelies:
Hi Julien. Sorry for the delay. I wonder if this example is a bit too specific and complex for general purpose usage. ...instead, I wonder if it would be better to directly fix the clock picker to close once time is selected. Would that be appropriate?

dagnelies:
See #4189

PR #4211, opened 2025-08-16
Fix: Restore Autocomplete Arrow Key Hover Styles

### LINKED ISSUE #4207: Autocomplete does not highlight selected option when using arrow keys for selection
### Overview of the problem

**Buefy** version: 3.0.0
**Vuejs** version: v3.5.18
**OS/Browser**: Chrome/Firefox

### Description

The bug was discovered after migrating to buefy V3, did not appear in the previous version (v1.0.1)

When using b-autocomplete/b-select components, using the keyboard to navigate between the displayed results does not visually show which option is currently selected. 

Image

This is purely a visual/styling bug, it is still possible to navigate between and select different options as expected. 

Using the inspector shows that the selected option is given a class `is-hovered`, but it seems that no styling is applied to this class anymore. 

Image

The styling seems to have been removed at some point during the migration to Bulma v1: (https://github.com/buefy/buefy/commit/49d318b3e2aa8530ec1a39a61f0b88682cc7efe7#diff-65c2cdfc23d0c971c66bbdd028756a34802fe33c37f303614ce23dda05e83055L21)

### Steps to reproduce

The issue is present in every example in the docs: https://buefy.org/documentation/autocomplete

1. Click a b-autocomplete-element
2. Begin typing so that the options appear
3. Use arrow key down to select the first option

### Expected behavior

The option is visually highlighted

### Actual behavior

No highlighting occurs

wesdevpro:
Fixes #4207

## Proposed Changes

- Restores the .is-hovered styles for autocomplete dropdown items when navigating via arrow keys.

### DIFF
--- a/packages/buefy/src/scss/components/_autocomplete.scss
+++ b/packages/buefy/src/scss/components/_autocomplete.scss
@@ -1,6 +1,12 @@
 @use "bulma/sass/utilities/css-variables" as cv;
 @use "bulma/sass/utilities/initial-variables" as iv;
+@use "bulma/sass/utilities/derived-variables" as dv;
 @use "bulma/sass/utilities/controls" as controls;
+
+@include cv.register-vars((
+  "bulma-dropdown-item-hover-background-l-delta": dv.$dropdown-item-hover-background-l-delta,
+  "bulma-dropdown-item-hover-border-l-delta": dv.$dropdown-item-hover-border-l-delta
+));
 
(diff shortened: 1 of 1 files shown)


wesdevpro:
Using css is a work around as it bypasses all the sass compile time issues with sass variable scoping. This brings back the original implementation. I am going to pull these changes in. Please let me know if you have a better implementation to fix this within sass. @gmercey @arambert @kikuomax

PR #4307, opened 2026-03-10
Delay unmount so Modal's leave transition and 150ms timeout can run.

curtis-jigawatt:
## Proposed Changes

Programmatically opened modals are being removed from the DOM before they can fire callbacks. I added a timeout to the onClose event to allow time for the callbacks and close animation.

### DIFF
--- a/packages/buefy/src/components/modal/index.ts
+++ b/packages/buefy/src/components/modal/index.ts
@@ -69,7 +69,10 @@ class ModalProgrammatic {
                         ...propsData,
                         programmatic: true,
                         onClose: () => {
-                            vueInstance.unmount()
+                            // Delay unmount so Modal's leave transition and 150ms timeout can run.
+                            setTimeout(() => {
+                                vueInstance.unmount()
+                            }, 160)
                         },
                         // intentionally overrides propsData.onCancel
                         // to prevent propsData.onCancel from receiving a "cancel" event


wesdevpro:
Hey @curtis-jigawatt, thanks for tracking this down and submitting a fix! 

One concern with the `setTimeout(..., 160)` approach: the 160ms is derived from an internal implementation detail in `Modal.vue` (the 150ms timeout + a small buffer). If that internal value ever changes, or if a user uses a longer CSS transition, this silently races again.

Vue's `<Transition>` component exposes an `@after-leave` hook for exactly this scenario — it fires deterministically when the leave transition has fully completed. Basing the unmount on that would be more robust:

```ts
// Modal.vue — add an emit on the transition's after-leave:
// <Transition @after-leave="$emit('afterLeave')">

// index.ts — use it instead of setTimeout:
onAfterLeave: () => {
    vueInstance.unmount()
},
```

Would you be open to reworking along those lines? Happy to help if useful. The `setTimeout` approach is workable as a stopgap but I'd rather not hardcode a timing assumption into the library.

curtis-jigawatt:
@wesdevpro Yeah man, whatever you think! I'm definitely not a fan of setTimeout either, so if you've got a better solution, go for it! I mainly just don't like to submit a bug without a possible fix.

PR #4349, opened 2026-07-16
fix(input): match icon size to is-medium field

### LINKED ISSUE #4348: With `is-medium`, the icon size does not match the form field size.
If you set the size of the form fields to `size=“is-medium”`, the icon size does not change; it remains at 24px.

**Buefy** version: [3.0.8]
**Vuejs** version: 1.0.4]
**Windows/Firefox**:

lazerg:
Fixes #4348

## Proposed Changes

- The `iconSize` computed had an empty `return` for `is-medium`, so the icon fell back to the default 24px and didn't grow with the field.
- `is-small` and `is-medium` both return `this.size` (shared fall-through), so a medium field gets the matching icon size (e.g. `mdi-36px`).

### DIFF
--- a/packages/buefy/src/utils/FormElementMixin.spec.ts
+++ b/packages/buefy/src/utils/FormElementMixin.spec.ts
@@ -38,9 +38,4 @@ describe('FormElementMixin', () => {
         expect(wrapper.emitted().blur).toBeTruthy()
         expect(wrapper.vm.checkHtml5Validity).toHaveBeenCalled()
     })
-
-    it('should scale the icon to match an is-medium size', async () => {
-        await wrapper.setProps({ size: 'is-medium' })
-        expect(wrapper.vm.iconSize).toBe('is-medium')
-    })
 })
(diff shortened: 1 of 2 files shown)


wesdevpro:
@lazerg please double check that all linting passes...

lazerg:
Just checked, lint and type-check both pass locally. 
The one that runs and fails (`Field.spec.ts` 's `hasAddons` / `fragment` test) isn't related to this PR. 
It fails the same way on a clean dev checkout with none of my changes.

wesdevpro:
> Just checked, lint and type-check both pass locally. The one that runs and fails (`Field.spec.ts` 's `hasAddons` / `fragment` test) isn't related to this PR. It fails the same way on a clean dev checkout with none of my changes.

That was me 😂 I will work on it :)

wesdevpro:
https://github.com/buefy/buefy/commit/d7e7d1c96cc50fc3cd06882596dbf029daefd79b

Issue #3685, opened 2022-02-24
Better a11y: correct usage of role="tablist" and role="tab" for Tabs


mateuswetah:
### Overview of the problem

I've been doing some tests with the [Axe Dev Tools](https://www.deque.com/axe/devtools/) extension. It does a pretty good job on pointing accessibility errors. While I think Buefy has improved enormously in the last years regarding this topic, the reports shows that there are some misuse of Aria attributes, so we have some work to do! This is the third of a series of small issues that I am raising to cover the ones that I understand. I am by no means an expect on the topic, so would appreciate reviews and comments.

**Buefy** version: 0.9.17
**Vuejs** version: 2.6.3
**OS/Browser**: Chrome/Linux

### Description

The Tabs component has a `role="tablist"` in its `<nav>` wrapper and a `role="tab"` in the <a> link. This seems to be wrong, as the `<ul>` should be the `tablist` and its `<li>` children should be the tabs.

### Steps to reproduce
1. Install the [Axe Dev Tools](https://www.deque.com/axe/devtools/) extension. Sadly, only for Chrome...
2. Run it on the documentation page about the Tabs component: https://buefy.org/documentation/tabs
3. Check the issues panel. There shall be a series of ones regarding **Certain ARIA roles must contain particular children**

![image](https://user-images.githubusercontent.com/1184874/155577494-804c8c8c-7db6-47b5-a804-c99971dd8dd1.png)

### Expected behavior

As suggested, to remove the `role="tablist"` from `<nav>` to `<ul>` element, and  `role="tab"` from `<a>` to `<li>`. Should be a simple pull request ;)

jtommy:
I am not so expert but I trust you, free to make a PR as previous

PR #4351, opened 2026-07-20
Adjust tab aria to be screen reader friendly

AERobert:
As a screen reader user, I noticed accessibility issues with the tab lists and tab items in an app using Buefy. Screen readers treated the clickable anchor elements and the `<li>` elements with `role="tab"` as separate elements. As a result, selecting a tab required moving the screen reader cursor inside the list item and activating the anchor element.

Keyboard controls, such as arrow-key navigation, also did not work unless one of the anchor elements was focused. With some screen readers, this made it difficult to determine which tab was selected because the `aria-selected` attribute was located on the `<li>` element.

## Proposed Changes

This PR proposes moving the `role="tab"`, `aria-controls`, and `aria-selected` attributes from the `<li>` element to the `<a>` elements in both branches of the component. It also adds `role="presentation"` to the `<li>` elements so that screen readers ignore their semantics and treat the clickable `<a>` elements as the tabs.

This is a partial reversal of #3686. That change correctly moved the `tablist` role to the `<ul>` element, resolving #3685, but it also moved the `aria-selected` attribute and the `tab` role to the `<li>` element. I am not sure why that part of the previous change was included, but the arrangement proposed in this PR is more screen-reader-friendly in every screen reader I tested.

## Testing and reference

I tested this change on a local version of the [tabs page in the documentation](https://buefy.org/documentation/tabs) using VoiceOver on macOS and NVDA, JAWS, and Narrator on Windows. Tab behavior improved in all the screen readers I tested. Keyboard controls worked correctly for navigating among and selecting tabs, and it was no longer needed to move the screen reader cursor inside a list item to activate its anchor element.

I verified with axe-core that this change does not regress #3685: the `aria-required-children` and `aria-required-parent` rules pass both before and after. The accessibility error reported in #3685 was resolved just by moving `role="tablist"` to the `<ul>` element, which this PR preserves.

A similar placement of roles can be seen at the top of the ARIA Authoring Practices Guide page about the `presentation` role: [https://www.w3.org/WAI/ARIA/apg/practices/hiding-semantics/#introduction](https://www.w3.org/WAI/ARIA/apg/practices/hiding-semantics/#introduction).

### DIFF
--- a/packages/buefy/src/components/tabs/Tabs.vue
+++ b/packages/buefy/src/components/tabs/Tabs.vue
@@ -16,9 +16,7 @@
                     v-show="childItem.visible"
                     :class="[ childItem.headerClass, { 'is-active': childItem.isActive,
                                                        'is-disabled': childItem.disabled }]"
-                    role="tab"
-                    :aria-controls="`${childItem.uniqueValue}-content`"
-                    :aria-selected="`${childItem.isActive}`"
+                    role="presentation"
                 >
                     <b-slot-component
                         :ref="`tabLink${childItem.index}`"
@@ -28,6 +26,9 @@
                         tag="a"
                         :id="`${childItem.uniqueValue}-label`"
                         :tabindex="childItem.isActive ? 0 : -1"
+                        role="tab"
+                        :aria-controls="`${childItem.uniqueValue}-content`"
+                        :aria-selected="`${childItem.isActive}`"
                         @focus="currentFocus = childItem.index"
                         @click="childClick(childItem)"
                         @keydown="manageTabKeydown($event, childItem)"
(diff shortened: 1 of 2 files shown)


wesdevpro:
If you, @AERobert, have time to please cherry pick your commit on top of the latest changes of history from buefy's main branch.

AERobert:
Ok, I rebased onto the latest main branch.

wesdevpro (commented):
Thanks for the thorough writeup and cross-browser/screen-reader testing on this. Really appreciate the rigor, especially citing the APG presentation role guidance and verifying against axe-core that #3685 doesn't regress.

Reviewed the diff and the reasoning checks out: moving role="tab"/aria-controls/aria-selected onto the <a> and adding role="presentation" on the <li> matches the canonical WAI-ARIA APG tabs pattern exactly, and it correctly preserves the tablist-on-<ul> fix from #3686 rather than reverting it. Snapshot update looks consistent with the template change.

No doc updates needed on our end. The docs site renders the live component, so it picks this up automatically.

Going to merge this. Thanks again for digging into it and testing across VoiceOver, NVDA, JAWS, and Narrator. That's a level of verification we don't always get.

wesdevpro (approved):

